PR #4, opened 2015-02-01
Update class-wcpdf-export.php

LDAV:


Spreeuw:
Hi, thanks for the PR - can you give a bit of background behind these changes and/or describe the use case?

LDAV:
my english is awful... sorry. I'll try.

Probably you know that in Europe if you are a professional or an entrepreneur, when you buy something you have to specify a VAT Number. In this case we can talk about an "invoice". All other case is a "receipt". It depends on what the customer decides when he creates the order. More, the invoices must have a number, sequential, different from the receipts.
https://github.com/LDAV/woocommerce-pdf-italian-add-on
The code I've added is useful if you have different templates for different purposes (invoices or receipt or return or...). More, you can decide or not if you want to increase the invoice counter.

The best would be if you can specify more than 2 templates, not only invoice or packing-slip, and have a different counters for each type of document.

Spreeuw:
What you call a receipt is already a feature of the Professional extension, called "Proforma Invoice".
https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/
Invoices (in Europe too) are not limited to companies with VAT numbers, you can send tax invoices to people as well.

I believe you shouldn't rewrite the invoice template with a receipt, because internally it will still be handled as an 'invoice', with everything that comes with it.

LDAV:
Talking about Italy, a "Proforma Invoice" could have a different meaning. For a Professional, for instance, a Proforma Invoice is a special document that professionals use to start a payment process of a Public Administration. Means "it follows a regular invoice", that's because we call it "pro-forma". A receipt is different, is a definitive document, has different rules. When I buy something I have to ask before if I want an invoice OR a receipt. Is the customer that decides, not the seller. So, to extend your code we had to reset some internal function you use for the invoices, to delete that "everything that comes with it".
I don't know if your commercial plugin have these functionalities. I suppose you have different counters for each type of document (a Credit Note in Italy should have), but maybe you have not the additional fields we need to add and check in some cases. We have a dozen different RegEx for each European VAT number type. And - for us - it's easier to provide also different templates.

PR #6, opened 2015-07-04
Added additional shop info settings

amereservant:
I added 3 additional shop info settings (telephone/fax/email) and matching invoice/packing slip functions.  I excluded a function for shop_url since `bloginfo('url')` can be used, however it might be worth adding generic functions  `get_shop_url` and `shop_url` that return the bloginfo() value for better continuity in designing invoice templates.


Spreeuw:
Thanks for the suggestion! I'm hesitant to include it though, as I don't see any benefit of doing this,  since you can already enter it in the address directly. If you're using it in a different place altogether (which would require a custom template), why not put it in the template?

If you need 'custom' fields for your template, you can use the extra fields from the premium templates for this purpose. (extra_1, extra_2 & extra_3).

Let me know if I misunderstood you though!

PR #8, opened 2015-08-15
Correct Polish translation of "Invoice Number:" (pl_PL)

damianwajer:
I've added the correct Polish translation of "Invoice Number:", it should be "Nr zamówienia:" instead of "Zamówienie nr:" (which already exist as "Order Number:").


Spreeuw:
Thanks for the pull request. The previous translator said that because the invoice cannot be used as an official invoice in Poland, it shouldn't be translated as Nr zamówienia... Personally I think in that case you should not display the invoice number at all, but he said it would be confusing.
What are your thoughts here?

dwajer:
I think now it's more confusing, because we have duplicate label with wrong translation.

I'm not sure if or why the invoice cannot be used as an official invoice in Poland, but if so, we should hide this field in the template, not display it like this. If anyone wants to customize the template it can be done with this plugin by creating the new template but that's a different story, right? ;)

Spreeuw:
Couldn't agree more. After all, it is still possible to make the invoice compatible with a different template, or use custom translations if you really want to show 'order number'. I have merged the pull request!

PR #9, opened 2015-09-05
Improve Polish translations regarding "invoice" word

damianwajer:
This pull request is related to: [Correct Polish translation of "Invoice Number:"](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/8)

I've found more confusing translations of "invoice" word (e.g. another duplicate label: "Invoice date" as "Data zamówienia" which means "Order date" but it should be "Data faktury" - "Invoice date"). 

If we've decided to use "faktura" word ("invoice") instead of "zamówienie" ("order") we should use it everywhere for consistency. In line 28 of .po file I've added a comment in Polish which is explanation for this change (to meet legal requirements we can customize invoice template, add additional fields etc.).

PS. I've added some missing Polish translation in following commits: 12f4e6f, 3da363e.


rgb4u:
I'm also interested in this translation, it will be nice if woothemes 
add this translation to original pdf invoice plugin

Spreeuw:
@rgb4u Thanks for the heads up! I merged an earlier pull request from @damianwajer already, but completely missed this update. I have just merged the changes into the plugin (note that this plugin is not developed by WooThemes but me (for WP Overnight) :smile:

rgb4u:
Thank you for respond, i wish i could use this translation in my woocommerce pdf invoice plugin, i fight with polish signs currently...without effects ;/

PR #12, opened 2015-12-10
Safety check for floats with 1.7763568394003E-15

jnachtigall:
See this error report https://wordpress.org/support/topic/get_order_discount-return-value-of-000-e?replies=1

### DIFF
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -800,8 +800,9 @@ if ( !class_exists( 'WooCommerce_PDF_Invoices' ) ) {
 				'value'		=> $this->export->wc_price($discount_value),
 				'raw_value'	=> $discount_value,
 			);
-
-			if ( $discount_value != 0 ) {
+			// discount is float and (int) 0 can equal (float) 1.7763568394003E-15
+			// Using 0.005 instead of 0.01 to have a safety margin for rounding
+			if ( $discount_value >= 0.005 ) {
 				return apply_filters( 'wpo_wcpdf_order_discount', $discount, $type, $tax );
 			}
 		}


jnachtigall:
Cannot post to the support forum at wordpress.org right now, so I just post it here now:

jeez, I just typed in a lengthy message and its lost now.

Anyway again:

I sent you a PR at https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/12

Turns out that this <a href="https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/master/woocommerce-pdf-invoices-packingslips.php#L764">line</a>:
`$discount_value = $this->export->order->get_total_discount( false );`
sometimes returns
`$discount_value = (float) 1.7763568394003E-15`

If you google for this number, you see why:
http://forums.devshed.com/php-development-5/1-1-equals-0-a-384040.html
http://board.phpbuilder.com/showthread.php?10287705-Weird-math-results

I think the code should check if discount is at least 1 cent instead of against 0.

Spreeuw:
It's better to round instead, because as you can see, it doesn't check for > 0 anymore but != 0 so that 'positive' discounts also work (some shops that sell produce end up selling more than what was ordered and then the total discount is actually an overcharge). Even more importantly, this return value comes from WooCommerce, so perhaps they should do the check/fix... Will give this some thought!

jnachtigall:
I agree that this should be fixed at woocommerce itself. Still it would not hurt to have a further safety guard by checking against >= 0.01 and <= -0.01

In the end this is more of a php bug imho, have not seen this in any other language.

--
Sorry for shortness, mail was sent from my smartphone.



Am 11. Dezember 2015 08:41:29 MEZ, schrieb Spreeuw <notifications@github.com>:

PR #17, opened 2016-11-02
WC 2.2: use wc_get_order instead of new WC_Order

Dartui:


Spreeuw:
Hi! Thanks for the pull request! This would break backwards compatibility with WC2.1, so I'd prefer building in a wrapper function that uses wc_get_order() only when possible. I'm working on a complete rewrite of the plugin and will take this into account. For now, new WC_Order works on all versions so I'm leaving this for now. I really appreciate the contribution though!

Dartui:
You should work on actual master branch of WooCommerce. 2.7 version (which will launch in the new year) will introduce a lot of tweaks, and some functions might not work. Thanks for fast response!

Spreeuw:
Very much aware of this, CRUD will introduce a lot of new challenges and opportunities :)
Like I said though, the rewrite will address this. Thanks!

Spreeuw:
fixed in 0b41cfe0899d9563887ad30398e835b3f37b3cba

Issue #34, opened 2017-07-15
move `wpo_wcpdf_check_privs` on generate_pdf_ajax()


fauzie:
Just downloading and trying this plugin today, to allow download pdf from my account page.
I have other user to allow view and check specific order from my account page besides order `customer_id`.
But seriously, you put `apply_filters` before your defaults. It becomes useless.

includes/class-wcpdf-export.php Line 438

Spreeuw:
Can you explain what you're trying to do?

Taking the filter from 2.0 (the current master branch won't be updated, and RC2 is release ready), it's [class-wcpdf-main.php, line 156](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/v2.0-rc.2/includes/class-wcpdf-main.php#L156)
``` php
apply_filters( 'wpo_wcpdf_check_privs', !current_user_can( 'manage_woocommerce_orders' ) && !current_user_can( 'edit_shop_orders' ) && !isset( $_GET['my-account'] ), $order_ids )
```
This filter takes 2 arguments, the first is the default return, e.g.:
```
 !current_user_can( 'manage_woocommerce_orders' ) && !current_user_can( 'edit_shop_orders' ) && !isset( $_GET['my-account'] )
```
and the second is the array of `$order_ids`.

You can override these defaults with:
``` php
add_filter( 'wpo_wcpdf_check_privs', 'wpo_wcpdf_check_privs_custom', 10, 2 );
function wpo_wcpdf_check_privs_custom( $not_allowed, $order_ids ) {
    $not_allowed = false;
    return $not_allowed;
}
```
Am I missing something?

fauzie:
In latest version 2.x was fixed of course. Glad to hear that.

Look at my comment date, it's July 15. I downloaded this plugin from wp.org plugin site and it's version is 1.6.6 still. And the file is ```includes/class-wcpdf-export.php Line 438```

> I have other user to allow view and check specific order from my account page

I mean, move line ```438``` after your own check on line ```453``` at [this file](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/v1.6.6/includes/class-wcpdf-export.php)

I will update your plugin to version 2 later.
Great plugin btw, and thank you for your reply.

PR #36, opened 2017-07-24
better use filter wpo_wcpdf_check_privs

fauzie:
Continue from #34 

Here's what I want to do:
* I have sales user role that has permission to view order, generate pdf on frontend
* The salesman user_id has added in order meta, and it's not ```customer_id``` meta

If i use filter ```wpo_wcpdf_check_privs``` it still remain obstructed by [class-wcpdf-main.php, line 174](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/master/includes/class-wcpdf-main.php#L174)

So this is my solution, and maybe better usability filter ```wpo_wcpdf_check_privs``` for others.


Spreeuw:
Nice improvement, thanks!

PR #37, opened 2017-08-14
Filter to disable page break

luizbills:


Spreeuw:
Thanks for the suggestion! I have actually implemented this in a slightly different way in feb8ed0646222212ea3a4269ed2fcb2f6425c248, which allows for more advanced filtering. That way you could wrap each document in a table cells + rows and adding a wrapper table.

Let me know what you think!

luizbills:
thank you, sir

PR #39, opened 2017-08-16
Added [invoice_year_short]

zahradt:
Added [invoice_year_short] code allowing to add years without centuries into invoice numbers.


Spreeuw:
rather than adding extra shortcodes like this I would propose to add a catch all shortcode that takes a date format argument, something like:
```
[invoice_date="y"]
[invoice_date="y-m-d-His"]
```

PR #44, opened 2018-01-16
- add actions to add additional fields to invoice and hook to save them

Delwing:


slaFFik:
@Delwing Your PR is using spaces, while the plugin source code is using WordPress Coding Standards for PHP and that means to use tabs.

Delwing:
Thanks for pointing that out. I've corrected that.

PR #46, opened 2018-04-12
add extra order parameter to filter wpo_wcpdf_attach_documents

coccoinomane:
Hello!
I am calling the `wpo_wcpdf_attach_documents` filter but I need `$order` information in the callback.
This seems a fairly straigthforwards improvement, unless I am making some horrible mistake!
Thanks,
Guido


Spreeuw:
While it wouldn't do much harm, why not just use `wpo_wcpdf_custom_attachment_condition` instead?
The documents is just an attachment/order agnostic list of documents and per document you can then determine whether it should be attached or not

Spreeuw:
No worries, glad to hear this works for you too :)

PR #49, opened 2018-07-02
FIX Order_Document::__construct undefined variable

christopherdarling:
when passing type int to __construct $order_id variable is undefined


christopherdarling:
No problem :-) thanks for the plugin

PR #52, opened 2018-08-13
Added _year_short option for document numbers

dylankiss:


Spreeuw:
already possible in a much more flexibale `[invoice_date="y"]` format: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/v2.1.10/includes/documents/class-wcpdf-document-number.php#L151-L166

PR #82, opened 2020-05-14
Download invoice shortcode

alexmigf:


includes/class-wcpdf-frontend.php
```diff
@@ -79,6 +80,50 @@ public function woocommerce_api_invoice_number ( $data, $order ) {
...
+		if( !is_object($order) || !isset($order) || empty($order) ) return;
+
+		// Link text
+		$link_text = __('Download invoice (PDF)', 'woocommerce-pdf-invoices-packing-slips');
+		if( ! empty($values['link_text']) ) {
+			$link_text = $values['link_text'];
+		}
+
+		// User permissions
+		$debug_settings = get_option('wpo_wcpdf_settings_debug');
```
Spreeuw, line 114: @alexmigf I would set default to array so that `isset($debug_settings['guest_access'])` will not throw any notices
e.d.
```php
$debug_settings = get_option('wpo_wcpdf_settings_debug',array());
```

PR #86, opened 2020-05-22
Regenerate document

Terminator-Barbapapa:


assets/css/order-styles.css
```diff
@@ -68,12 +68,46 @@
 }
 
 .dashicons-edit,
-.dashicons-trash {
+.dashicons-trash,
```
Spreeuw, line 71: These should all be more specific to the WCPDF data panel

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -297,6 +297,25 @@ public function delete( $order = null ) {
 		do_action( 'wpo_wcpdf_delete_document', $this, $order );
 	}
 
+	public function regenerate( $order = null ) {
+		$order = empty( $order ) ? $this->order : $order;
+		if ( empty( $order ) ) {
+			return; //Nothing to update
+		}
+
+		//Get most current settings
```
Spreeuw, line 306: would there be a way to unify this with the normal `$this->get_settings()` method? It's good practice to prevent code duplication whenever possible.

includes/class-wcpdf-admin.php
```diff
@@ -672,6 +688,46 @@ public function delete_document() {
...
+			$document = wcpdf_get_document( $document, wc_get_order( $order_id ) );
+			if ( !empty($document) && $document->exists() ) {
+				$document->regenerate();
+				$response = array(
+					'message' => $document->get_type()." regenerated",
+				);
+				wp_send_json_success($response);
+			} else {
+				wp_send_json_error( array(
+					'message' => 'document does not exist',
```
Spreeuw, line 721: this also happens when you click the delete button and then the regenerate button, but the above error is just internal, without feedback. I would suggest to remove (or hide) the regenerate button altogether when deleting a document

assets/css/order-styles.css
Terminator-Barbapapa: Solved with: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/d66f3cd72ed257f979294e997ecddeaa5b581da8

includes/class-wcpdf-admin.php
Terminator-Barbapapa: Solved with: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/1196cb364f728f037e2d66ab25cb29352818cbe8

includes/documents/abstract-wcpdf-order-document.php
Spreeuw: Since this is more of an optimization and doesn't change anything functionality wise, I'll open separate issue and merge this one

Issue #79, opened 2020-05-01
Add logo width to the settings


alexmigf:

Spreeuw:
we currently set the height in CSS, perhaps it's best to make the setting control height rather than width too?

alexmigf:
I agree.

PR #87, opened 2020-05-22
Add logo height setting

alexmigf:
#79

### DIFF
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -255,7 +255,6 @@ class Install {
 					'currency_font'				=> array( 'wpo_wcpdf_template_settings' => 'currency_font' ),
 					'paper_size'				=> array( 'wpo_wcpdf_template_settings' => 'paper_size' ),
 					'header_logo'				=> array( 'wpo_wcpdf_template_settings' => 'header_logo' ),
-					'header_logo_height'		=> array( 'wpo_wcpdf_template_settings' => 'header_logo_height' ),
 					'shop_name'					=> array( 'wpo_wcpdf_template_settings' => 'shop_name' ),
 					'shop_address'				=> array( 'wpo_wcpdf_template_settings' => 'shop_address' ),
 					'footer'					=> array( 'wpo_wcpdf_template_settings' => 'footer' ),
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -255,6 +255,7 @@ class Install {
 					'currency_font'				=> array( 'wpo_wcpdf_template_settings' => 'currency_font' ),
 					'paper_size'				=> array( 'wpo_wcpdf_template_settings' => 'paper_size' ),
 					'header_logo'				=> array( 'wpo_wcpdf_template_settings' => 'header_logo' ),
+					'header_logo_height'		=> array( 'wpo_wcpdf_template_settings' => 'header_logo_height' ),
 					'shop_name'					=> array( 'wpo_wcpdf_template_settings' => 'shop_name' ),
 					'shop_address'				=> array( 'wpo_wcpdf_template_settings' => 'shop_address' ),
 					'footer'					=> array( 'wpo_wcpdf_template_settings' => 'footer' ),
--- a/includes/class-wcpdf-settings-callbacks.php
+++ b/includes/class-wcpdf-settings-callbacks.php
@@ -80,6 +80,10 @@ class Settings_Callbacks {
 			$type = 'text';
 		}
 
+		if (empty($current)) {
+			$current = $default;
+		}
+
 		printf( '<input type="%1$s" id="%2$s" name="%3$s" value="%4$s" size="%5$s" placeholder="%6$s" %7$s/>', $type, $id, $setting_name, esc_attr( $current ), $size, $placeholder, !empty($disabled) ? 'disabled="disabled"' : '' );
 	
 		// output description.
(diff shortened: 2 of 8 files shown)


templates/Simple/template-functions.php
```diff
@@ -5,3 +5,15 @@
  */
 if ( ! defined( 'ABSPATH' ) ) exit; // Exit if accessed directly
 
+add_action( 'wpo_wcpdf_custom_styles', 'wpo_wcpdf_templates_get_footer_settings', 10, 2 );
+
+function wpo_wcpdf_templates_get_footer_settings( $document_type, $document ) {
+    $header_logo_height = str_replace( ' ', '', $document->get_header_logo_height() );
```
Spreeuw, line 11: I think sanitizing should be done in the getter method

includes/class-wcpdf-settings-general.php
```diff
@@ -132,6 +132,19 @@ public function init_settings() {
...
 			),
+			array(
+				'type'		=> 'setting',
+				'id'		=> 'header_logo_height',
+				'title'		=> __( 'Logo height', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback'	=> 'text_input',
+				'section'	=> 'general_settings',
+				'args'		=> array(
+					'option_name'	=> $option_name,
+					'id'			=> 'header_logo_height',
```
Spreeuw, line 143: For clarity, set the 30mm as default here?

includes/class-wcpdf-install.php
```diff
@@ -255,6 +255,7 @@ protected function upgrade( $installed_version ) {
 					'currency_font'				=> array( 'wpo_wcpdf_template_settings' => 'currency_font' ),
 					'paper_size'				=> array( 'wpo_wcpdf_template_settings' => 'paper_size' ),
 					'header_logo'				=> array( 'wpo_wcpdf_template_settings' => 'header_logo' ),
+					'header_logo_height'		=> array( 'wpo_wcpdf_template_settings' => 'header_logo_height' ),
```
Spreeuw, line 258: This is the settings migration routine, but this setting wasn't there in 1.X so nothing to migrate here! This can be omitted

includes/legacy/class-wcpdf-legacy-settings.php
```diff
@@ -28,6 +28,7 @@ public function load_legacy_settings() {
 				'currency_font'				=> array( 'wpo_wcpdf_template_settings' => 'currency_font' ),
 				'paper_size'				=> array( 'wpo_wcpdf_template_settings' => 'paper_size' ),
 				'header_logo'				=> array( 'wpo_wcpdf_template_settings' => 'header_logo' ),
+				'header_logo_height'		=> array( 'wpo_wcpdf_template_settings' => 'header_logo_height' ),
```
Spreeuw, line 31: this wasn't part of the legacy settings, can be omitted

templates/Simple/style.css
```diff
@@ -100,7 +100,7 @@ table.head {
 }
 
 td.header img {
-	max-height: 3cm;
+	/* max-height: 3cm; */
```
Spreeuw, line 103: I would leave this in as default, maybe with an annotation that it is overriden by the settings by default

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -504,7 +504,7 @@ public function get_header_logo_id() {
 	 */
 	public function get_header_logo_height() {
 		if ( !empty( $this->settings['header_logo_height'] ) ) {
-			return apply_filters( 'wpo_wcpdf_header_logo_height', $this->settings['header_logo_height'], $this );
+			return apply_filters( 'wpo_wcpdf_header_logo_height', sanitize_text_field($this->settings['header_logo_height']), $this );
```
Spreeuw, line 507: I would apply the `str_replace` that you applied in the earlier commit here (in case someone puts a space between the number and the unit)

includes/class-wcpdf-settings-general.php
```diff
@@ -247,6 +250,29 @@ public function init_settings() {
 		return;
 	}
 
+	public function text_element_callback( $args ) {
```
Spreeuw, line 253: what does this do that the text_input doesn't do? Would it be something to add to the default text_input?

alexmigf: How should work? I tested and didn't work

includes/class-wcpdf-settings-callbacks.php
```diff
@@ -80,6 +80,10 @@ public function text_input( $args ) {
 			$type = 'text';
 		}
 
+		if (empty($current) && isset($default)) {
+			$current = $default;
+		}
```
Spreeuw, line 85: is this not already handled here?
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c64fa0e23eacdd55dfb8df1b14fd8a1b7b27766e/includes/class-wcpdf-settings-callbacks.php#L535-L538

alexmigf: 6f9fe5d137ac5544a0ef87a15a7e86681e4785f0

Spreeuw: I think it makes more sense to do this in the `normalize_settings_args` method so that all callbacks can use it?

alexmigf: Yes, i will fix.

PR #92, opened 2020-08-13
More billing/shipping address visibility options

### LINKED ISSUE #85: add option to always show shipping address on invoice
Change this:
display-shipping-address-old

to this:
display-shipping-address-new

(and do the same for packing slip, proforma etc.)

Existing templates do not support this yet:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e9ae661705bacd3ca650d769cf6bb9b6a66a2a7d/templates/Simple/invoice.php#L43

so I see two possibilities:

1. We use the [wpo_wcpdf_ships_to_different_address](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e9ae661705bacd3ca650d769cf6bb9b6a66a2a7d/includes/documents/abstract-wcpdf-order-document-methods.php#L95) filter to return true for all documents with the 'Always' setting enabled.
2. We change the template and accept that legacy templates will not support the option.

Any thoughts?

Spreeuw:
fixes #85

### DIFF
--- a/includes/documents/class-wcpdf-invoice.php
+++ b/includes/documents/class-wcpdf-invoice.php
@@ -211,12 +211,17 @@ class Invoice extends Order_Document_Methods {
 				'type'			=> 'setting',
 				'id'			=> 'display_shipping_address',
 				'title'			=> __( 'Display shipping address', 'woocommerce-pdf-invoices-packing-slips' ),
-				'callback'		=> 'checkbox',
+				'callback'		=> 'select',
 				'section'		=> 'invoice',
 				'args'			=> array(
 					'option_name'		=> $option_name,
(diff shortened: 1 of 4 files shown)


alexmigf:
Option **Always** only works if they are different.

Spreeuw:
Are you on the "Simple" template? I haven't applied these changes to the other templates yet

alexmigf:
Yes, i am, i saw that.

Spreeuw:
I double checked, works for me on Simple:

Same shipping & billing
1. Always = shows
2. Only when shipping address is different from the billing address = doesn't show

Different shipping & billing
1. Always = shows
2. Only when shipping address is different from the billing address = shows

Issue #95, opened 2020-08-20
Add new setting for 'Invoice notes'


alexmigf:
Add a 'Invoice notes' section to the PDF invoice data tab before... Just like the 'customer notes' but specifically for the invoice, which can be entered in the backend.

alexmigf:
@Spreeuw 

About this, i think we talked before but i can't find the conversation. The notes are added in the document settings correct?

Spreeuw:
no, these are per-order notes, which would be in the "PDF Invoice Data" panel.

PR #97, opened 2020-10-02
Added invoices notes to orders and simple invoice template. #95

alexmigf:


Spreeuw:
rather than using functions (and styles etc.) `invoice_notes`, I would propose to make this more generic as `document_notes` so that adding it to other documents becomes trivial and requires less duplication

alexmigf:
Makes sense, i will do that.

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1207,6 +1207,18 @@ public function invoice_date() {
 		echo $this->get_invoice_date();
 	}
 
+	public function get_document_notes( $template_type ) {
```
Spreeuw, line 1210: do we need the `$template_type` argument here? I think we can safely assume that 99.99% of usage will be within the same (current) document. For the exceptions (custom setups) people can load that document by themselves I think?

Issue #105, opened 2020-10-15
long word (or email address) in first address column pushing order data column away


Spreeuw:
In this example, the email address is very wide), and it seems dompdf doesn't want to break it up, pushing everything else away
![address-column-issue](https://user-images.githubusercontent.com/3217773/96129846-6e968000-0ef7-11eb-900a-3b639b3f108f.png)
Should be addressed in Simple Premium Too

For this particular scenario, the issue could be fixed by hiding the shipping address column because there only was a billing address. While thay may be a good improvement (we'd need to detect if there is a shipping address and if so, hide or disable it), the issue would still occur if both need to be displayed.

Spreeuw:
@alexmigf we have closed this, but I have only just now noticed that this is applied to all `th` & `td`. This seems to be producing unexpected results, a customer reported this:
![colon-next-line](https://user-images.githubusercontent.com/3217773/100341166-8116d580-2fdc-11eb-9cc0-74f302d782c3.png)
While on the one hand a catch-all solution is nice, I wonder if we should leave this primarily to `td` (since `th` is usually static), or even limit this specifically to the address columns. What do you think?

alexmigf:
Yes, only to addresses seems more reasonable.

Spreeuw:
closed by 6a94b35

PR #106, opened 2020-10-16
Fixes bug on long text inside billing address #105

alexmigf:


Spreeuw:
Does the `word-wrap: break-word;` CSS on its own fix this too?

alexmigf:
Yes, it does.

PR #108, opened 2020-10-21
Add option to disable customer notes

alexmigf:


includes/documents/class-wcpdf-invoice.php
```diff
@@ -247,6 +247,18 @@ public function init_settings() {
...
+				'id'			=> 'display_customer_notes',
+				'title'			=> __( 'Display customer notes', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback'		=> 'checkbox',
+				'section'		=> 'invoice',
+				'args'			=> array(
+					'option_name'		=> $option_name,
+					'id'				=> 'display_customer_notes',
+					'store_unchecked'	=> true,
+				)
+			),
```
Spreeuw, line 261: The checkbox is now unchecked by default, but it should be enabled for existing customers. Otherwise people update the plugin, they change something in the settings (not noticing the new option, so not changing it either), save the settings and suddenly the notes are no longer displayed.

includes/class-wcpdf-settings-callbacks.php
```diff
@@ -53,6 +53,11 @@ public function checkbox( $args ) {
 
 		// output checkbox	
 		printf( '<input type="checkbox" id="%1$s" name="%2$s" value="%3$s" %4$s %5$s/>', $id, $setting_name, $value, checked( $value, $current, false ), !empty($disabled) ? 'disabled="disabled"' : '' );
+
+		// print store empty input if true
+		if( $store_unchecked ) {
+			printf( '<input type="hidden" name="wpo_wcpdf_setting_store_empty[]" value="%s"/>', $id );
```
Spreeuw, line 59: this doesn't work because the input `name` needs to be in the options array, so like this:
```php
printf( '<input type="hidden" name="%s[wpo_wcpdf_setting_store_empty][]" value="%s"/>', $option_name, $id );
```
otherwise it is not passed to the settings sanitization callback and still not saved as `0`...

Spreeuw (changes requested):
Please test with legacy orders (latest settings & test mode turned off) too, and make sure that 0 is stored for unchecked, and 'not set' is really interpreted as enabled as discussed

Terminator-Barbapapa (approved):
Seems to be working as expected. Also with old orders in combination with test mode/most current settings option.

Spreeuw:
@Terminator-Barbapapa Thanks for the review. You wrote:

> Also with old orders in combination with test mode/most current settings option.

But it should work with old orders **without** test mode/most current settings. Was that a typo?

Terminator-Barbapapa:
@Spreeuw Yes, I think we're on the same page. Sorry for my bad description. 

Old order -> Test mode/Most current settings: Off -> Display customer notes: Off -> Customer notes display: yes
Old order -> Test mode/Most current settings: On -> Display customer notes: Off -> Customer notes display: no

Which is the expected behaviour correct?

includes/documents/class-wcpdf-packing-slip.php
```diff
@@ -132,6 +132,18 @@ public function init_settings() {
...
+			array(
+				'type'			=> 'setting',
+				'id'			=> 'display_customer_notes',
+				'title'			=> __( 'Display customer notes', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback'		=> 'checkbox',
+				'section'		=> 'packing_slip',
+				'args'			=> array(
+					'option_name'		=> $option_name,
+					'id'				=> 'display_customer_notes',
+					'store_unchecked'	=> true,
```
Spreeuw, line 144: default is missing here

Spreeuw (changes requested):
default for Packing Slip needs to be set, other than that it looks good to me!

Spreeuw (approved):

PR #111, opened 2020-10-22
Save invoice number and date with regenerate document #102

### LINKED ISSUE #102: On regenerate document, also update document number/date when in edit mode
If you update the document number/date and not click on 'Update' but instead on the regenerate icon, these values will not be saved and therefore not shown in the regenerated document. Which can lead to confusion. 

Schermafbeelding-2020-10-14-om-12-17-17

So on regenerating possible edits to document number/date should be included.

First referred to this issue in the following ticket: https://secure.helpscout.net/conversation/1308091149/25853?folderId=221726

alexmigf:
closes #102

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -416,19 +416,19 @@ class Admin {
 				<!-- Editable -->
 				<div class="editable">
 					<p class="form-field _wcpdf_invoice_number_field ">
-						<label for="wcpdf_invoice_number"><?php _e( 'Invoice Number (unformatted!)', 'woocommerce-pdf-invoices-packing-slips' ); ?>:</label>
+						<label for="_wcpdf_invoice_number"><?php _e( 'Invoice Number (unformatted!)', 'woocommerce-pdf-invoices-packing-slips' ); ?>:</label>
 						<?php if ( $invoice->exists() && !empty($invoice_number) ) : ?>
-						<input type="text" class="short" style="" name="wcpdf_invoice_number" id="_wcpdf_invoice_number" value="<?php echo $invoice_number->get_plain(); ?>" disabled="disabled">
+						<input type="text" class="short" style="" name="_wcpdf_invoice_number" id="_wcpdf_invoice_number" value="<?php echo $invoice_number->get_plain(); ?>" disabled="disabled">
 						<?php else : ?>
-						<input type="text" class="short" style="" name="wcpdf_invoice_number" id="_wcpdf_invoice_number" value="" disabled="disabled">
+						<input type="text" class="short" style="" name="_wcpdf_invoice_number" id="_wcpdf_invoice_number" value="" disabled="disabled">
 						<?php endif; ?>
 					</p>
 					<p class="form-field form-field-wide">
-						<label for="wcpdf_invoice_date"><?php _e( 'Invoice Date:', 'woocommerce-pdf-invoices-packing-slips' ); ?></label>
+						<label for="_wcpdf_invoice_date"><?php _e( 'Invoice Date:', 'woocommerce-pdf-invoices-packing-slips' ); ?></label>
 						<?php if ( $invoice->exists() && !empty($invoice_date) ) : ?>
-						<input type="text" class="date-picker-field" name="wcpdf_invoice_date" id="wcpdf_invoice_date" maxlength="10" value="<?php echo $invoice_date->date_i18n( 'Y-m-d' ); ?>" pattern="[0-9]{4}-(0[1-9]|1[012])-(0[1-9]|1[0-9]|2[0-9]|3[01])" disabled="disabled"/>@<input type="number" class="hour" placeholder="<?php _e( 'h', 'woocommerce' ) ?>" name="wcpdf_invoice_date_hour" id="wcpdf_invoice_date_hour" min="0" max="23" size="2" value="<?php echo $invoice_date->date_i18n( 'H' ) ?>" pattern="([01]?[0-9]{1}|2[0-3]{1})" />:<input type="number" class="minute" placeholder="<?php _e( 'm', 'woocommerce' ) ?>" name="wcpdf_invoice_date_minute" id="wcpdf_invoice_date_minute" min="0" max="59" size="2" value="<?php echo $invoice_date->date_i18n( 'i' ); ?>" pattern="[0-5]{1}[0-9]{1}" />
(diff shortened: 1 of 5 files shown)


Spreeuw:
Looking at this I wonder if there would be a way to make this more universal for other documents - currently this only works for the invoice but any other documents that do not have `use_historical_settings` enabled are not addressed by this fix yet.
It also looks like it would be better to do this before `$document->regenerate()` because otherwise it's too late for when `archive_pdf` is enabled (in Professional).

alexmigf:
What do you suggest to address the `use_historical_settings`? Check the document type?

Spreeuw:
I think you could work around this by not making it invoice specific but by simply passing all the form fields withing the document panel/block that the "regenerate document" is triggered from. I think you can do that as serialized form data but there may be better methods (not sure if you can put all data from a form in a JSON object easily?)

Spreeuw:
This is still not ideal because while this passes the data as ` $POST['json_data']`  to the `save_invoice_number_date` method, this does not change anything for other documents. Would it help to pass the document data as an optional argument to `$document->regenerate()`?
So that you could do something like:
```php
$data = array(
    'number' => 101,
    'date' => '2020-11-11 15:15',
);
$document->regenerate( $data );
```
and then within the `regenerate` method do something like:
```php
if ( !empty( $data ) ) {
    foreach ($data as $key => $value) {
        if ( is_callable( array( $this, "set_{$key}" ) ) ) {
            $this->{"set_{$key}"}( $value );
        }
    }
}
```

alexmigf:
Can't I use `set_data()` instead of the loop you suggest?

Spreeuw:
If you have full control over the data, then that's indeed simpler

includes/class-wcpdf-admin.php
```diff
@@ -774,7 +708,7 @@ public function regenerate_document() {
 			) );
 		}
 
-		if( ! isset($_POST['json_data']) || empty($_POST['json_data']) ) {
+		if( ! isset($_POST['form_data']) || empty($_POST['form_data']) || ! isset($_POST['order_id']) || empty($_POST['order_id']) || ! isset($_POST['document_type']) || empty($_POST['document_type']) ) {
```
Spreeuw, line 711: if you're already testing `empty()` you can skip all the `!isset()`

includes/class-wcpdf-admin.php
```diff
@@ -786,18 +720,17 @@ public function regenerate_document() {
...
+		$order_id = absint( $_POST['order_id'] );
+		$document_type = sanitize_text_field( $_POST['document_type'] );
 
 		try {
-			$document = wcpdf_get_document( $document, wc_get_order( $order_id ) );
+			$document = wcpdf_get_document( $document_type, wc_get_order( $order_id ) );
 			if ( !empty($document) && $document->exists() ) {
 				// save number and date
-				$this->save_invoice_number_date($order_id, null);
+				$this->save_invoice_number_date( $order_id, null );
```
Spreeuw, line 730: shouldn't this be removed?

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -304,12 +304,17 @@ public function delete( $order = null ) {
 		do_action( 'wpo_wcpdf_delete_document', $this, $order );
 	}
 
-	public function regenerate( $order = null ) {
+	public function regenerate( $data = null, $order = null ) {
```
Spreeuw, line 307: this is assuming nobody is currently using this function yet... it may be safer to keep the order of arguments in line with previous versions unless you have a good reason for needing `$data` first?

includes/class-wcpdf-admin.php
```diff
@@ -826,6 +759,76 @@ public function debug_enabled_warning( $wp_admin_bar ) {
 			$wp_admin_bar->add_node( $args );
 		}
 	}
+
+	public function process_order_document_form_data( $request )
+	{
+		$data = array();
+
+		if( isset($request['form_data']) && ! empty($request['form_data']) ) {
```
Spreeuw, line 767: rather than passing `form_data`, I think this should receive the data directly

includes/class-wcpdf-admin.php
```diff
@@ -826,6 +759,76 @@ public function debug_enabled_warning( $wp_admin_bar ) {
...
+			$form_data = json_decode( stripslashes( $request['form_data'] ), true );
+
+			if( isset($request['document_type']) && ! empty($request['document_type']) ) {
+				$document_type = sanitize_text_field( $request['document_type'] );
+
+				if( isset($form_data['wcpdf_'.$document_type.'_number']) && ! empty($form_data['wcpdf_'.$document_type.'_number']) ) {
```
Spreeuw, line 773: this needs the document slug for Pro compatibility, not the type

includes/class-wcpdf-admin.php
```diff
@@ -416,11 +416,11 @@ public function data_input_box_content ( $post ) {
 				<!-- Editable -->
 				<div class="editable">
 					<p class="form-field _wcpdf_invoice_number_field ">
-						<label for="_wcpdf_invoice_number"><?php _e( 'Invoice Number (unformatted!)', 'woocommerce-pdf-invoices-packing-slips' ); ?>:</label>
+						<label for="wcpdf_invoice_number"><?php _e( 'Invoice Number (unformatted!)', 'woocommerce-pdf-invoices-packing-slips' ); ?>:</label>
```
Spreeuw, line 419: in pro we actually prefix everything with an underscore already... so maybe follow suit here in free too

includes/class-wcpdf-admin.php
```diff
@@ -530,77 +530,11 @@ public function save_invoice_number_date($post_id, $post) {
...
-						),
-						'b'		=> array(),
-					);
-					// sanitize
-					$invoice_notes = wp_kses( stripslashes($_POST['wcpdf_invoice_notes']), $allowed_html );
-					// set notes
-					$invoice->set_notes( $invoice_notes );
-				}
-
+				$document_data = $this->process_order_document_form_data( $_POST );
```
Spreeuw, line 536: I don't think this will work because `$_POST` will not have `form_data` set (see comment below)

Spreeuw (commented):
some additional remarks

includes/class-wcpdf-admin.php
alexmigf: Number is not being saved, are you sure i don't need this?

Spreeuw: If the number is not being saved, then your `set_data` below this is not working properly. But at this point I can't tell because you have resolved all conversations without pushing the commits that resolved it 😄

alexmigf: I know, sorry, let me try debug quickly to see if i can find the "bug".

Issue #109, opened 2020-10-21
"Disable for free orders" only affects automatic creation


Spreeuw:
I can't imagine a scenario where someone would want to disable the automatic creation of attachments for free orders but still be able to generate them manually. I think we should disable then entirely by using the `wpo_wcpdf_document_is_allowed` filter (perhaps with the exception to allow existing invoices too, so that those don't disappear: for backwards compatibility)

PR #112, opened 2020-10-23
Disable free always instead of only in attachments, but keeping alrea…

alexmigf:
…dy generated invoices #109


alexmigf:
TODO: add `is_callable` for `get_total` method.

Spreeuw:
Looks good to me! One thing we may want to change is the settings text, currently it still refers to automatic generation:

> Disable automatic creation/attachment when the order total is €0,00

PR #116, opened 2020-11-09
New approach for the NGINX notice

alexmigf:


includes/class-wcpdf-settings-debug.php
```diff
@@ -36,101 +36,133 @@ public function debug_tools( $tab, $section ) {
...
+						// check permissions
+						if ( !check_admin_referer( 'wpo_wcpdf_debug_tools_action', 'security' ) ) {
+							return;
 						}
+
+						WPO_WCPDF()->main->generate_random_string();
+						$old_path = WPO_WCPDF()->main->get_tmp_base( false );
+						$new_path = WPO_WCPDF()->main->get_tmp_base();
+						WPO_WCPDF()->main->copy_directory( $old_path, $new_path );
+						printf('<div class="notice notice-success"><p>%s</p></div>', __( 'Random string created!', 'woocommerce-pdf-invoices-packing-slips' ) );
```
Spreeuw, line 57: I would be more explicit here:
```php
sprintf( __( 'Temporary folder moved to `%s`', 'woocommerce-pdf-invoices-packing-slips' ) ), $new_path );
```

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -333,20 +333,34 @@ public function nginx_detected()
 		}
 		$tmp_path = $this->main->get_tmp_path('attachments');
 		$server_software   = isset( $_SERVER['SERVER_SOFTWARE'] ) ? sanitize_text_field( wp_unslash( $_SERVER['SERVER_SOFTWARE'] ) ) : false;
```
Spreeuw, line 335: If you're going to perform `stristr`  on this it's better to fall back to `''` instead of `false`.

Spreeuw (changes requested):
Some minor adjustments and I think we're good to go here!

includes/class-wcpdf-settings-debug.php
```diff
@@ -54,7 +54,7 @@ public function debug_tools( $tab, $section ) {
 						$old_path = WPO_WCPDF()->main->get_tmp_base( false );
 						$new_path = WPO_WCPDF()->main->get_tmp_base();
 						WPO_WCPDF()->main->copy_directory( $old_path, $new_path );
-						printf('<div class="notice notice-success"><p>%s</p></div>', __( 'Random string created!', 'woocommerce-pdf-invoices-packing-slips' ) );
+						sprintf( __( 'Temporary folder moved to `%s`', 'woocommerce-pdf-invoices-packing-slips' ), $new_path );
```
Spreeuw, line 57: you removed too much :octocat:  only `__()` should have been replaced by the `sprintf()`. Now there is no notice at all anymore.

Spreeuw (changes requested):

Spreeuw (approved):
Looks good to me!

PR #124, opened 2020-12-02
Adds notice if MailPoet MTA method selected is MailPoet and asks to c…

### LINKED ISSUE #123: Mailpoet compatibility notice
https://wordpress.org/support/topic/no-pdf-in-emails/

This is a recurring issue, covered in the Mailpoet documentation:
https://kb.mailpoet.com/article/292-choose-how-to-send-your-wordpress-websites-emails#attachments

> **Attachments are not permitted**
> When sending transactional emails with MailPoet 3 and the active sending method is MailPoet Sending Service, MailPoet does not include attachments in the emails that are sent because of bandwidth concerns and security implications with spam filters.
> We recommend to send links to files instead of attaching them to the email. If you need to send attachment for your WooCommerce or, disable this feature in MailPoet.
> 
> Meaning this option should be set to "**The default WordPress sending method**" (**NOT** selected in this screenshot):
> file-rXbF4BzOcB

If we can easily detect Mailpoet users that have this Mailpoet method selected, we should show a notice in the backend

alexmigf:
…hange

closes #123


woocommerce-pdf-invoices-packingslips.php
```diff
@@ -384,6 +385,56 @@ public function nginx_detected()
 		}
 	}
 
+	/**
+	 * Detect MailPoet.
+	 * @return void
+	 */
+	public function mailpoet_mta_detected() {
+		$mailpoet = $this->namespace_exists( 'MailPoet' );
```
Spreeuw, line 393: What's the reason you're checking the namespace? I think you can directly move to the `is_callable` check:
```php
if( is_callable( array( '\\MailPoet\\Settings\\SettingsController', 'getInstance' ) ) ) {
```
with that you don't need `class_exists` either.

alexmigf: Fixed and tested.

PR #129, opened 2021-01-07
Replaces the old DOMPDF version 0.8.3 by the new 1.0.1 with PHP8 support

### LINKED ISSUE #126: PHP8 compatibility
The base plugin is PHP8.0 ready, but dompdf 0.8.3 (the currently bundled version) is not.

The reason we are not using 0.8.6 (latest release at the time of writing) is because it does not support PHP5.6 and PHP7.0, whereas 0.8.3 does. If we want to move forward and support PHP8.0, this means we have to drop support for PHP < 7.1.

I would like to propose a dual approach:

- Bundle the upcoming dompdf 1.0.0 (which will have PHP8.0 support) by default, but **only load it on PHP7.1+**. Creating PDF should be disabled elegantly. (e.g. no crashes/fatal errors, it should simply not try to create PDFs).
- Show an admin notice for sites using PHP < 7.1, linking to a wordpress.org plugin that replaces the bundled library the current 0.8.3 dompdf library
- The 0.8.3 dompdf plugin should show a notice that it's no longer needed as soon as the site has PHP7.1+

According to our plugin description we support PHP5.3 but David Anderson recently reported that is a false statement:

> I noticed, though, that even 0.8.3 says that it requires PHP 5.4+. Out of curiosity, I ran your current release through syntax checks on PHP 5.3 ...
> 
> ```
> # php -v
> PHP 5.3.3 (cli) (built: Nov 1 2019 12:28:08)
> Copyright (c) 1997-2010 The PHP Group
> Zend Engine v2.3.0, Copyright (c) 1998-2010 Zend Technologies
> 
> # find . -name '*\.php' -exec php -l {} \; | grep -v vendor/ | grep -v 'No syntax errors detected'
> PHP Fatal error: Can't use function return value in write context in ./includes/class-wcpdf-frontend.ph
> p on line 100
> PHP Fatal error: Can't use function return value in write context in ./includes/class-wcpdf-settings-do
> cuments.php on line 35
> PHP Fatal error: Can't use method return value in write context in ./includes/class-wcpdf-admin.php on
> line 375
> PHP Parse error: syntax error, unexpected '.', expecting ')' in ./vendor/composer/autoload_static.php o
> n line 23
> ```
> 
> Some of those errors are ones I recognise as being code that requires PHP 5.5+, so I ran it through 5.4 as well:
> 
> ```
> $ php54 -v
> PHP 5.4.45 (cli) (built: Oct 22 2019 13:23:07)
> Copyright (c) 1997-2014 The PHP Group
> Zend Engine v2.4.0, Copyright (c) 1998-2014 Zend Technologies
> $ find . -name '*\.php' -exec php54 -l
> {} \; | grep -v vendor/ | grep -v 'No syntax errors detected'
> PHP Parse error: syntax error, unexpected '.', expecting ')' in ./vendor/composer/autoload_st
> atic.php on line 23
> Parse error: syntax error, unexpected '.', expecting ')' in ./vendor/composer/autoload_static.
> php on line 23
> PHP Fatal error: Can't use function return value in write context in ./includes/class-wcpdf-f
> rontend.php on line 100
> Fatal error: Can't use function return value in write context in ./includes/class-wcpdf-fronte
> nd.php on line 100
> PHP Fatal error: Can't use method return value in write context in ./includes/class-wcpdf-adm
> in.php on line 375
> Fatal error: Can't use method return value in write context in ./includes/class-wcpdf-admin.ph
> p on line 375
> PHP Fatal error: Can't use function return value in write context in ./includes/class-wcpdf-s
> ettings-documents.php on line 35
> Fatal error: Can't use function return value in write context in ./includes/class-wcpdf-settin
> gs-documents.php on line 35
> ```
> 
> If your autoload_static.php is getting run, it requires at least PHP 5.6:
> 
> ```
> $ php55 -v
> PHP 5.5.38 (cli) (built: Oct 22 2019 14:10:44)
> Copyright (c) 1997-2015 The PHP Group
> Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
> $ find . -name '*\.php' -exec php55 -l
> {} \; | grep -v vendor/ | grep -v 'No syntax errors detected'
> PHP Parse error: syntax error, unexpected '.', expecting ')' in ./vendor/composer/autoload_st
> atic.php on line 23
> Parse error: syntax error, unexpected '.', expecting ')' in ./vendor/composer/autoload_static.
> php on line 23
> ```

This means that in practice we already require PHP5.6 so the description should be updated to reflect this too.
The three `Can't use function return value in write context` errors can be fixed easily, but the `autoload_static.php` errors are harder. I don't know if/when this gets executed though.

Regardless, dropping 5.3 and later 5.6 support entirely is certainly not a bad thing.

alexmigf:
closes #126


composer.json
```diff
@@ -1,6 +1,11 @@
...
-    "dompdf/dompdf": "*"
-  }
+	"repositories": [
+		{
+		"type": "git",
+		"url": "https://github.com/dompdf/dompdf.git"
+		}
+	],
+	"require": {
+		"dompdf/dompdf": "dev-master#45d260f2f6c6a542e0db3459470374c986a46d93"
```
Spreeuw, line 9: why use dev-master instead of the official 1.0.1 release? (or the soon-to-be-released 1.0.2)

includes/class-wcpdf-install.php
```diff
@@ -61,8 +61,8 @@ public function do_install() {
 	 * Plugin install method. Perform any installation tasks here
 	 */
 	protected function install() {
-		// only install when php 5.3 or higher
-		if ( version_compare( PHP_VERSION, '5.3', '<' ) ) {
+		// only install when php 7.1 or higher
```
Spreeuw, line 64: Plugin should install from 5.6+, otherwise the people that use the 0.8.3 extension miss out on the installation  procedure

includes/class-wcpdf-install.php
```diff
@@ -185,8 +185,8 @@ protected function install() {
 	 * @param string $installed_version the currently installed ('old') version
 	 */
 	protected function upgrade( $installed_version ) {
-		// only upgrade when php 5.3 or higher
-		if ( version_compare( PHP_VERSION, '5.3', '<' ) ) {
+		// only upgrade when php 7.1 or higher
```
Spreeuw, line 188: same as install: should be 5.6+

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -273,11 +272,17 @@ public function need_woocommerce() {
 	/**
 	 * PHP version requirement notice
 	 */
-	
 	public function required_php_version() {
-		$error = __( 'WooCommerce PDF Invoices & Packing Slips requires PHP 5.3 or higher (5.6 or higher recommended).', 'woocommerce-pdf-invoices-packing-slips' );
-		$how_to_update = __( 'How to update your PHP version', 'woocommerce-pdf-invoices-packing-slips' );
-		$message = sprintf('<div class="error"><p>%s</p><p><a href="%s">%s</a></p></div>', $error, 'http://docs.wpovernight.com/general/how-to-update-your-php-version/', $how_to_update);
+		$message = sprintf(
```
Spreeuw, line 276: Required PHP version is 5.6, 7.1+ recommended. Maybe make the notice about the addon separate? Or at least conditional for 5.6+

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -273,11 +272,17 @@ public function need_woocommerce() {
...
+			'<div class="error"><p>%s</p><p>%s <a href="%s">%s</a> %s <a href="%s">%s</a>.</p></div>',
+			__( 'WooCommerce PDF Invoices & Packing Slips requires PHP 7.1 or higher (7.1 or higher recommended).', 'woocommerce-pdf-invoices-packing-slips' ),
+			__( 'You should', 'woocommerce-pdf-invoices-packing-slips' ),
+			'http://docs.wpovernight.com/general/how-to-update-your-php-version/',
+			__( 'update your PHP version', 'woocommerce-pdf-invoices-packing-slips' ),
+			__( 'or', 'woocommerce-pdf-invoices-packing-slips' ),
```
Spreeuw, line 282: I think you should take a different approach here because translating parts of sentences may not work that well in all languages. You could also try:
```
You should %supdate your PHP version%s or %sdownload and install this add-on%s to enable PHP5.6 support
```
where you insert the `<a>` tags + url at the %s
or separated:
```
We strongly recommend to %supdate your PHP version%s.
If you cannot upgrade your PHP version you can download %sthis addon%s to enable backwards compatibility with PHP5.6
```

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -273,11 +272,17 @@ public function need_woocommerce() {
...
-	
 	public function required_php_version() {
-		$error = __( 'WooCommerce PDF Invoices & Packing Slips requires PHP 5.3 or higher (5.6 or higher recommended).', 'woocommerce-pdf-invoices-packing-slips' );
-		$how_to_update = __( 'How to update your PHP version', 'woocommerce-pdf-invoices-packing-slips' );
-		$message = sprintf('<div class="error"><p>%s</p><p><a href="%s">%s</a></p></div>', $error, 'http://docs.wpovernight.com/general/how-to-update-your-php-version/', $how_to_update);
+		$message = sprintf(
+			'<div class="error"><p>%s</p><p>%s <a href="%s">%s</a> %s <a href="%s">%s</a>.</p></div>',
+			__( 'WooCommerce PDF Invoices & Packing Slips requires PHP 7.1 or higher (7.1 or higher recommended).', 'woocommerce-pdf-invoices-packing-slips' ),
+			__( 'You should', 'woocommerce-pdf-invoices-packing-slips' ),
+			'http://docs.wpovernight.com/general/how-to-update-your-php-version/',
```
Spreeuw, line 280: should be `https`

Spreeuw (changes requested):
Technically we should distinguish between requirement (5.6) and recommended (7.1) PHP version

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -273,17 +273,18 @@ public function need_woocommerce() {
...
-		);
-	
+		$error_message	= __( 'WooCommerce PDF Invoices & Packing Slips requires PHP 5.6 or higher (7.1 or higher recommended).', 'woocommerce-pdf-invoices-packing-slips' );
+		$php_message	= __( 'We strongly recommend to %supdate your PHP version%s.', 'woocommerce-pdf-invoices-packing-slips' );
+		$add_on_message	= __( 'If you cannot upgrade your PHP version, you can download %sthis addon%s to enable backwards compatibility with PHP5.6.', 'woocommerce-pdf-invoices-packing-slips' );
+
+		$message = '<div class="error">';
+		$message .= sprintf( '<p>%s</p>', $error_message );
+		$message .= sprintf( '<p>'.$php_message.'</p>', '<a href="https://docs.wpovernight.com/general/how-to-update-your-php-version/" target="_blank">', '</a>' );
+		if ( version_compare( PHP_VERSION, '5.6', '<' ) ) {
```
Spreeuw, line 283: 5.6 is NOT supported at all. The addon is for 5.6-7.1. So should be `>` I think?

alexmigf: Sorry, forgot that

Spreeuw: Actually, 5.6 is supported too I meant below 5.6 is not supported

Spreeuw: so should be `>=`

Spreeuw: (but maybe that doesn't make a big difference because 5.6.1 would be greater than 5.6 already)

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -273,12 +272,19 @@ public function need_woocommerce() {
...
-	
+		$error_message	= __( 'WooCommerce PDF Invoices & Packing Slips requires PHP 5.6 or higher (7.1 or higher recommended).', 'woocommerce-pdf-invoices-packing-slips' );
+		$php_message	= __( 'We strongly recommend to %supdate your PHP version%s.', 'woocommerce-pdf-invoices-packing-slips' );
+		$add_on_message	= __( 'If you cannot upgrade your PHP version, you can download %sthis addon%s to enable backwards compatibility with PHP5.6.', 'woocommerce-pdf-invoices-packing-slips' );
+
+		$message = '<div class="error">';
+		$message .= sprintf( '<p>%s</p>', $error_message );
+		$message .= sprintf( '<p>'.$php_message.'</p>', '<a href="https://docs.wpovernight.com/general/how-to-update-your-php-version/" target="_blank">', '</a>' );
+		if ( version_compare( PHP_VERSION, '5.6', '>' ) ) {
+			$message .= sprintf( '<p>'.$add_on_message.'</p>', '<a href="https://github.com/wpovernight/wcpdf-dompdf-0.8.3/releases" target="_blank">', '</a>' );
```
Spreeuw, line 284: this links to a page that says:

> There aren’t any releases here

Maybe we should even provide a direct link to the release zip instead?

Spreeuw: Thinking more about this, I think we should actually link to a page in our documentation instead, where we can always link to the latest version and provide more detailed instructions on installing the extension. We need to make this transition as smooth as possible.

PR #131, opened 2021-01-15
Set current screen on Wizard

### LINKED ISSUE #130: No screen object set on Wizard
Plugins that check the current screen via hooks that are present in the Wizard like `admin_footer` will crash because the screen object has not been set. For instance returning:

`Error: Call to a member function is_block_editor() on null`

First mentioned in the following ticket: https://secure.helpscout.net/conversation/1393863690/27954?folderId=221726

Terminator-Barbapapa:
Closes #130

### DIFF
--- a/includes/class-wcpdf-setup-wizard.php
+++ b/includes/class-wcpdf-setup-wizard.php
@@ -37,6 +37,11 @@ class Setup_Wizard {
 		if ( empty( $_GET['page'] ) || 'wpo-wcpdf-setup' !== $_GET['page'] ) {
 			return;
 		}
+
+		if ( is_null ( get_current_screen() ) ) {
+			set_current_screen();
+		}
+				
 		$this->steps = array(
 			'shop-name' => array(
 				'name'	=> __( 'Shop Name', 'woocommerce-pdf-invoices-packing-slips' ),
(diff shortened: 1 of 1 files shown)


alexmigf (commented):
Shouldn't you set that only if returns null for the screen?

```php
if ( is_null ( get_current_screen() ) ) {
	set_current_screen();
}
```

Shouldn't be better to place that inside the `setup_wizard()` method? Right after this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5f96c5580769c1b0950d938a18b6393777ab2c99/includes/class-wcpdf-setup-wizard.php#L37-L39

Terminator-Barbapapa:
At the moment it is an exact copy of how WooCommerce used to handle its setup through a wizard (now deprecated): https://github.com/woocommerce/woocommerce/blob/f9eb0727d58994e6abf7c53ae91b0bf75bcee32c/includes/admin/class-wc-admin-setup-wizard.php#L339

But I see no harm in moving it up to the `setup_wizard()` function and adding an additional check. Agreed @Spreeuw?

Spreeuw:
Good for me 👍

PR #134, opened 2021-01-25
New date formatting function wcpdf_date_format()

### LINKED ISSUE #132: Replace all wc_date_format() with custom function
@Spreeuw said:

> Maybe we should introduce our own date formatting function for easier filtering, where we can pass a `$type` variable. Then we replace all our `wc_date_format( $date )` calls with `wcpdf_date_format( $date, $type )`  with a filter that includes both arguments.

https://wpovernight.slack.com/archives/C0121KYKGN6/p1611576680007100

Ticket: https://secure.helpscout.net/conversation/1405634973/28262?folderId=221726

alexmigf:
closes #132

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -400,7 +400,7 @@ class Admin {
 							<p class="form-field form-field-wide">
 								<p>
 									<span><strong><?php _e( 'Invoice Date:', 'woocommerce-pdf-invoices-packing-slips' ); ?></strong></span>
-									<span><?php if (!empty($invoice_date)) echo $invoice_date->date_i18n( wcpdf_date_format( $invoice, 'invoice_date' ).' @ '.wc_time_format() ); ?></span>
+									<span><?php if (!empty($invoice_date)) echo $invoice_date->date_i18n( wc_date_format().' @ '.wc_time_format() ); ?></span>
 								</p>
 							</p>
 						</div>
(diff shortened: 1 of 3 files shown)


includes/class-wcpdf-admin.php
```diff
@@ -400,7 +400,7 @@ public function data_input_box_content ( $post ) {
 							<p class="form-field form-field-wide">
 								<p>
 									<span><strong><?php _e( 'Invoice Date:', 'woocommerce-pdf-invoices-packing-slips' ); ?></strong></span>
-									<span><?php if (!empty($invoice_date)) echo $invoice_date->date_i18n( wc_date_format().' @ '.wc_time_format() ); ?></span>
+									<span><?php if (!empty($invoice_date)) echo $invoice_date->date_i18n( wcpdf_date_format( $invoice, 'invoice_date' ).' @ '.wc_time_format() ); ?></span>
```
Spreeuw, line 403: This is perhaps not an ideal use case for this function because of the usage of `wc_time_format` here. It's backend so maybe we don't need it here?

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -411,7 +411,7 @@ public function order_notes( $filter = 'customer', $include_system_notes = true
 	 * Return/Show the current date
 	 */
 	public function get_current_date() {
-		return apply_filters( 'wpo_wcpdf_date', date_i18n( wc_date_format() ) );
+		return apply_filters( 'wpo_wcpdf_date', date_i18n( wcpdf_date_format( null, 'current_date' ) ) );
```
Spreeuw, line 414: why are you not passing `$this` here? (same for order date)

Spreeuw (approved):

PR #135, opened 2021-01-26
Create universal document data functions for meta box

### LINKED ISSUE #133: Create universal document data functions
Third party documents as well as our own Professional extension documents currently need to 'reinvent the wheel' by writing their own code to create all the input fields and buttons to be shown on the "PDF Document data" tab.

If we use the code from `\WPO\WC\PDF_Invoices_Pro\Writepanels::output_number_date_edit_fields()` in the Professional extension and update it to the standards currently applied in rendering the Invoice fields, this can be reused by other plugins.

To facilitate this reuse, we can include the class instance as a second parameter in the `wpo_wcpdf_meta_box_end` action, making it super easy to call the function. Likewise, we can add an additional action inside the `save_invoice_number_date()` method, passing the required data as well as the class instance.

Some fields (like document notes etc.) should be made optional in the field rendering function.

We can use our own Professional extension as a test case.

alexmigf:
closes #133

### DIFF
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -89,22 +89,12 @@ jQuery(document).ready(function($) {
 		$(this).addClass('wcpdf-regenerate-spin');
 		$form = $(this).closest('.wcpdf-data-fields');
 
-		// create an object with the form inputs data
-		form_inputs_data = {};
-		$form.find(':input').each( function() {
-			if (!$(this).is(':disabled')) {
-				var name = $(this).attr("name");
-				name = name.replace(/[[0-9]+]/, ''); // for credit-note array []
-				value = $(this).val();
-				form_inputs_data[name] = value;
-			}
-		} );
-
-		// convert data to json string
-		form_data_json = JSON.stringify( form_inputs_data );
-
 		// create an object with the data attributes
 		form_data_attributes = $form.data();
+
+		// create a serialized string with the form inputs name/data and removes the [id] part globally in the string with regex expression
+		regex      = new RegExp("%5B"+$form.data('order_id')+"%5D", 'g');
+		serialized = $form.find(":input:not(:disabled)").serialize().replace(regex, '');
 
 		// Make sure all feedback icons are hidden before each call
 		$form.find('.document-action-success, .document-action-failed').hide();
@@ -114,7 +104,7 @@ jQuery(document).ready(function($) {
 			data: {
 				action:         'wpo_wcpdf_regenerate_document',
 				security:       $(this).data('nonce'),
-				form_data:      form_data_json,
+				form_data:      serialized,
 				order_id:       form_data_attributes.order_id,
 				document_type:  form_data_attributes.document,
 			},
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -31,6 +31,9 @@ jQuery(document).ready(function($) {
 
 	$('#wpo_wcpdf-data-input-box').insertAfter('#woocommerce-order-data');
 
+	// disable inputs from start
+	$( '.wcpdf-data-fields' ).find( ':input' ).attr( 'disabled', true );
+	
 	// enable invoice number edit if user initiated
 	$( ".wpo-wcpdf-set-date-number, .wpo-wcpdf-edit-date-number, .wpo-wcpdf-edit-document-notes" ).click(function() {
 		$form = $(this).closest('.wcpdf-data-fields-section');
@@ -42,11 +45,11 @@ jQuery(document).ready(function($) {
 		if( $form.find(".read-only").is(":visible") ) {
 			$form.find(".read-only").hide();
 			$form.find(".editable").show();
-			$form.find(':input').prop('disabled', false);
+			$form.find(':input').attr('disabled', false);
 		} else {
 			$form.find(".read-only").show();
 			$form.find(".editable").hide();
-			$form.find(':input').prop('disabled', true);
+			$form.find(':input').attr('disabled', true);
 		}
 	});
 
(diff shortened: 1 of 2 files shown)


Spreeuw:
I wonder if this still works now that date is an array?
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/64b7518c37d2c5f132c5f3ce718c4ef2c807d9d4/assets/js/order-script.js#L94-L101

assets/js/order-script.js
```diff
@@ -93,8 +93,8 @@ jQuery(document).ready(function($) {
 		form_inputs_data = {};
 		$form.find(':input').each( function() {
 			if (!$(this).is(':disabled')) {
-				name = $(this).attr("name");
-				name = name.split('[', 1)[0]; // for credit-note array []
+				var name = $(this).attr("name");
+				name = name.replace(/[[0-9]+]/, ''); // for credit-note array []
```
Spreeuw, line 97: This does not work, because instead of passing an array for the date, it will use the same name for each field and so instead of for example:
```JSON
{
	"_wcpdf_invoice_number":"1146",
	"_wcpdf_invoice_date": {
		"date":"2021-01-14",
		"hour":"10",
		"minute":"04"
	}
}
```
it is currently passing:
```
{
	"_wcpdf_invoice_number":"1146",
	"_wcpdf_invoice_date[date]":"2021-01-14",
	"_wcpdf_invoice_date[hour]":"10",
	"_wcpdf_invoice_date[minute]":"04"
}
```
Which is not parsed correctly and so the date is not updated with the regeneration data

Spreeuw: I think you could make it easier for yourself by doing the parsing at the PHP end, with just this on the JS end:
```JS
form_inputs_data = $form.serialize();
```
and then on the PHP end:
```PHP
parse_str($_POST['form_data'], $form_data);
```

alexmigf: What you think of this?

```js
		// create an object with the form inputs data
		var form_inputs_data = {};
		var c = 0;
		$form.find(':input').each( function() {
			if (!$(this).is(':disabled')) {
				var name = $(this).attr("name");
				name     = name.replace(/[[0-9]+]/, ''); // removes the ID, eg. [198]
				var key  = name.match(/[^[\]]+(?=\])/g); // extracts the keys (date, hour, minute) to a new variable
				name     = name.split('[', 1)[0];        // removes the remaining square brackets content: [date], [hour], [minute]

				var value = $(this).val();

				if (key) {
					c++;
					if( c == 1 ) {
						form_inputs_data[name] = {};
					}
					form_inputs_data[name][key[0]] = value;
				} else {
					form_inputs_data[name] = value;
				}
			}
		} );
```

If you prefer the PHP way let me know.

Spreeuw: seems an unnecessary amount of trouble for something that can be done with a single line of code ;)

Spreeuw (changes requested):
# Backwards compatibility for regenerate
Non-core documents (such as Proforma and Credit Note) where the fields are from older versions will still be using the old `_wcpdf_credit_note_date` / `_wcpdf_credit_note_date_hour` `_wcpdf_credit_note_date_minute` fields. They handle their own updates upon saving, but with the regenerate function, this is handled by core. We should build a fallback that checks if `_wcpdf_credit_note_date` is an array, and if not, convert the old fields to the array format before further processing.

# Notes
Notes seem to behave strangely, but comparing to the current release I think that it already broke before these changes.

If I enter this:
```
\\ %20
\n
/
"text"
<div>some html</div>
``` 

I get different results depending on whether I save via "Regenerate" or via saving the order:
**Regenerate:**
```
\\ <br />
\n<br />
/<br />
"text"<br />
some html
```
**save:**
```
\\\\ <br />
\\n<br />
/<br />
\"text\"<br />
some html
```

I believe we discussed this before so I don't know if this is a regression or if we decided to always convert to `<br>`, but some of this is still buggy, also the fact that it's not handled the same between saving and regenerating.

# Order ID tagged data
I realize that we have already discussed this to death so apologies for opening this up again...
But I was thinking about the JS solution and realized that it also affects the values themselves (in particular the notes!). I think it's safer to do this at the PHP end after all. So passing the serialized data as JS provides it, then in PHP:
```PHP
if ( is_array( $form_data ) ) {
	foreach ( $form_data as $key => &$value ) {
		if ( is_array( $value ) && !empty( $value[$order_id] ) ) {
			$value = $value[$order_id];
		}
	}
}
```

# Form data passed
This is also an existing bug, but we should only pass data that is being edited. So if none of the edit fields are open, it should regenerate the document without passing any number/date/note data. This is already done properly with the notes field, (only passed when enabled), but if I click the edit button twice (open > close) for both of them and then clicking regenerate, it throws an error ('incomplete request') because the form data is empty. It should still generate.

To resolve this, we should either check on the JS end if the input is visible (so `":input:visible:not(:disabled)"`) or prevent this in the first place by making the non-visible input fields disabled from the start (which I think we should definitely do to prevent the invoice number from regenerating upon order saving!).

includes/class-wcpdf-admin.php
```diff
@@ -725,11 +770,25 @@ public function regenerate_document() {
...
+					$value = $value[$order_id];
+				}
+			}
+		}
 
 		try {
 			$document = wcpdf_get_document( $document_type, wc_get_order( $order_id ) );
 			if ( !empty($document) && $document->exists() ) {
+				// perform legacy date fields replacements check
+				if( in_array( $document->get_type(), array( 'proforma', 'credit-note' ) ) && ! is_array( $form_data["_wcpdf_{$document->slug}_date"] ) ) {
```
Spreeuw, line 788: this shouldn't be limited to proforma and credit-note, could be third party documents as well. It should also have a check for whether `_wcpdf_{$document->slug}_date` isset at all (form data could be empty just regenerating).

assets/js/order-script.js
```diff
@@ -80,6 +80,21 @@ jQuery(document).ready(function($) {
 		});
 	});
 
+	// disable inputs from start
+	$( '.wcpdf-data-fields' ).find( ':input' ).attr( 'disabled', true );
```
Spreeuw, line 84: any idea what's different here compared to what you are already setting server side? because you already set it to disabled there too 🤔 (which should be enough already).
Because I don't see anything different anymore if I'm comparing. If I disabled this whole code block it works fine so I wonder if I was testing incorrectly... Or is it because you added `:visible` to the selector used for serialization?

Spreeuw (changes requested):
I can no longer replicate the issue I had with data being passed even if the fields are not enabled... So that part appears to have been working properly already.
It should still be possible the regenerate a document without passing data though + a few other minor tweaks.

Spreeuw (approved):
Looks good to me! 🚀

PR #144, opened 2021-02-05
allow filtering button classes using 

Spreeuw:
using `wpo_wcpdf_action_button_class`  the appearance of the button can now be controlled more easily. Here's an example implementation showing a grey checkmark for created documents and a green checkmark for printed ones:
```php
add_action( 'wpo_wcpdf_before_document', function( $document_type, $order ) {
	$order->update_meta_data( "wpo_wcpdf_{$document_type}_last_printed", date_i18n( 'Y-m-d H:i:s' ) );
	$order->save_meta_data();
}, 10, 2 );

add_filter( 'wpo_wcpdf_action_button_class', function( $class, $document ) {
	if ( $order = $document->order ) {
		$document_type = $document->get_type();
		$printed = $order->get_meta( "wpo_wcpdf_{$document_type}_last_printed" );
		if ( !empty( $printed ) ) {
			$class .= " printed";
		}
	}
	return $class;
}, 10, 2 );

add_action('admin_enqueue_scripts', function(){
	wp_add_inline_style( 'wpo-wcpdf-order-styles', '.wpo_wcpdf.button.exists:not(.printed):after { color: grey !important; }' );
}, 99999 );
```

result:
button-class-filter-result

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -182,19 +182,24 @@ class Admin {
 			$icon = !empty($document->icon) ? $document->icon : WPO_WCPDF()->plugin_url() . "/assets/images/generic_document.png";
 			if ( $document = wcpdf_get_document( $document->get_type(), $order ) ) {
 				$document_title = is_callable( array( $document, 'get_title' ) ) ? $document->get_title() : $document_title;
+				$document_exists = is_callable( array( $document, 'exists' ) ) ? $document->exists() : false;
 				$listing_actions[$document->get_type()] = array(
-					'url'		=> wp_nonce_url( admin_url( "admin-ajax.php?action=generate_wpo_wcpdf&document_type={$document->get_type()}&order_ids=" . WCX_Order::get_id( $order ) ), 'generate_wpo_wcpdf' ),
-					'img'		=> $icon,
-					'alt'		=> "PDF " . $document_title,
-					'exists'	=> is_callable( array( $document, 'exists' ) ) ? $document->exists() : false,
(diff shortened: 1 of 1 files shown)


alexmigf:
@Spreeuw you don't want to implement this? The grey and green checks? I think is a good idea.

Spreeuw:
I wasn't planning to... I don't know if it's clear what it means. @Terminator-Barbapapa thoughts?

Terminator-Barbapapa:
I think it will cause confusion. The difference is too subtle and it may lead to a user thinking the document was created when in fact it was not.

alexmigf:
@Terminator-Barbapapa I believe it created is in both situations, just the green indicates you clicked (printed) right @Spreeuw ?

Terminator-Barbapapa:
@alexmigf I see now. I'm sorry, I thought it was meant to indicate the difference between if a document was created or not. But a change in color might not be enough to explain what is happening to the user. How will they know what each color means?

alexmigf:
A tooltip?

Spreeuw:
> I thought it was meant to indicate the difference between if a document was created or not. But a change in color might not be enough to explain what is happening to the user. How will they know what each color means?

The above example was just something I created as an example for a customer that wanted to see which documents were downloaded by the admin. So I added order meta when that happened.

In the example:

- no checkmark = no document
- gray checkmark = document has been created
- green checkmark = document has been downloaded by the admin

YordanSoares:
Another issue to consider is accessibility. There are several types of color blindness in which there is no distinction between green and gray, so @alexmigf proposal would help people who have this condition.

Spreeuw:
Maybe the question of implementing this by default should be a separate issue, as the filter can be added regardless, I think. @alexmigf did you review the code/implementation itself?

alexmigf (approved):
Works fine!

PR #153, opened 2021-02-17
Create store tables appending the year

### LINKED ISSUE #149: Create new store tables in DB for each year if reset setting enabled
@Spreeuw said:

> It seems a fair number of people have recently reported issues with the yearly number reset. It's a tough problem to tackle because there is no straight forward way to reliably handle sequential numbers for previous years. Last year we introduced an 'easier' setting that allows people to set the invoice date to the order date (previously the only way to do that was via a filter), and this is what's causing the issues because the generation of invoices is no longer truly sequential. So people could generate invoices for this year, then invoice for last year (triggering a reset) then invoices for this year again (triggering another reset, etc.)
This afternoon I was thinking about possible solutions and I wonder if making year-based store-tables is a solution here..
In this case, the plugin would suffix the store name (e.g. invoice_number ) with the year of the document ( so invoice_number_2021 for an invoice from 2021) As a result, rather than resetting the current store, it would simply move to a new, empty store (automatically!) as soon as the new year starts. And then, when an invoice is made for last year, it simply appends a row to last year's store (invoice_number_2020)
This will probably require a DB migration, but may be the most reliable option that we have.

> Migration shouldn't be too difficult, basically we only need to rename the table upon upgrading.
This will need thorough testing of course since it's such a vital part of the plugin... We also need to think about the scenario where people switch on or switch off the reset option and how this is handled internally (keeping the existing 'next invoice number' in tact).

Discussion: https://wpovernight.slack.com/archives/C0121KYKGN6/p1613149015029400

alexmigf:
closes #149

#### Main test

* 1: Have legacy table created and see if gets renamed :heavy_check_mark: 
* 2: Having the new year table, simulate a new year change with:
    * 2.1: Reset numeration on, should start with 1 on the new table :heavy_check_mark: 
    * 2.2: Reset numeration off, should continue the prior numeration on the new table :heavy_check_mark: 

#### Others

* 3: Re-test the above using the filter `wpo_wcpdf_document_sequential_number_store`
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: 
    * 2.2: :heavy_check_mark: 
* 4: Re-test the above using the filter `wpo_wcpdf_{$this->slug}_number_store_name`
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark:  (resets monthly)
    * 2.2: :heavy_check_mark:  (resets monthly)
* 5: Re-test the above using the filter `wpo_wcpdf_number_store_table_name`
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: 
    * 2.2: :heavy_check_mark: 
* 6: Test with orders from last year
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: 
    * 2.2: :heavy_check_mark: 
* 7: In two 2020 orders set "Invoice date" to "order date" and generate Invoices
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: (keeps using the 2020 store, when the year changes it resets)
    * 2.2: :heavy_check_mark: (uses the last store in use, and continues the same numeration on year reset)
* 8: In two 2021 orders, set "Invoice date" to "order date" and generate Invoices
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: (keeps using the 2021 store, when the year changes it resets)
    * 2.2: :heavy_check_mark: (uses the last store in use, and continues the same numeration on year reset)
* 9: Test number stores migration on install :heavy_check_mark: 

#### Used filters

3:
```php
add_filter( 'wpo_wcpdf_document_sequential_number_store' , function( $number_store_name, $document ) {
	return 'test_' . $number_store_name;
}, 10, 2 );
```
4:
```php
// resets number store monthly
add_filter( 'wpo_wcpdf_invoice_number_store_name', function( $number_store_name, $store_base_name, $date, $method, $document ) {
	$year  = $date->date_i18n( 'Y' );
	$month = $date->date_i18n( 'm' );
	return "test_{$month}_{$year}";
}, 10, 5 );
```
5:
```php
add_filter( 'wpo_wcpdf_number_store_table_name', function( $table_name, $store_name, $method ) {
	global $wpdb;
	return "{$wpdb->base_prefix}_wcpdf_{$store_name}";
}, 10, 3 );
```


includes/class-wcpdf-install.php
```diff
@@ -347,6 +347,17 @@ protected function upgrade( $installed_version ) {
 			update_option( 'wpo_wcpdf_settings_debug', $debug_settings );
 		}
 
+		// 2.8.2: rename invoice number store table
+		if ( $installed_version == 'versionless' || version_compare( $installed_version, '2.8.2', '<=' ) ) {
+			global $wpdb;
+			$legacy_table_name = "{$wpdb->prefix}wcpdf_invoice_number";
+			$current_year      = date('Y');
+			$new_table_name    = "{$wpdb->prefix}wcpdf_invoice_number_{$current_year}";
```
Spreeuw, line 355: table name could have been filtered:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/05efc0fa692a0a75a17645744d6c1e671239f3aa/includes/documents/class-wcpdf-sequential-number-store.php#L42
We should include this scenario in tests, looking up code snippets that we have provided to users in the past.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,50 @@ public function init_number() {
...
-		$number_store = new Sequential_Number_Store( $number_store_name, $number_store_method );
-		// reset invoice number yearly
-		if ( isset( $this->settings['reset_number_yearly'] ) ) {
-			$current_year = date("Y");
-			$last_number_year = $number_store->get_last_date('Y');
-			// check if we need to reset
-			if ( $current_year != $last_number_year ) {
-				$number_store->set_next( apply_filters( 'wpo_wcpdf_reset_number_yearly_start', 1, $this ) );
-			}
+		if( ! empty( $document_date = $this->get_date() ) ) {
```
Spreeuw, line 106: `set_date()` is called before `init_number()`:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/05efc0fa692a0a75a17645744d6c1e671239f3aa/includes/documents/class-wcpdf-invoice.php#L68-L74

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,50 @@ public function init_number() {
...
+			$number_store_method = WPO_WCPDF()->settings->get_sequential_number_store_method();
+			$number_store_name   = $this->store_name( $document_date );
```
Spreeuw, line 108: I think this could be simplified by renaming the method `get_sequential_number_store()` and letting it return the instance. You don't need to pass `$document_date` either since it's an internal property.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,50 @@ public function init_number() {
...
+			$this->set_number( $invoice_number );
+
+			return $invoice_number;
 		}
+	}
 
-		$invoice_date = $this->get_date();
-		$invoice_number = $number_store->increment( $this->order_id, $invoice_date->date_i18n( 'Y-m-d H:i:s' ) );
+	public function store_name( $document_date ) {
+		$document_date      = date( 'Y', strtotime( $document_date ) );
```
Spreeuw, line 119: `$document_date` is an instance of `WC_DateTime`, which has a `__toString` magic method but it's preferrable to just use the internal functions (also for timezone safety)

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,50 @@ public function init_number() {
...
+		$reset_numeration   = isset( $this->settings['reset_number_yearly'] ) ? true : false;
+		$current_store_name = "{$this->slug}_number_{$document_date}";
 
-		$this->set_number( $invoice_number );
+		// reset: on
+		if( $reset_numeration ) {
+			$store_name          = $current_store_name;
+			$number_store_method = WPO_WCPDF()->settings->get_sequential_number_store_method();
+			$number_store        = new Sequential_Number_Store( $store_name, $number_store_method );
+			$number_store->set_next( apply_filters( 'wpo_wcpdf_reset_number_yearly_start', 1, $this ) );
```
Spreeuw, line 129: this will reset the number each document. Maybe you want to do this only if the store exists? You could make that check the primary condition and make the `$reset_numeration` check secondary

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,50 @@ public function init_number() {
...
+
+		// reset: off
+		} else {
+			$store_name   = "{$this->slug}_number_{$current_date}";
+			$store_exists = Sequential_Number_Store::store_name_exists( $store_name );
```
Spreeuw, line 134: store name could have been filtered. What you could do instead is directly instantiate (before these conditionals) and in the `init()` of `Number_Store`, set a property `$is_new` or something like that when this bit has been executed:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/05efc0fa692a0a75a17645744d6c1e671239f3aa/includes/documents/class-wcpdf-sequential-number-store.php#L61-L74

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,50 @@ public function init_number() {
...
+
+			// if store don't exist it's a new year
+			if( ! $store_exists ) {
+				$number_store_method = 'calculate';
+				$next_number         = Sequential_Number_Store::get_next( $current_store_name, $number_store_method );
+
+				// set next number based on current store
+				if( ! empty( $next_number ) ) {
```
Spreeuw, line 142: this seems redundant to me (wouldn't that always be the case if `! $store_exists` ?

includes/documents/class-wcpdf-sequential-number-store.php
```diff
@@ -110,18 +110,23 @@ public function increment( $order_id = 0, $date = null ) {
 	 * Get the number that will be used on the next increment
 	 * @return int next number
 	 */
-	public function get_next() {
+	public static function get_next( $store_name, $method ) {
```
Spreeuw, line 113: if you can avoid making this static, that has my preference for better backwards compatibility (same for the usage of function arguments, which 3rd party integrations do not have!)

includes/documents/class-wcpdf-sequential-number-store.php
```diff
@@ -173,6 +178,22 @@ public function get_last_date( $format = 'Y-m-d H:i:s' ) {
 
 		return $formatted_date;
 	}
+
+	/**
+	 * @return bool
+	 */
+	public static function store_name_exists( $store_name ) {
```
Spreeuw, line 185: see other comment about `$store_exists`

Spreeuw (changes requested):
I think it would be a good idea to make a list of tests that we should run, including scenarios where people are using filters for the table name.

includes/documents/class-wcpdf-invoice.php
alexmigf: In the document or in the settings class?

includes/documents/class-wcpdf-sequential-number-store.php
alexmigf: Are you referring to the `static`?

Spreeuw: I'm referring to this comment: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/153/files#r577528349

includes/documents/class-wcpdf-invoice.php
Spreeuw: in the document class. This:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/97ae3403ade900fc26c3d703fa814550d3b6583c/includes/documents/class-wcpdf-invoice.php#L106
Would then also be in that same `get_sequential_number_store()` method, keeping it clean

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,27 +103,25 @@ public function init_number() {
 			return $invoice_number;
 		}
 
-		$number_store_method = WPO_WCPDF()->settings->get_sequential_number_store_method();
-		$number_store_name   = $this->store_name();
-		$number_store        = new Sequential_Number_Store( $number_store_name, $number_store_method );
-		$invoice_number      = $number_store->increment( $this->order_id, $this->get_date()->date_i18n( 'Y-m-d H:i:s' ) );
+		$store_name     = $this->get_store_name();
```
Spreeuw, line 106: I think this should be done inside `get_sequential_number_store()`.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,27 +103,25 @@ public function init_number() {
...
 		$current_date        = date( 'Y' );
 		$reset_numeration    = isset( $this->settings['reset_number_yearly'] ) ? true : false;
 		$current_store_name  = "{$this->slug}_number_{$document_date}";
-		$number_store_method = WPO_WCPDF()->settings->get_sequential_number_store_method();
 
 		// reset: on
 		if( $reset_numeration ) {
 			$store_name   = $current_store_name;
-			$number_store = new Sequential_Number_Store( $store_name, $number_store_method );
+			$number_store = $this->get_sequential_number_store( $store_name );
```
Spreeuw, line 124: this makes more sense the other way around, `get_sequential_number_store()`  calling `get_store_name()`

Spreeuw: That is, if you need the separation at all!

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,58 @@ public function init_number() {
...
 
-		$invoice_date = $this->get_date();
-		$invoice_number = $number_store->increment( $this->order_id, $invoice_date->date_i18n( 'Y-m-d H:i:s' ) );
+		return apply_filters( "wpo_wcpdf_{$this->slug}_store_name", $store_name, $this );
+	}
 
-		$this->set_number( $invoice_number );
+	public function get_sequential_number_store( $store_name ) {
+		global $wpdb;
+		$method = isset( WPO_WCPDF()->settings->debug_settings['calculate_document_numbers'] ) ? 'calculate' : 'auto_increment';
```
Spreeuw, line 149: default should be `auto_increment` (this is the plugin default after all!)

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,25 +103,58 @@ public function init_number() {
...
 
-		return $invoice_number;
+		// safety first - always use calculate when auto_increment_increment is not 1
```
Spreeuw, line 151: in which scenario does this happen? If this can happen it's better to handle it internally in the `Number_Store` class I think

includes/documents/class-wcpdf-invoice.php
alexmigf: And isn't like that?

alexmigf: Is using isset, I believe when it's checked is `calculate`.

includes/documents/class-wcpdf-invoice.php
alexmigf: This was added by you when the method was in the settings class https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/153/files/f3a3542be5df1a5479249410ecba52f8641e843a#diff-3e73a6f9e3725b0651b3b195d872a261e1d5658b8b2421ac8b4c3452fc59abe5L248

includes/documents/class-wcpdf-invoice.php
Spreeuw: correct, I was sleeping! 🤐

includes/documents/class-wcpdf-invoice.php
Spreeuw: I see 🤔 maybe let's keep it like that just in case then... If it aint broken, don't fix it!
I would keep the call there then, and replace all occurences of this:
```php
isset( WPO_WCPDF()->settings->debug_settings['calculate_document_numbers'] ) ? 'auto_increment' : 'calculate'
```
to this:
```php
WPO_WCPDF()->settings->get_sequential_number_store_method();
```

alexmigf:
@Spreeuw I only can find this snippet from the Helpscout search:

```php
add_filter( 'wpo_wcpdf_number_store_table_name', 'wpo_wcpdf_number_store_table_name', 10, 3 );
function wpo_wcpdf_number_store_table_name( $table_name, $store_name, $method ) {
	global $wpdb;
	return "{$wpdb->base_prefix}_wcpdf_{$store_name}";
}
```

Do you know more?

Spreeuw:
For that filter it's indeed the only snippet I can see we provided. I think this would still function, since we provide `$store_name` from outside the class? (I'm not 100% sure about this)

A more difficult one to tackle may be `wpo_wcpdf_document_sequential_number_store` since people could have created multiple stores depending on properties of the order. This filter appears to have been deleted in this branch, we could reintroduce it here:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a9225134ec1a9ed8efeb1bad4ef05da2ed1746c4/includes/documents/class-wcpdf-invoice.php#L146-L151
But that would still not help for the migration in the upgrade routine.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,27 +103,53 @@ public function init_number() {
...
+			$store_name   = $this->get_sequential_number_store_name( $now );
+			$number_store = new Sequential_Number_Store( $store_name, $method );
+	
+			if ( $number_store->is_new ) {
+				// get last year's store if available
+				$last_year         = $now->modify('-1 year');
+				$last_store_name   = $this->get_sequential_number_store_name( $last_year );
+				$last_number_store = new Sequential_Number_Store( $last_store_name, $method );	
+				if ( ! $number_store->is_new ) {
+					$number_store->set_next( apply_filters( 'wpo_wcpdf_reset_number_yearly_start', $last_number_store->get_next(), $this ) );
```
Spreeuw, line 138: I think we should drop the filter here, because in this scenario reset is off

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,27 +103,53 @@ public function init_number() {
...
+		}
+	
+		return $number_store;
+	}
+	
+	public function get_sequential_number_store_name( $date ) {
+		$year       = $date->date_i18n( 'Y' );
+		$store_name = "{$this->slug}_number_{$year}";
+	
+		return apply_filters( "wpo_wcpdf_{$this->slug}_store_name", $store_name, $this );
```
Spreeuw, line 150: just `_store_name` may be confusing, I would explicitly call it `_number_store_name` in the filter

includes/documents/class-wcpdf-sequential-number-store.php
```diff
@@ -35,19 +35,28 @@ class Sequential_Number_Store {
 	 */
 	public $table_name;
 
+	/**
+	 * Store name or table name could be filtered
```
Spreeuw, line 39: Seems an incorrect description

alexmigf:
> But that would still not help for the migration in the upgrade routine.

Does it need migration? Since the number store is filtered, it should be using other tables and not the default.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -146,8 +153,11 @@ public function get_sequential_number_store() {
 	public function get_sequential_number_store_name( $date ) {
 		$year       = $date->date_i18n( 'Y' );
 		$store_name = "{$this->slug}_number_{$year}";
-	
-		return apply_filters( "wpo_wcpdf_{$this->slug}_store_name", $store_name, $this );
+
+		// legacy filter for backward compatibility
+		$store_name = apply_filters( 'wpo_wcpdf_document_sequential_number_store', $store_name, $this );
```
Spreeuw, line 158: This should be filtered before apending the `_{$year}`, otherwise it breaks current implementations.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -133,9 +133,16 @@ public function get_sequential_number_store() {
 				// get last year's store if available
 				$last_year         = $now->modify('-1 year');
 				$last_store_name   = $this->get_sequential_number_store_name( $last_year );
-				$last_number_store = new Sequential_Number_Store( $last_store_name, $method );	
+
+				// if the filter 'wpo_wcpdf_document_sequential_number_store' is used and the name is the same
```
Spreeuw, line 137: the year suffix is independent (should be), so the `wpo_wcpdf_document_sequential_number_store` filter will not affect this.

Spreeuw (changes requested):

Spreeuw:
> Does it need migration? Since the number store is filtered, it should be using other tables and not the default.
Yes because otherwise the number reset on those custom stores will be broken.

I was thinking that maybe instead of doing a one-time migration after updating the plugin, we could also do an auto-migration inside the `get_sequential_number_store_name` method. Basically you check if the `yearless` version exists, and if so, copy it to the 'current year' and then return the name of that new store as if nothing happened :) (logging this would be wise though)
```php
public function get_sequential_number_store_name( $date ) {
	$year            = $date->date_i18n( 'Y' );
	$store_base_name = apply_filters( 'wpo_wcpdf_document_sequential_number_store', "{$this->slug}_number", $this );

	// migrate old stores without year suffix
	$this->maybe_migrate( $store_base_name ); // we could make this a setting in the status tab? or just remove the migration after a year...

	return apply_filters( "wpo_wcpdf_{$this->slug}_number_store_name", "{$store_base_name}_{$year}", $this );
}
```

includes/documents/class-wcpdf-invoice.php
```diff
@@ -151,13 +145,35 @@ public function get_sequential_number_store() {
...
-		$year       = $date->date_i18n( 'Y' );
-		$store_name = "{$this->slug}_number_{$year}";
-
-		// legacy filter for backward compatibility
-		$store_name = apply_filters( 'wpo_wcpdf_document_sequential_number_store', $store_name, $this );
+		$year            = $date->date_i18n( 'Y' );
+		$store_base_name = apply_filters( 'wpo_wcpdf_document_sequential_number_store', "{$this->slug}_number", $this );
+	
+		// migrate old stores without year suffix
+		$number_store_migrated = get_option( "wpo_wcpdf_number_store_{$store_base_name}_migrated" );
```
Spreeuw, line 152: rather than working with a per-store option, I would suggest using a global setting to enable or disable migration, which you enable in the plugin upgrade procedure

includes/documents/class-wcpdf-invoice.php
```diff
@@ -151,13 +145,35 @@ public function get_sequential_number_store() {
...
+		if( $wpdb->get_var( "SHOW TABLES LIKE '{$store_base_name}'") == $store_base_name ) {
+			$query = $wpdb->query( "ALTER TABLE {$store_base_name} RENAME {$new_table_name} ");
+			if( $query ) {
+				update_option( "wpo_wcpdf_number_store_{$store_base_name}_migrated", 1 );
+				wcpdf_log_error( sprintf( __( 'Number store migration from %s (old table) to %s (new table) run successfully!', 'woocommerce-pdf-invoices-packing-slips' ), $store_base_name, $new_table_name ), 'info' );
+			} else {
+				wcpdf_log_error( sprintf( __( 'An error occurred while trying to migrate the number store %s: %s', 'woocommerce-pdf-invoices-packing-slips' ), $store_base_name, $wpdb->last_error ) );
+			}
+		} else {
+			wcpdf_log_error( sprintf( __( 'Legacy number store table %s not found: %s', 'woocommerce-pdf-invoices-packing-slips' ), $store_base_name, $wpdb->last_error ), 'warning' );
```
Spreeuw, line 175: this would keep triggering errors for each `Number_Store` instance for sites that never had the legacy store in the first place

includes/documents/class-wcpdf-invoice.php
```diff
@@ -151,13 +145,35 @@ public function get_sequential_number_store() {
...
+	}
 
-		return apply_filters( "wpo_wcpdf_{$this->slug}_number_store_name", $store_name, $this );
+	private function maybe_migrate_number_store( $store_base_name ) {
+		global $wpdb;
+		$now            = new \WC_DateTime( "@".time(), new \DateTimeZone( 'UTC' ) );
+		$year           = $now->date_i18n( 'Y' );
+		$new_table_name = "{$store_base_name}_{$year}";
+
+		if( $wpdb->get_var( "SHOW TABLES LIKE '{$store_base_name}'") == $store_base_name ) {
```
Spreeuw, line 166: This doesn't seem right, does this not need the prefix and the filter?
```php
apply_filters( "wpo_wcpdf_number_store_table_name", "{$wpdb->prefix}wcpdf_{$store_name}", $store_name, $method )
```

includes/documents/class-wcpdf-invoice.php
Spreeuw: You could make this setting disappear after it's unchecked (and as a result, invisible for new users), and provide an override with a filter.

alexmigf: But doing that, if the user doesn't uncheck, the `maybe_migrate` procedure will always try to run right? Do you need that? Because could slow down the `init_number` process.

Spreeuw: we can remove the procedure after a year, and I don't think `get_option` is much faster than `SHOW TABLES LIKE '{$store_base_name}` but it's something you could measure.

includes/documents/class-wcpdf-invoice.php
alexmigf: Sorry, don't what you mean.

alexmigf: Yes, sorry.

alexmigf:
### Tests

#### Main test

* Have legacy table created and see if gets renamed on upgrade
* Having the new year table, simulate a new year change with:
    * Reset numeration on, should start with 1 on the new table
    * Reset numeration off, should continue the prior numeration on the new table

#### Others

* Re-test the above using the filter 'wpo_wcpdf_document_sequential_number_store'
* Re-test the above using the filter 'wpo_wcpdf_{$this->slug}_number_store_name'
* Re-test the above using the filter 'wpo_wcpdf_number_store_table_name'

Spreeuw:
Thanks for sharing your tests! I would also test with orders from last year and invoice date set to order date (in between 2 orders from this year). So: 2021 invoice > 2020 invoice > 2021 invoice > 2020 invoice.

And please share your filters. You can update the OP with this list of tests.

includes/documents/class-wcpdf-invoice.php
Spreeuw: You still didn't apply the filter though
```php
$old_table_name = "{$wpdb->prefix}wcpdf_{$store_base_name}"; 
```
Which will not work correctly for multisite with shared invoice numbers (one of the filters from your tests I assume?)
```php
$old_table_name = apply_filters( "wpo_wcpdf_number_store_table_name", "{$wpdb->prefix}wcpdf_{$store_base_name}", $store_base_name, $method ); 
```
you would need to pass `$method` here for full support of the filter though

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,27 +103,77 @@ public function init_number() {
...
+	
+		return $number_store;
+	}
+	
+	public function get_sequential_number_store_name( $date ) {
+		$year            = $date->date_i18n( 'Y' );
+		$store_base_name = apply_filters( 'wpo_wcpdf_document_sequential_number_store', "{$this->slug}_number", $this );
+	
+		// migrate old stores without year suffix
+		$migrate_number_stores = WPO_WCPDF()->settings->debug_settings['migrate_number_stores'];
```
Spreeuw, line 152: This will throw a notice when the setting is disabled/unchecked.

includes/documents/class-wcpdf-invoice.php
alexmigf: Do you prefer to pass it from function to function or call it again?

includes/documents/class-wcpdf-invoice.php
```diff
@@ -135,8 +135,10 @@ public function get_sequential_number_store() {
 				$last_store_name   = $this->get_sequential_number_store_name( $last_year, $method );
 				$last_number_store = new Sequential_Number_Store( $last_store_name, $method );
 				
-				if ( ! $number_store->is_new ) {
+				if ( $last_number_store ) {
```
Spreeuw, line 138: this condition will always be true

includes/class-wcpdf-install.php
```diff
@@ -347,6 +347,13 @@ protected function upgrade( $installed_version ) {
 			update_option( 'wpo_wcpdf_settings_debug', $debug_settings );
 		}
 
+		// 2.8.2: enable number store migration
+		if ( version_compare( $installed_version, '2.8.2', '<=' ) ) {
+			$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
+			$debug_settings['migrate_number_stores'] = 1;
+			update_option( 'wpo_wcpdf_settings_debug', $debug_settings );
+		}
```
Spreeuw, line 355: while doing my first tests, my tables were not being migrated and I was confused why until I realized that this was because my version was still at 2.8.2 and this flag was never set. While this is not a realistic scenario for the actual release, it did make me think about whether this is reliable enough (I'm thinking about the odd 'reinstall fonts' issue that comes along, even though that should be done automatically on each upgrade).

Would it be a big 'sacrifice' to do this the other way around? always migrate unless it's disabled? We could set the flag to disable migration in the install routine, which will only run for new installs.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -103,27 +103,73 @@ public function init_number() {
...
+	
+		return $number_store;
+	}
+	
+	public function get_sequential_number_store_name( $date, $method ) {
+		$year            = $date->date_i18n( 'Y' );
+		$store_base_name = apply_filters( 'wpo_wcpdf_document_sequential_number_store', "{$this->slug}_number", $this );
+	
+		// migrate old stores without year suffix
+		if( isset( WPO_WCPDF()->settings->debug_settings['migrate_number_stores'] ) ) {
```
Spreeuw, line 149: we should add a filter to this condition

Spreeuw (changes requested):
Looking great already, though I did run into some issues, see comments above

includes/documents/class-wcpdf-invoice.php
```diff
@@ -305,7 +351,7 @@ public function init_settings() {
 				'callback'		=> 'next_number_edit',
 				'section'		=> 'invoice',
 				'args'			=> array(
-					'store'			=> 'invoice_number',
+					'store'			=> 'invoice_number_'.date('Y'),
```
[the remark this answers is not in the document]
Spreeuw: You marked this as resolved but you forgot the backwards compatibility I mentioned:

> This way you can check in the settings callback if a `Number_Store` instance is passed and if so, use that, _otherwise use the method currently used_.

includes/class-wcpdf-install.php
alexmigf: @Spreeuw but is `<=`

Spreeuw: yes but the `upgrade` method only runs when the version number _changes_
Like I say though, my own scenario with the same version number is not really relevant, I'm more worried about other scenarios where the upgrade procedure it's ran for whatever reason

alexmigf: I believe this will have no effect because the settings callback runs first and will do the migration.

includes/class-wcpdf-install.php
```diff
@@ -177,6 +177,14 @@ protected function install() {
...
 		// set transient for wizard notification
 		set_transient( 'wpo_wcpdf_new_install', 'yes', DAY_IN_SECONDS * 2 );
+
+		// 2.8.2: migrate number stores
+		if ( empty( get_option( 'wpo_wcpdf_number_stores_migrated' ) ) ) {
+			$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
+			$debug_settings['migrate_number_stores'] = 1;
+			update_option( 'wpo_wcpdf_settings_debug', $debug_settings );
+			update_option( 'wpo_wcpdf_number_stores_migrated', 1 );
+		}
```
Spreeuw, line 187: This is now the wrong way around. This will enable the migration for _new installs_ (which have nothing to migrate) and not do any migration for the people that updated (that need their old tables migrated)! and `wpo_wcpdf_number_stores_migrated` will always be empty here since it's a fresh install and will never run again after that.

Instead, you should set the setting to 0 in the debug defaults here:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/70d8cc08f9a6bfbc2fdabf6046fe19c7703ba3f6/includes/class-wcpdf-install.php#L165-L172

Then you can set the default of the setting itself to 1 and also enable `store_unchecked` like you did here:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/70d8cc08f9a6bfbc2fdabf6046fe19c7703ba3f6/includes/documents/class-wcpdf-invoice.php#L309-L310

Finally, when reading the setting, you default to 1 in case no value was set, something like this:
```php
// migrate when setting not set (means default = set) or when set and not 0
$migrate_number_stores = !isset( WPO_WCPDF()->settings->debug_settings['migrate_number_stores'] ) ) || WPO_WCPDF()->settings->debug_settings['migrate_number_stores'] != 0 );

// migrate old stores without year suffix
if( apply_filters( 'wpo_wcpdf_migrate_number_stores', $migrate_number_stores ) ) {
```

but please test thoroughly... And I may be overcomplicating things so if you have a simpler solution, feel free.

The bottom line should be: if anything is not set properly for whatever reason, migration should always be enabled by default.

Spreeuw (changes requested):
@alexmigf I've done another review of the code and it looks good to me. One final modification I would propose, is to move the new functions in `includes\documents\class-wcpdf-invoice.php` to the document abstract (`includes\documents\abstract-wcpdf-order-document.php`), so that other documents can use this too.

I propose to then remove the 'Draft' status from this PR and invite other team members to test the changes too.

Spreeuw:
I noticed I'm getting a ton of these in my logs:

> ERROR An error occurred while trying to migrate the number store wp_wcpdf_invoice_number: Table 'wp_wcpdf_invoice_number_2021' already exists

This is because I was on this branch first (which migrated the store to `wp_wcpdf_invoice_number_2021`), then worked on the current master for some time (which recreated the original `wp_wcpdf_invoice_number`, and then went back to the branch. It keeps trying to migrate the table now with the same error each time.

Terminator-Barbapapa:
Tests were successful. I'll do a code review this Monday.

alexmigf:
#### Potential clients for beta testing

- https://secure.helpscout.net/conversation/1454358453/29531/
- https://secure.helpscout.net/conversation/866334526/17900/
- https://secure.helpscout.net/conversation/628054577/14069/
- https://secure.helpscout.net/conversation/538044276/12492/
- https://secure.helpscout.net/conversation/516320459/12091/
- https://secure.helpscout.net/conversation/1372074924/27437/
- https://secure.helpscout.net/conversation/1389077075/27843/

@Spreeuw do you have others to include?

alexmigf:
Re-open

PR #156, opened 2021-03-09
Fix translation in get_payment_method()

Mysh3ll:
Fix translation for payment method.

Currently if the payment method is set to "**Other**" this method returns the string "**other**" without the first capital letter.

But translations in the domain '**woocommerce**' start with a capital letter. So in the case of a translation, this method returns the payment method in English instead of the translation because it cannot find a match.

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -430,7 +430,7 @@ abstract class Order_Document_Methods extends Order_Document {
 			$payment_method_title = WCX_Order::get_prop( $this->order, 'payment_method_title', 'view' );
 		}
 
-		$payment_method = __( $payment_method_title, 'woocommerce' );
+		$payment_method = __( ucfirst($payment_method_title), 'woocommerce' );
 
 		return apply_filters( 'wpo_wcpdf_payment_method', $payment_method, $this );
 	}


Spreeuw:
I'm not sure if doing `ucfirst()` is a good approach here, after all, this is the actual frontend string passed by WooCommerce.
The root cause of this lies in WooCommerce.

In the meta box itself they have a the uppercase frontend string:
https://github.com/woocommerce/woocommerce/blob/23710744c01ded649d6a94a4eaea8745e543159f/includes/admin/meta-boxes/class-wc-meta-box-order-data.php#L408-L412
But when saving the order, they don't use that string anymore:
https://github.com/woocommerce/woocommerce/blob/23710744c01ded649d6a94a4eaea8745e543159f/includes/admin/meta-boxes/class-wc-meta-box-order-data.php#L599-L601

To do this the correct way, they should take that into the order data too:
```php
if ( isset( $methods ) && isset( $methods[ $payment_method ] ) ) {
	$payment_method_title = $methods[ $payment_method ]->get_title();
} elseif ( 'other' == $payment_method ) {
	$payment_method_title = __( 'Other', 'woocommerce' );
}
```

You could either file a PR for this at WooCommerce (this would be the best option for future compatibility accross the board), or if you want to keep your existing method, you could use our `wpo_wcpdf_payment_method` hook:
```php
add_action( 'wpo_wcpdf_payment_method', function( $payment_method_title, $document ){
	if ( 'other' == $payment_method ) {
		$payment_method_title = __( 'Other', 'woocommerce' );
	}
	return $payment_method_title;
}, 10, 2 );
```
Hope that helps!

PR #164, opened 2021-03-29
Maintain year-based number stores to handle yearly resets more reliably

### LINKED ISSUE #149: Create new store tables in DB for each year if reset setting enabled

alexmigf:
_Edit: this solution evolved slowly during development and we ended up using the following strategy:_
> # Only generating the additional tables as a fallback
> 
> Instead of appending the year by default, we could continue to work with a single table for the 'current' sequence, and retiring that table as soon as a new period has started.
> 
> Basically, upon installation we would have only one table for each document:
> 
> ```
> wp_wcpdf_invoice_number // 2021, CURRENT YEAR
> ```
> 
> this would be used exactly as in the current release, and if yearly reset is **not** enabled, nothing would change. When yearly reset is enabled, as soon as the new year starts, this table is 'retired' by renaming it by appending the year, now we have:
> 
> ```
> wp_wcpdf_invoice_number_2021 // PAST YEAR, renamed from wp_wcpdf_invoice_number
> wp_wcpdf_invoice_number      // 2022, CURRENT YEAR
> ```
> 
> Likewise, when a number is requested for a future year, this would create a dedicated table, but only in case the yearly reset is enabled.
> 
> ```
> wp_wcpdf_invoice_number      // 2021, CURRENT YEAR
> wp_wcpdf_invoice_number_2022 // FUTURE
> ```
> 
> In that case, when the new year starts it would check if that table exists and use that (removing the year suffix), or otherwise simply create a new table `wp_wcpdf_invoice_number` since the old version would have been renamed
> 
> To make this reliable around the year changes, we need to know what year the unprefixed table (`wp_wcpdf_invoice_number`) belongs to. I am not sure if using the date column is a good way to do this (e.g. retrieving the last row from the table and checking the year), or if it's faster/easier to keep a simple `wp_option` field storing the 'current' year value.
> 
> This approach does **not** address concerns about additional table creation per se, but it removes most of the **reliability** on those additional tables: if a backup would not have the additional tables it wouldn't cause much damage (if any) because the current sequence is always in the same table.

_[continuation from #153 that was merged by mistake.]_

closes #149

#### Main test

* 1: Have legacy table created and see if gets renamed :heavy_check_mark: 
* 2: Having the new year table, simulate a new year change with:
    * 2.1: Reset numeration on, should start with 1 on the new table :heavy_check_mark: 
    * 2.2: Reset numeration off, should continue the prior numeration on the new table :heavy_check_mark: 

#### Others

* 3: Re-test the above using the filter `wpo_wcpdf_document_sequential_number_store`
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: 
    * 2.2: :heavy_check_mark: 
* 4: Re-test the above using the filter `wpo_wcpdf_{$this->slug}_number_store_name`
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark:  (resets monthly)
    * 2.2: :heavy_check_mark:  (resets monthly)
* 5: Re-test the above using the filter `wpo_wcpdf_number_store_table_name`
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: 
    * 2.2: :heavy_check_mark: 
* 6: Test with orders from last year
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: 
    * 2.2: :heavy_check_mark: 
* 7: In two 2020 orders set "Invoice date" to "order date" and generate Invoices
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: (keeps using the 2020 store, when the year changes it resets)
    * 2.2: :heavy_check_mark: (uses the last store in use, and continues the same numeration on year reset)
* 8: In two 2021 orders, set "Invoice date" to "order date" and generate Invoices
    * 1: :heavy_check_mark: 
    * 2.1: :heavy_check_mark: (keeps using the 2021 store, when the year changes it resets)
    * 2.2: :heavy_check_mark: (uses the last store in use, and continues the same numeration on year reset)
* 9: Test number stores migration on install :heavy_check_mark: 

#### Used filters

3:
```php
add_filter( 'wpo_wcpdf_document_sequential_number_store' , function( $number_store_name, $document ) {
	return 'test_' . $number_store_name;
}, 10, 2 );
```
4:
```php
// resets number store monthly
add_filter( 'wpo_wcpdf_invoice_number_store_name', function( $number_store_name, $store_base_name, $date, $method, $document ) {
	$year  = $date->date_i18n( 'Y' );
	$month = $date->date_i18n( 'm' );
	return "{$store_base_name}_{$month}_{$year}";
}, 10, 5 );
```
5:
```php
add_filter( 'wpo_wcpdf_number_store_table_name', function( $table_name, $store_name, $method ) {
	global $wpdb;
	return "{$wpdb->base_prefix}_wcpdf_{$store_name}";
}, 10, 3 );
```


Spreeuw:
One of our beta testers raised concerns about the fact that the plugin creates tables dynamically rather than only during the plugin installation and pointed out that this may not work well with fine-tuned backup solutions and uninstallation watchers that would track databases being added by plugins and remove them upon uninstalling the plugin.

I think this is a valid concern and we should think about ways to avoid this as much as possible.

The client suggested using a single table with offsets indicating the reset boundaries, but apart from technical complications of not having direct access to the number itself (possibly introducing a new race condition), I don't see how that would fix the issue of intermittand invoice creation between different years.
Consider this scenario (raw number, year)

001 - 2020 => 001 (offset = 0)
...
101 - 2020 => 101
102 - 2020 => 102
103 - 2021 => 1 (offset = 102)
104 - 2021 => 2
105 - 2020 => 105 (so 103+104 are missing!)

I see two alternative approaches

# Only generating the additional tables as a fallback
Instead of appending the year by default, we could continue to work with a single table for the 'current' sequence, and retiring that table as soon as a new period has started.

Basically, upon installation we would have only one table for each document:
```
wp_wcpdf_invoice_number // 2021
```
this would be used exactly as in the current release, and if yearly reset is **not** enabled, nothing would change.
When yearly reset is enabled, as soon as the new year starts, this table is 'retired' by renaming it by appending the year, now we have:
```
wp_wcpdf_invoice_number_2021 // renamed from wp_wcpdf_invoice_number
wp_wcpdf_invoice_number // 2022
```
Likewise, when a number is requested for a future year, this would create a dedicated table, but only in case the yearly reset is enabled.
```
wp_wcpdf_invoice_number // 2021
wp_wcpdf_invoice_number_2022
```
In that case, when the new year starts it would check if that table exists and use that (removing the year suffix), or otherwise simply create a new table `wp_wcpdf_invoice_number` since the old version would have been renamed

To make this reliable around the year changes, we need to know what year the unprefixed table (`wp_wcpdf_invoice_number`) belongs to. I am not sure if using the date column is a good way to do this (e.g. retrieving the last row from the table and checking the year), or if it's faster/easier to keep a simple `wp_option` field storing the 'current' year value.

This approach does **not** address concerns about additional table creation per se, but it removes most of the **reliability** on those additional tables: if a backup would not have the additional tables it wouldn't cause much damage (if any) because the current sequence is always in the same table.

# Using slow methods for non-current years
Another approach would be to change nothing to the current approach, but keep a list of 'last numbers per year' in a `wp_option`. It would be a simple array looking something like this:
```
array(
    2020 => 42,
    2021 => 8,
    2022 => 2,
}
```
This could either be updated at every document number generation or only during the year transition.

With this in place, we could **calculate** the number for any years other than the current year (read option, then store +1).

This addresses the concerns about additional db tables, but does expose the user to the race condition, opening the risk for 
duplicate invoice numbers. The most likely cause of such a race condition would be a combination of using the **order date** as **invoice date**, and external processes (for example bulk completion via the API when payments are being processed, or external apps using the API for synchronous order status change processes).

@alexmigf what do you think about the above options? Any ideas for alternative approaches?

alexmigf:
Maybe the first option is better, since this introduces an important change to the database structure. If the user wants to reset the numeration yearly then we switch to the new structure next year, if not, things will keep the same and nothing changes for them.

I think this is the best approach, less risky, and because the switch will happen in the beginning of a new year, if something goes wrong should be less stressful to fix.

Spreeuw:
> If the user wants to reset the numeration yearly then we switch to the new structure next year

I'm not sure if you're saying the same as I was here. My proposed approach would always keep the same structure and use the additional tables as fallback for **non current** years only. It would always use e.g. `wp_wcpdf_invoice_number` for the current year, even with yearly reset.

alexmigf:
I see, I still prefer that solution. What's your preferred?

Spreeuw:
Technically, I definitely prefer option 1 ("Only generating the additional tables as a fallback") too, but it doesn't completely solve the "additional tables" issue that was raised by the beta tester (well, if they don't enable yearly reset, no additional tables would be created!). Option 2 does solve that but has the race-condition downside and is technically less elegant.

Personally I would give more priority to addressing the race condition than to prevent additional tables being created.

alexmigf:
> The most likely cause of such a race condition would be a combination of using the order date as invoice date...

Shouldn't this be the contrary?

Spreeuw:
the wording is a bit unclear, but what I mean is using the order date for the invoice date (or 'as if it was the invoice date')

alexmigf:
Updating the order status with REST works fine, numeration keeps sequential.

alexmigf:
Update batch order status with REST works fine too. Updated 8 orders with one API request, numbers keep sequential.

- Invoice date as order date :heavy_check_mark: 
- Using year table :heavy_check_mark: 
- Invoice attached to order complete email :heavy_check_mark:

Spreeuw:
What code did you test this on (I don't see any commits for option 2)? Did you test this with _concurrent_ requests?

alexmigf:
> Personally I would give more priority to addressing the race condition than to prevent additional tables being created.

I understood you wanted to go with number 2

Spreeuw:
That's exactly the opposite of what the text you quoted says though :)

1: has additional tables but addresses race condition
2: has race condition but addresses table issue

I wrote that I rather prevent a race condition than preventing additional tables :)

alexmigf:
> Likewise, when a number is requested for a future year, this would create a dedicated table, but only in case the yearly reset is enabled.

How I will distinguish this from a new year reset? Because the scenario is identical:

- Current year: 2021
- Future year: 2022
- Last year: 2021 (after applying the new year)

Should I do a check for month (January) and day (1)?

And when this "Future" scenario happens? Can you tell me an example?

Spreeuw:
as written above

> To make this reliable around the year changes, we need to know what year the unprefixed table (`wp_wcpdf_invoice_number`) belongs to. I am not sure if using the date column is a good way to do this (e.g. retrieving the last row from the table and checking the year), or if it's faster/easier to keep a simple `wp_option` field storing the 'current' year value.

So if the current year is not equal to the year registered for the unprefixed table, it means you're lapsing.

> And when this "Future" scenario happens? Can you tell me an example?

If a customer places an order in December and asks for the invoice to be dated January next year. Admin sets the invoice date (leave invoice number empty) and generates the invoice.

alexmigf:
> So if the current year is not equal to the year registered for the unprefixed table, it means you're lapsing.

Let's take the example of the current date 12/04/2021:

> Current year (saved option or DB table date): 2021 (let's assume we have already some Invoices created)
> Future year: 2022
> Last year: 2022 - 1 = 2021

Let's take the new year example: 01/01/2021

> Current year (saved option or DB table date): 2020
> Future year: 2021
> Last year: 2021 - 1 = 2020

I believe it's the same scenario of the year reset or not?

alexmigf:
Tested locally simulating year changes using the filter 'wpo_wcpdf_current_year'.

@YordanSoares, @Terminator-Barbapapa can you test too?

Spreeuw:
@alexmigf I think it's a little bit too early to ask for functionality tests (@YordanSoares, @Terminator-Barbapapa), let's review the code first

Spreeuw (approved):

dpeyou:
> Have last year table created and see if gets renamed 

What is the best way to test this for those without orders from last year?

Terminator-Barbapapa:
I think I've found some unexpected behaviour.

### Master branch

#### Settings:
* Display invoice date: _order date_
* Reset invoice number yearly: _off_
* No invoices generated yet

#### Three orders:
* Order number 1: order date 2020
* Order number 2: order date 2021
* Order number 3: order date 2021

#### Generating invoices:
* Generate invoice for order number 2 (order date 2021): invoice number: 1
* Generate invoice for order number 1 (order date 2020): invoice number: 2

#### Number store tables:
* `wp_wcpdf_invoice_number`: containing two rows (id 1 and 2)

<br>

### Switch to store by year branch

#### Settings:
* Reset invoice number yearly: _on_

#### Number store tables:
* `wp_wcpdf_invoice_number`: empty
* `wp_wcpdf_invoice_number_2020`: containing two rows (id 1 and 2)

#### Generating invoices:
* Generate invoice for order number 3 (order date 2021): invoice number: 1

<br>

### Expected behaviour

I would have expected the invoice for order number 3 to have invoice number 3. But instead the rows from the `wp_wcpdf_invoice_number` table are pulled into the `wp_wcpdf_invoice_number_2020` table, causing the next invoice number for the current year to be 1.

Spreeuw:
Thanks, that's a good find.

I think this is actually not a big issue. The reason this happened is because the _last_ invoice you created with the old version (having yearly reset turned off) was from 2020. So invoice number 2 belongs to 2020. When you then enabled yearly reset and generated the new invoice with the branch from this PR, it retired that old table as 2020.

I actually don't see a better way of handling this, right now. I would argue `1` for the new invoice is actually more correct than `3`, because `2` was for 2020. What do you think?

@alexmigf what's your opinion?

alexmigf:
I think it's a normal behavior since the order 1 and 2 were created with the current master and reset was off. After switching the branch and enable the reset I would expect that behavior.

Terminator-Barbapapa:
If I understand correctly the rows of the `wp_wcpdf_invoice_number` table get moved to a new table when '_Reset invoice number yearly_' is enabled, the last invoice created was not from the current year and the '_Display invoice date_' is set to '_Order date_'. 

Maybe in this case it is good to show a warning next to the 'Reset invoice number yearly' to avoid any confusion. Something like:

"Please note enabling this option will reset your current invoice number sequence, as your last generated invoice was from 2020."

But I agree it is a very niche case.

Spreeuw:
> If I understand correctly the rows of the `wp_wcpdf_invoice_number` table get moved to a new table when '_Reset invoice number yearly_' is enabled, the last invoice created was not from the current year and the '_Display invoice date_' is set to '_Order date_'.

Almost correct: they get moved when `reset_number_yearly` is enabled and the last invoice created was from before the current year (regardless of which settings resulted in that state).

> "Please note enabling this option will reset your current invoice number sequence, as your last generated invoice was from 2020."

Following your message I looked into adding this notice, but it's fairly complex (technically) because it needs to check two different sources in a context where these are not readily available. I think this case is so rare and arguably the result is what should be expected (the last invoice is from 2020 and the number is 2, yearly reset is enabled so the new 2021 invoice being set to 1 is not strange at all), that I think we don't need to bother with the message.

But it's very good to be aware of these cases so we can make a balanced decision about how to handle them.

YordanSoares:
### Tests 

1. Have last year table created and see if gets renamed
2. Having the default and the last year tables, simulate a new year change with:
 2.1. Reset numeration on, should start with 1 on the default table and move the last year numbers to the last year table
 2.2. Reset numeration off, should continue the prior numeration on the default table 

_I repeated all this steps @alexmigf left above in each of the following scenarios_:

- :heavy_check_mark: Main test:

image
_The first table is the default one, used by the current year (2021)_

- :heavy_check_mark: `wpo_wcpdf_document_sequential_number_store` test:

image
_The second table is the one used by the current year (2021)_

Just a remark, it's OK that the default `wp_wcpdf_invoice_number` table is created, even using this filter? That table remains empty, while using the filter, but I wanted to mention it to know if this is the expected behavior.

- :heavy_check_mark: `wpo_wcpdf_{$this->slug}_number_store_name` + month reset test:

image

- :heavy_check_mark: `wpo_wcpdf_number_store_table_name` test:

image

- :heavy_check_mark: Test with orders from last year
- :heavy_check_mark: In two 2020 orders set "Invoice date" to "order date" and generate Invoices
- :heavy_check_mark: In two 2022 orders and generate Invoices
- **Rare scenario test**: I don't know if this is what expected, but I got the same behavior @alexmigf described: after following the same scenario, the invoices from 2022 were using the default table, while the ones from 2021 uses the table with the 2021 suffix. It seems that when the reset year setting is enabled, the table takes the last invoice year like the current one (?)

dpeyou:
When using the filter to reset the table store monthly, should this be moved to the new table given that it is from December & it is the last row in this table?
Screenshot_2021-12-01_09-23-52

December Table:
Screenshot_2021-12-01_09-24-34

- Reset yearly enabled

Spreeuw:
@dpeyou can you give a more precise description of what you did in what order? the `{prefix}_wcpdf_invoice_number`  table is the regular table, presumably these were the numbers you created _before_ adding the filter?

dpeyou:
@Spreeuw Using this filter & all invoices in the screenshots were generated right after I created their order.
```php
// resets number store monthly
add_filter( 'wpo_wcpdf_invoice_number_store_name', function( $number_store_name, $store_base_name, $date, $method, $document ) {
	$year  = $date->date_i18n( 'Y' );
	$month = $date->date_i18n( 'm' );
	return "{$store_base_name}_{$month}_{$year}";
}, 10, 5 );
```

- 433's invoice already existed in `{prefix}_wcpdf_invoice_number`, I activated the filter above, then created 435. I was expecting 433\s invoice to be moved.

Spreeuw:
> 433's invoice already existed in `{prefix}_wcpdf_invoice_number`, I activated the filter above, then created 435. I was expecting 433\s invoice to be moved.

I see. The filter you used simply tells the plugin to use a table with the year+month suffixed for any new numbers that are generated. It doesn't handle existing tables or numbers. Basically the old table is not used anymore, so what you were seeing is expected if you applied that filter after creating the invoice for `433`.

Spreeuw:
@Terminator-Barbapapa @YordanSoares you have not approved the PR yet, does that mean you're not done testing or that you have any further questions/doubts?

dpeyou:
I just realized I'm unable to make a table ending in `_2022`. Instead, they are using the EXISTING  `{prefix}_wcpdf_invoice_number` table & orders from the current year are going to `{prefix}_wcpdf_invoice_number_2021`
- I am creating orders & setting the invoice year to 2022 before hitting the `Create` button.
- If I delete table `_2021`, create a new invoice for 2022, table  `{prefix}_wcpdf_invoice_number_2021` gets regenerated (empty) & I'm still looking for the latest  2022 invoice I created, it isn't in  `{prefix}_wcpdf_invoice_number`, which is where they were going.
- No code snippets active
- Reset Yearly = yes.

Spreeuw:
@dpeyou what's the last invoice in `{prefix}_wcpdf_invoice_number` ? If there was a 2022 in that table before you turned on yearly reset (or turned off any filters), it would continue on that as the 'current year'.

Terminator-Barbapapa:
> you have not approved the PR yet, does that mean you're not done testing or that you have any further questions/doubts?

Done testing. Will approve now!

Terminator-Barbapapa (approved):

dpeyou:
@Spreeuw 
> what's the last invoice in {prefix}_wcpdf_invoice_number ? 

Screenshot_2021-12-02_07-07-56
I went ahead & changed the date to 2021 & now the table  ending in `_2021` is generating.

I think I've found an issue however. If the Invoice number is set, I lose the Invoice (cannot find it in relevant database tables), regardless of whether I set the year to 2022 or 2021. If I **do not** set an Invoice number:
- 2021 -> `{prefix}_wcpdf_invoice_number`,
- 2022 ->  `{prefix}_wcpdf_invoice_number_2022`,

Screenshot_2021-12-02_07-37-14

Spreeuw:
> I think I've found an issue however. If the Invoice number is set, I lose the Invoice (cannot find it in relevant database tables), regardless of whether I set the year to 2022 or 2021. If I do not set an Invoice number:

That's expected behavior - if you manually set the invoice number, it doesn't need to (shouldn't!!!) generate an invoice number, so the number stores are never called.

dpeyou (approved):
Bug hunting unsuccessful. LGTM!

YordanSoares (approved):
> @Terminator-Barbapapa @YordanSoares you have not approved the PR yet, does that mean you're not done testing or that you have any further questions/doubts?

I only had doubts about the **rare scenario**, but I guess this is the expected behavior, so I proceed to approve this PR: Great job! 🎉

PR #167, opened 2021-04-13
Fixed issue where variable is not set

sebastien-roch:


alexmigf:
Thank you @sebastien-roch :rocket:

PR #171, opened 2021-04-16
Log Invoice generate info to order notes with setting enabled.

### LINKED ISSUE #170: Log PDF Invoice generation info to order notes on setting
To not use meta to save this information, @Spreeuw suggest to use the order notes to store PDF Invoice generation info, on a setting (disabled by default).

alexmigf:
closes #170

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -178,11 +178,8 @@ class Main {
 				do_action( 'wpo_wcpdf_email_attachment', $pdf_path, $document_type, $document );
 
 				// log document creation to order notes
-				$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
-				if( isset( $debug_settings['log_to_order_notes'] ) ) {
-					$action = __( 'attachment', 'woocommerce-pdf-invoices-packing-slips' );
-					$this->log_document_creation_to_order_notes( $document, $action );
-				}
+				$this->log_document_creation_to_order_notes( $document, 'attachment' );
+				
 			} catch ( \Exception $e ) {
 				wcpdf_log_error( $e->getMessage(), 'critical', $e );
 				continue;
@@ -346,25 +343,10 @@ class Main {
 			$document = wcpdf_get_document( $document_type, $order_ids, true );
 
 			if ( $document ) {
-				do_action( 'wpo_wcpdf_document_created_manually', $document, $order_ids ); // note that $order_ids is filtered and may not be the same as the order IDs used for the document (which can be fetched from the document object itself)
+				do_action( 'wpo_wcpdf_document_created_manually', $document, $order_ids ); // note that $order_ids is filtered and may not be the same as the order IDs used for the document (which can be fetched from the document object itself with $document->order_ids)
 
 				// log document creation to order notes
-				$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
-				if( isset( $debug_settings['log_to_order_notes'] ) ) {
-					if( count( $order_ids ) > 1 ) {
-						$action = __( 'bulk', 'woocommerce-pdf-invoices-packing-slips' );
-						foreach( $order_ids as $order_id ) {
-							$order     = wc_get_order( $order_id );
-							if( empty( $order ) ) continue;
-
-							$_document = wcpdf_get_document( $document_type, $order, true );
-							$this->log_document_creation_to_order_notes( $_document, $action );
-						}
-					} else {
-						$action = __( 'single', 'woocommerce-pdf-invoices-packing-slips' );
-						$this->log_document_creation_to_order_notes( $document, $action );
-					}
-				}
+				$this->log_document_creation_to_order_notes( $document, array( 'bulk', 'single' ) );
 
 				$output_format = WPO_WCPDF()->settings->get_output_format( $document_type );
 				// allow URL override
@@ -941,10 +923,34 @@ class Main {
 	/**
 	 * Logs the document creation to the order notes
 	 */
-	public function log_document_creation_to_order_notes( $document, $action ) {
-		if( ! empty( $document ) && ! empty( $action ) && ! empty( $order = $document->order ) ) {
-			$note = sprintf( __( 'PDF %s %s created on %s action.', 'woocommerce-pdf-invoices-packing-slips' ), $document->get_title() , $document->get_number()->get_plain(), $action );
-			$order->add_order_note( $note );
+	public function log_document_creation_to_order_notes( $document, $context ) {
+		if( ! empty( $document ) && ! empty( $context ) && isset( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) ) {
+			$message = __( 'PDF %s created on %s action.', 'woocommerce-pdf-invoices-packing-slips' );
+
+			// bulk document
+			if( $document->is_bulk ) {
+				foreach( $document->order_ids as $order_id ) {
+					$order          = wc_get_order( $order_id );
+					if( empty( $order ) ) continue;
+
+					$document_title = ucwords( str_replace( '-', ' ', $document->type ) );
+					$note           = sprintf( $message, $document_title, 'bulk' );
+					$order->add_order_note( $note );
+				}
+				return;
+
+			// attachment & single documents
+			} else {
+				$order = $document->order;
+				if( empty( $order ) ) return;
+
+				if( $context != 'attachment' ) {
+					$context = 'single';
+				}
+
+				$note  = sprintf( $message, $document->get_title(), $context );
+				$order->add_order_note( $note );
+			}
 		}
 	}
 
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -176,6 +176,13 @@ class Main {
 				$attachments[] = $pdf_path;
 
 				do_action( 'wpo_wcpdf_email_attachment', $pdf_path, $document_type, $document );
+
+				// log document creation to order notes
+				$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
+				if( isset( $debug_settings['log_to_order_notes'] ) ) {
+					$action = __( 'attachment', 'woocommerce-pdf-invoices-packing-slips' );
+					$this->log_document_creation_to_order_notes( $document, $action );
+				}
 			} catch ( \Exception $e ) {
 				wcpdf_log_error( $e->getMessage(), 'critical', $e );
 				continue;
@@ -340,6 +347,24 @@ class Main {
 
 			if ( $document ) {
 				do_action( 'wpo_wcpdf_document_created_manually', $document, $order_ids ); // note that $order_ids is filtered and may not be the same as the order IDs used for the document (which can be fetched from the document object itself)
+
+				// log document creation to order notes
+				$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
+				if( isset( $debug_settings['log_to_order_notes'] ) ) {
+					if( count( $order_ids ) > 1 ) {
+						$action = __( 'bulk', 'woocommerce-pdf-invoices-packing-slips' );
+						foreach( $order_ids as $order_id ) {
+							$order     = wc_get_order( $order_id );
+							if( empty( $order ) ) continue;
+
+							$_document = wcpdf_get_document( $document_type, $order, true );
+							$this->log_document_creation_to_order_notes( $_document, $action );
+						}
+					} else {
+						$action = __( 'single', 'woocommerce-pdf-invoices-packing-slips' );
+						$this->log_document_creation_to_order_notes( $document, $action );
+					}
+				}
 
 				$output_format = WPO_WCPDF()->settings->get_output_format( $document_type );
 				// allow URL override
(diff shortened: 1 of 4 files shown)


includes/documents/class-wcpdf-invoice.php
```diff
@@ -77,14 +77,14 @@ public function init() {
...
+				$request_type = __( 'manual', 'woocommerce-pdf-invoices-packing-slips' );
 			// attachment request
 			} else {
-				$request_type = __( 'attachment request', 'woocommerce-pdf-invoices-packing-slips' );
+				$request_type = __( 'attachment', 'woocommerce-pdf-invoices-packing-slips' );
 			}
 
 			$invoice_date = $this->get_date();
-			$note         = sprintf( __( 'PDF Invoice %s generated on %s with %s.', 'woocommerce-pdf-invoices-packing-slips' ), $invoice_number, $invoice_date->date_i18n( wcpdf_date_format( $this, 'invoice_date' ) ), $request_type );
+			$note         = sprintf( __( 'PDF Invoice %s generated on %s with %s request.', 'woocommerce-pdf-invoices-packing-slips' ), $invoice_number, $invoice_date->date_i18n( wcpdf_date_format( $this, 'invoice_date' ) ), $request_type );
```
Spreeuw, line 87: invoice date is not necessarily the creation date. You can exclude the date altogether because the note itself will have a date.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -77,14 +77,14 @@ public function init() {
 		if( isset( $document_settings['log_to_order_notes'] ) && ! empty( $this->order ) ) {
 			// manual request
 			if( isset( $_REQUEST['action'] ) && $_REQUEST['action'] == 'generate_wpo_wcpdf' ) {
-				$request_type = __( 'manual request', 'woocommerce-pdf-invoices-packing-slips' );
+				$request_type = __( 'manual', 'woocommerce-pdf-invoices-packing-slips' );
```
Spreeuw, line 80: I would distinguish between single and bulk too.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -71,7 +71,23 @@ public function init() {
...
+		$invoice_number = $this->init_number();
+
+		// log to order notes if setting is enabled
+		if( isset( $document_settings['log_to_order_notes'] ) && ! empty( $this->order ) ) {
+			// manual request
+			if( isset( $_REQUEST['action'] ) && $_REQUEST['action'] == 'generate_wpo_wcpdf' ) {
+				$request_type = __( 'manual', 'woocommerce-pdf-invoices-packing-slips' );
+			// attachment request
+			} else {
+				$request_type = __( 'attachment', 'woocommerce-pdf-invoices-packing-slips' );
```
Spreeuw, line 83: you don't know that, could have been programmatically too

Spreeuw (changes requested):
I would make this general for all documents, and instead of doing this inside the document (after all it's not really a function of the document!), I would suggest to move it to `\WPO\WC\PDF_Invoices\Main`:
create a function `log_document_creation( $document, $context )` and call that function from the respective contexts (`generate_pdf_ajax` and `attach_pdf_to_email`)

includes/documents/class-wcpdf-invoice.php
```diff
@@ -423,6 +439,18 @@ public function init_settings() {
 					                   . __( "<strong>Caution:</strong> enabling this will also mean that if you change your company name or address in the future, previously generated documents will also be affected.", 'woocommerce-pdf-invoices-packing-slips' ),
 				)
 			),
+			array(
```
Spreeuw, line 442: should be a debugging setting, move to `Status tab`

includes/class-wcpdf-main.php
```diff
@@ -176,6 +176,13 @@ public function attach_pdf_to_email ( $attachments, $email_id, $order, $email =
 				$attachments[] = $pdf_path;
 
 				do_action( 'wpo_wcpdf_email_attachment', $pdf_path, $document_type, $document );
+
+				// log document creation to order notes
+				$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
```
Spreeuw, line 181: this is already loaded at init here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/830d26ae6a63d029c1763d6a12dbd95dd170beeb/includes/class-wcpdf-settings.php#L34
so you can avoid an extra db call

includes/class-wcpdf-main.php
```diff
@@ -176,6 +176,13 @@ public function attach_pdf_to_email ( $attachments, $email_id, $order, $email =
...
+				if( isset( $debug_settings['log_to_order_notes'] ) ) {
```
Spreeuw, line 182: Cleaner to do this check inside the log method itself

includes/class-wcpdf-main.php
```diff
@@ -176,6 +176,13 @@ public function attach_pdf_to_email ( $attachments, $email_id, $order, $email =
...
+					$action = __( 'attachment', 'woocommerce-pdf-invoices-packing-slips' );
```
Spreeuw, line 183: this should be a code/slug (and I think `$context` is a better description), not a translatable string.

includes/class-wcpdf-main.php
```diff
@@ -176,6 +176,13 @@ public function attach_pdf_to_email ( $attachments, $email_id, $order, $email =
...
+					$this->log_document_creation_to_order_notes( $document, $action );
```
Spreeuw, line 184: this can be a one-liner if `$context` is just a code anyway.

includes/class-wcpdf-main.php
```diff
@@ -341,6 +348,24 @@ public function generate_pdf_ajax() {
 			if ( $document ) {
 				do_action( 'wpo_wcpdf_document_created_manually', $document, $order_ids ); // note that $order_ids is filtered and may not be the same as the order IDs used for the document (which can be fetched from the document object itself)
 
+				// log document creation to order notes
+				$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
+				if( isset( $debug_settings['log_to_order_notes'] ) ) {
+					if( count( $order_ids ) > 1 ) {
```
Spreeuw, line 354: As noted in the comment right about this, `$order_ids` may be different from `$document->order_ids`!!!

includes/class-wcpdf-main.php
```diff
@@ -341,6 +348,24 @@ public function generate_pdf_ajax() {
...
+						$action = __( 'bulk', 'woocommerce-pdf-invoices-packing-slips' );
+						foreach( $order_ids as $order_id ) {
```
Spreeuw, line 356: I would do this looping inside the log function itself (checking `$document->is_bulk`?)

includes/class-wcpdf-main.php
```diff
@@ -341,6 +348,24 @@ public function generate_pdf_ajax() {
...
+							$order     = wc_get_order( $order_id );
+							if( empty( $order ) ) continue;
+
+							$_document = wcpdf_get_document( $document_type, $order, true );
```
Spreeuw, line 360: This is an expensive operation. Do we really need this? If we exclude the number from the note, we can simply reuse the `$document` object we already have.

includes/class-wcpdf-main.php
Spreeuw: Also, number of orders is not an indication of whether it was created via bulk actions or not (could be a single document created via bulk actions).

includes/class-wcpdf-main.php
```diff
@@ -913,6 +938,16 @@ public function set_phpmailer_validator( $mailArray ) {
 		return $mailArray;
 	}
 
+	/**
+	 * Logs the document creation to the order notes
+	 */
+	public function log_document_creation_to_order_notes( $document, $action ) {
+		if( ! empty( $document ) && ! empty( $action ) && ! empty( $order = $document->order ) ) {
+			$note = sprintf( __( 'PDF %s %s created on %s action.', 'woocommerce-pdf-invoices-packing-slips' ), $document->get_title() , $document->get_number()->get_plain(), $action );
```
Spreeuw, line 946: daisy-chaining is risky if you don't do a check on the return values, could lead to crashes

Spreeuw (changes requested):

includes/class-wcpdf-main.php
```diff
@@ -941,10 +923,34 @@ public function set_phpmailer_validator( $mailArray ) {
...
+		if( ! empty( $document ) && ! empty( $context ) && isset( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) ) {
+			$message = __( 'PDF %s created on %s action.', 'woocommerce-pdf-invoices-packing-slips' );
+
+			// bulk document
+			if( $document->is_bulk ) {
+				foreach( $document->order_ids as $order_id ) {
+					$order          = wc_get_order( $order_id );
+					if( empty( $order ) ) continue;
+
+					$document_title = ucwords( str_replace( '-', ' ', $document->type ) );
```
Spreeuw, line 936: like this is better: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/201754d923d0a48a1dd7e0913412d3a681a832e0/includes/documents/class-wcpdf-bulk-document.php#L85 (outside the foreach of course).

includes/class-wcpdf-main.php
```diff
@@ -941,10 +923,34 @@ public function set_phpmailer_validator( $mailArray ) {
...
+					$note           = sprintf( $message, $document_title, 'bulk' );
```
Spreeuw, line 937: here you're printing to a human string, so I would apply gettext to `bulk` (and maybe make it more descriptive, e.g.
`PDF %s created via %s` + `$document->get_title()` + `__('bulk action')`

includes/class-wcpdf-main.php
```diff
@@ -913,6 +920,40 @@ public function set_phpmailer_validator( $mailArray ) {
...
 
+	/**
+	 * Logs the document creation to the order notes
+	 */
+	public function log_document_creation_to_order_notes( $document, $context ) {
+		if( ! empty( $document ) && ! empty( $context ) && isset( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) ) {
+			$message = __( 'PDF %s created on %s action.', 'woocommerce-pdf-invoices-packing-slips' );
+
+			// bulk document
+			if( $document->is_bulk ) {
```
Spreeuw, line 931: This will generate error notices because it will only be set on bulk documents. `isset( $document->is_bulk )` will probably avoid this (but please double check, I'm never sure between isset and !empty for these cases!)

includes/class-wcpdf-main.php
```diff
@@ -913,6 +920,40 @@ public function set_phpmailer_validator( $mailArray ) {
...
+				}
+				return;
+
+			// attachment & single documents
+			} else {
+				$order = $document->order;
+				if( empty( $order ) ) return;
+
+				if( $context != 'attachment' ) {
+					$context = 'single';
```
Spreeuw, line 948: wondering if we should include frontend too?

includes/class-wcpdf-main.php
```diff
@@ -913,6 +938,45 @@ public function set_phpmailer_validator( $mailArray ) {
...
+	 */
+	public function log_single_to_order_notes( $document ) {
+		$this->log_to_order_notes( $document, __( 'single action', 'woocommerce-pdf-invoices-packing-slips' ) );
+	}
+
+	/**
+	 * Logs the my account document creation to the order notes
+	 */
+	public function log_my_account_to_order_notes( $document ) {
+		$this->log_to_order_notes( $document, __( 'my account action', 'woocommerce-pdf-invoices-packing-slips' ) );
```
Spreeuw, line 959: leave out "action"

includes/class-wcpdf-main.php
```diff
@@ -913,6 +938,45 @@ public function set_phpmailer_validator( $mailArray ) {
...
+	}
+
+	/**
+	 * Logs the attachment document creation to the order notes
+	 */
+	public function log_attachment_to_order_notes( $document ) {
+		$this->log_to_order_notes( $document, __( 'attachment action', 'woocommerce-pdf-invoices-packing-slips' ) );
```
Spreeuw, line 966: "email attachment"

includes/class-wcpdf-main.php
```diff
@@ -913,6 +938,45 @@ public function set_phpmailer_validator( $mailArray ) {
...
+	}
+
+	/**
+	 * Logs the document creation to the order notes
+	 */
+	public function log_to_order_notes( $document, $context ) {
```
Spreeuw, line 972: rename `$context` to `$created_via`

includes/class-wcpdf-main.php
```diff
@@ -336,10 +351,20 @@ public function generate_pdf_ajax() {
 
 		// if we got here, we're safe to go!
 		try {
+			// log document creation to order notes
+			if( count( $order_ids ) > 1 && $bulk ) {
```
Spreeuw, line 355: it's cleaner if you check `$_GET['bulk']` etc directly

includes/class-wcpdf-main.php
```diff
@@ -128,6 +128,9 @@ public function attach_pdf_to_email ( $attachments, $email_id, $order, $email =
 			do_action( 'wpo_wcpdf_before_attachment_creation', $email_order, $email_id, $document_type );
 
 			try {
+				// log document generation to order notes
+				add_action( 'wpo_wcpdf_init_document', array( $this, 'log_email_attachment_to_order_notes' ) );
```
Spreeuw, line 132: Did you test if this results in duplicate notes when bulk completing (=bulk emailing?) orders? If so, we should do `remove_action` afterwards.

alexmigf: Did not:

Captura de ecrã de 2021-04-20 14-14-04

Spreeuw: Is that the last order from the bulk series? (the first would only the function hooked once)

alexmigf: I tested with 3 orders, all have one note.

Spreeuw (approved):

PR #175, opened 2021-04-28
Adds a Save and Cancel buttons to the Document Data metabox

### LINKED ISSUE #148: Improve the UI of the "PDF document data" metabox.
When users need to edit the PDF details from the **PDF document data** meta box, they should press the Enter key to save the changes, because we don't have any buttons there yet.

Adding buttons in this area will improve the UI.

image
_A mockup of the "PDF document data" buttons_

alexmigf:
closes #148

Captura de ecrã de 2021-04-28 15-14-00

### DIFF
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -100,9 +100,6 @@ jQuery( function( $ ) {
 		let data       = $form.data();
 		let serialized = $form.find(":input:visible:not(:disabled)").serialize();
 
-		// Make sure all feedback icons are hidden before each call
-		$form.find('.document-action-success, .document-action-failed').hide();
-
 		// regenerate specific
 		if( action == 'regenerate' ) {
 			if ( window.confirm( wpo_wcpdf_ajax.confirm_regenerate ) === false ) {
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -55,26 +55,46 @@ jQuery( function( $ ) {
 		}
 
 		let $form = $(this).closest('.wcpdf-data-fields');
+		let data  = $form.data();
 
 		// Hide regenerate button
 		$form.find('.wpo-wcpdf-regenerate-document').hide();
 
+		// block ui
+		$form.block( {
+			message: null,
+			overlayCSS: {
+				background: '#fff',
+				opacity: 0.6
+			}
+		} );
+
 		$.ajax({
-			url:     wpo_wcpdf_ajax.ajaxurl,
-			data:    {
-				action  : 'wpo_wcpdf_delete_document',
-				security: $(this).data('nonce'),
-				document: $form.data('document'),
-				order_id: $form.data('order_id')
+			url:               wpo_wcpdf_ajax.ajaxurl,
+			data: {
+				action:        'wpo_wcpdf_delete_document',
+				security:      $(this).data('nonce'),
+				document_type: data.document,
+				order_id:      data.order_id,
 			},
 			type:    'POST',
 			context: $form,
 			success: function( response ) {
-				if ( response.success ) {
-					$(this).find(':input').val("");
-					$(this).find('.read-only').hide();
-					$(this).find('.wpo-wcpdf-delete-document').hide();
-				}
+				toggle_edit_mode( $form );
+
+				// update document DOM data
+				$form.closest('#wpo_wcpdf-data-input-box').load( document.URL + ' #wpo_wcpdf-data-input-box .postbox-header, #wpo_wcpdf-data-input-box .inside', function() {
+					let notice_type;
+					if( response.success ) {
+						notice_type = 'success';
+					} else {
+						notice_type = 'error';
+					}
+					$(this).find( ".wcpdf-data-fields[data-document='" + data.document +"'][data-order_id='" + data.order_id +"']" ).before( '<div class="notice notice-'+notice_type+' inline" style="margin:0 10px 10px 10px;"><p>'+response.data.message+'</p></div>' );
+				});
+
+				// unblock ui
+				$form.unblock();
 			}
 		});
 	} );
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -45,13 +45,18 @@ jQuery( function( $ ) {
 
 		// check visibility
 		if( $form.find(".read-only").is(":visible") ) {
-			$form.find(".read-only").hide();
-			$form.find(".editable").show();
 			if( edit == 'notes' ) {
+				$form.find(".read-only").hide();
 				$form.find(".editable-notes").show();
+				$form.find('.editable-notes :input').attr('disabled', false);
+				$form.closest('.wcpdf-data-fields').find('.wpo-wcpdf-document-buttons').show();
+			} else {
+				$form.find(".read-only").hide();
+				$form.find(".editable").show();
+				$form.find(".editable-notes").show();
+				$form.find(':input').attr('disabled', false);
+				$form.closest('.wcpdf-data-fields').find('.wpo-wcpdf-document-buttons').show();
 			}
-			$form.find(':input').attr('disabled', false);
-			$form.closest('.wcpdf-data-fields').find('.wpo-wcpdf-document-buttons').show();
 		} else {
 			$form.find(".read-only").show();
 			$form.find(".editable").hide();
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -139,9 +139,8 @@ jQuery( function( $ ) {
 						$form.find('.wpo-wcpdf-document-buttons').hide();
 
 						// update document DOM data
-						let wrapper = data.document + '-fields .wrapper';
-						$form.load( document.URL + ' .' + wrapper, function() {
-							$form.find('.document-action-success').show();
+						$form.closest('#wpo_wcpdf-data-input-box').load( document.URL + ' #wpo_wcpdf-data-input-box .postbox-header, #wpo_wcpdf-data-input-box .inside', function() {
+							$('#wpo_wcpdf-data-input-box .notice').show();
 						});
 					} else if( action == 'regenerate' ) {
 						$form.find('.document-action-success').show();
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -36,21 +36,31 @@ jQuery( function( $ ) {
 	
 	// enable invoice number edit if user initiated
 	$( '#wpo_wcpdf-data-input-box' ).on( 'click', '.wpo-wcpdf-set-date-number, .wpo-wcpdf-edit-date-number, .wpo-wcpdf-edit-document-notes', function() {
-		let $form = $(this).closest('.wcpdf-data-fields');
+		let $form = $(this).closest('.wcpdf-data-fields-section');
+		if ( $form.length == 0 ) { // no section, take overall wrapper
+			$form = $(this).closest('.wcpdf-data-fields');
+		}
+
+		let edit = $(this).data( 'edit' );
 
 		// check visibility
 		if( $form.find(".read-only").is(":visible") ) {
 			$form.find(".read-only").hide();
 			$form.find(".editable").show();
+			if( edit == 'notes' ) {
+				$form.find(".editable-notes").show();
+			}
 			$form.find(':input').attr('disabled', false);
 			$form.closest('.wcpdf-data-fields').find('.wpo-wcpdf-document-buttons').show();
 		} else {
 			$form.find(".read-only").show();
 			$form.find(".editable").hide();
+			$form.find(".editable-notes").hide();
 			$form.find(':input').attr('disabled', true);
 			$form.closest('.wcpdf-data-fields').find('.wpo-wcpdf-document-buttons').hide();
 		}
 	} );
+
 
 	// delete document
 	$( '#wpo_wcpdf-data-input-box' ).on( 'click', '.wcpdf-data-fields .wpo-wcpdf-delete-document', function() {
@@ -165,6 +175,7 @@ jQuery( function( $ ) {
 
 		$form.find(".read-only").show();
 		$form.find(".editable").hide();
+		$form.find(".editable-notes").hide();
 		$form.find(':input').attr('disabled', true);
 		$form.find('.wpo-wcpdf-document-buttons').hide();
 	} );
--- a/assets/js/order-script.js
+++ b/assets/js/order-script.js
@@ -47,71 +47,14 @@ jQuery( function( $ ) {
 		toggle_edit_mode( $form, edit );
 	} );
 
-
-	// delete document
-	$( '#wpo_wcpdf-data-input-box' ).on( 'click', '.wcpdf-data-fields .wpo-wcpdf-delete-document', function() {
-		if ( window.confirm( wpo_wcpdf_ajax.confirm_delete ) === false ) {
-			return; // having second thoughts
-		}
-
-		let $form = $(this).closest('.wcpdf-data-fields');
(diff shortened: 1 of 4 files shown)


assets/js/order-script.js
```diff
@@ -121,5 +125,20 @@ jQuery(document).ready(function($) {
 		
 	});
 
+	// save document
+	$( ".wpo-wcpdf-save" ).on("click", function() {
+		$( ".save_order" ).trigger( "click" );
```
Spreeuw, line 130: That saves the whole order. Shouldn't this work more like regerenate, only saving the invoice data?

YordanSoares (commented):
✔️ Tested: It works as expected :)

assets/js/order-script.js
alexmigf: I thought about that, I just wanted you to see first before start working on that. But yes, I agree.

includes/class-wcpdf-admin.php
```diff
@@ -567,32 +576,84 @@ public function get_bulk_actions() {
...
 
-			$order = WCX::get_order( $post_id );
-			if ( $invoice = wcpdf_get_invoice( $order ) ) {
-				$_POST = stripslashes_deep( $_POST );
-				$document_data = $this->process_order_document_form_data( $_POST, $invoice->slug );
-				$invoice->set_data( $document_data, $order );
+		// remove legacy hooks (older versions of Pro)
+		if( function_exists( 'WPO_WCPDF_Pro' ) ) {
+			$pro_installed_version = get_option( 'wpo_wcpdf_pro_version' );
+			if ( version_compare( $pro_installed_version, '7.2.4', '<' ) && version_compare( $pro_installed_version, '7.2.2', '>' ) ) {
```
Spreeuw, line 623: what's the reason for this? We never released such versions, but maybe that's just a typo?

I actually think we should separate the processes of ajax saving order data (which should be handled like regenerate is currently) and regular order saving. It seems the latter doesn't work anymore, correct?

assets/js/order-script.js
```diff
@@ -121,5 +122,72 @@ jQuery(document).ready(function($) {
 		
 	});
 
+	// save document
+	$( ".wcpdf-data-fields .wpo-wcpdf-save" ).on("click", function() {
```
Spreeuw, line 126: there's a lot of code duplication here with the regenerate code, I suggest to merge the two and apply exceptions (differences) where necessary

Spreeuw (changes requested):

assets/js/order-script.js
```diff
@@ -27,49 +30,49 @@ jQuery(document).ready(function($) {
...
+	} );
 
-	$('#wpo_wcpdf-data-input-box').insertAfter('#woocommerce-order-data');
+	$( '#wpo_wcpdf-data-input-box' ).insertAfter('#woocommerce-order-data');
 	
 	// enable invoice number edit if user initiated
-	$( ".wpo-wcpdf-set-date-number, .wpo-wcpdf-edit-date-number, .wpo-wcpdf-edit-document-notes" ).on("click", function() {
-		$form = $(this).closest('.wcpdf-data-fields');
+	$( document ).on( 'click', '.wpo-wcpdf-set-date-number, .wpo-wcpdf-edit-date-number, .wpo-wcpdf-edit-document-notes', function() {
+		let form = $(this).closest('.wcpdf-data-fields');
```
Spreeuw, line 39: form is a jQuery object, we prefix all jQuery objects with $ in our plugin scripts

assets/js/order-script.js
```diff
@@ -27,49 +30,49 @@ jQuery(document).ready(function($) {
...
+			form.find(".editable").hide();
+			form.find(':input').attr('disabled', true);
+			form.closest('.wcpdf-data-fields').find('.wpo-wcpdf-document-buttons').hide();
 		}
-	});
+	} );
 
 	// delete document
-	$( ".wcpdf-data-fields .wpo-wcpdf-delete-document" ).on("click", function() {
+	$( document ).on( 'click', '.wcpdf-data-fields .wpo-wcpdf-delete-document', function() {
```
Spreeuw, line 56: Is there a performance difference? Normally you'd use this approach for elements that are dynamically added (e.g. not available in the DOM at page load). I don't think that's the case here. That said, I think you can also limit the scope of the trigger bind to a closer parent (available at page load)

assets/js/order-script.js
alexmigf: Because of the data loaded after saving.

Spreeuw: was this comment supposed to be to another thread?

alexmigf: Yes, sorry!

assets/js/order-script.js
alexmigf: Because after save I load the document data again with jQuery.

Spreeuw: In that case I would still limit the scope to a closer parent for better performance

assets/js/order-script.js
```diff
@@ -27,37 +30,37 @@ jQuery(document).ready(function($) {
...
+	$( '#wpo_wcpdf-data-input-box' ).insertAfter('#woocommerce-order-data');
 	
 	// enable invoice number edit if user initiated
-	$( ".wpo-wcpdf-set-date-number, .wpo-wcpdf-edit-date-number, .wpo-wcpdf-edit-document-notes" ).on("click", function() {
-		$form = $(this).closest('.wcpdf-data-fields-section');
-		if ($form.length == 0) { // no section, take overall wrapper
-			$form = $(this).closest('.wcpdf-data-fields');
-		}
+	$( '#wpo_wcpdf-data-input-box' ).on( 'click', '.wpo-wcpdf-set-date-number, .wpo-wcpdf-edit-date-number, .wpo-wcpdf-edit-document-notes', function() {
+		$form = $(this).closest('.wcpdf-data-fields');
```
Spreeuw, line 39: you've applied variable scoping using `let` in several changes above, is there a reason you haven't scoped this variable explicitly? Undeclared, it's a global variable by default, but you need to take into account that the browser may run in strict mode.

assets/js/order-script.js
```diff
@@ -78,48 +81,93 @@ jQuery(document).ready(function($) {
...
-					$(this).find('.document-action-success').show();
+					if( action == 'save' ) {
+						$form.find(".read-only").show();
+						$form.find(".editable").hide();
+						$form.find(':input').attr('disabled', true);
+						$form.find('.wpo-wcpdf-document-buttons').hide();
+
+						// update document DOM data
+						let wrapper = data.document + '-fields .wrapper';
+						$form.load( document.URL + ' .' + wrapper, function() {
```
Spreeuw, line 143: That's quite a big AJAX request and I'm not sure if this will always work as expected. It's quite simple though (which I really like), so maybe we should just try this out. If it causes any issues with specific setups we can always pass the HTML with the success response directly (which requires a bit more code separation server side but is much lighter in usage).

assets/js/order-script.js
```diff
@@ -78,48 +81,93 @@ jQuery(document).ready(function($) {
...
 				if ( response.success ) {
-					$(this).find('.document-action-success').show();
+					if( action == 'save' ) {
+						$form.find(".read-only").show();
+						$form.find(".editable").hide();
+						$form.find(':input').attr('disabled', true);
+						$form.find('.wpo-wcpdf-document-buttons').hide();
+
+						// update document DOM data
+						let wrapper = data.document + '-fields .wrapper';
```
Spreeuw, line 142: does this work correctly for orders with multiple credit notes? Since you're already getting the entire page already, maybe it's safer to simply replace the whole meta box content?

includes/class-wcpdf-admin.php
```diff
@@ -426,101 +428,110 @@ public function output_number_date_edit_fields( $document, $data ) {
...
-			<!-- Document Notes -->
-			<?php if( array_key_exists( 'notes', $data ) ) : ?>
-
-				<?php do_action( 'wpo_wcpdf_meta_box_before_document_notes', $document, $document->order ); ?>
-
-				<section class="wcpdf-data-fields-section notes">
-					<p class="form-field form-field-wide">
-						<div>
-							<span><strong><?= $data['notes']['label']; ?></strong></span>
-							<span class="wpo-wcpdf-edit-document-notes dashicons dashicons-edit"></span>
```
Spreeuw, line 496: Did you remove this by accident? I can no longer edit the notes without editing the number and date, which means that the invoice number format may change for example.

includes/class-wcpdf-admin.php
```diff
@@ -798,26 +816,56 @@ public function regenerate_document() {
...
+					$document->regenerate( $order, $document_data );
+
+					$response      = array(
+						'message' => $document->get_type()." regenerated",
+					);
+
+				// on save
+				} elseif( $action_type == 'save' ) {
+					$form_data     = stripslashes_deep( $form_data );
+					$document_data = $this->process_order_document_form_data( $form_data, $document->slug );
```
Spreeuw, line 840: I think the data processing part should be the same for both action types, so instead of applying the above two lines, move these lines out of the if-statement: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/ddb43fca5712729524557d2373098a3a085cc774/includes/class-wcpdf-admin.php#L824-L830

Spreeuw (changes requested):

includes/class-wcpdf-admin.php
alexmigf: I thought that in terms of experience would be better for the user to just click one time and have all the options there, but I can revert that.

assets/js/order-script.js
alexmigf: Sounds good to me :+1:

YordanSoares:
✔️  @alexmigf I just tested your latest changes, and it works well.

In this screenshot you can see that I save a new note, and then delete it:

testing-new-invice-notes-buttons

includes/class-wcpdf-admin.php
Spreeuw: clicking edit for the document should reveal both notes and numbers, clicking edit for the notes should reveal only the notes. I thought it was like that in the current release, if not, it should be.

Spreeuw: Ok, I just checked, and indeed, the notes requires an additional click in the current release. That wasn't the point of my comment though, rather that you should be able to add a note without resetting number formatting.

alexmigf: > clicking edit for the document should reveal both notes and numbers, clicking edit for the notes should reveal only the notes. I thought it was like that in the current release, if not, it should be.

That's how it is working in the last changes.

assets/js/order-script.js
```diff
@@ -78,48 +96,94 @@ jQuery(document).ready(function($) {
...
+					if( action == 'save' ) {
+						$form.find(".read-only").show();
+						$form.find(".editable").hide();
+						$form.find(".editable-notes").hide();
+						$form.find(':input').attr('disabled', true);
+						$form.find('.wpo-wcpdf-document-buttons').hide();
+
+						// update document DOM data
+						$form.closest('#wpo_wcpdf-data-input-box').load( document.URL + ' #wpo_wcpdf-data-input-box .postbox-header, #wpo_wcpdf-data-input-box .inside', function() {
+							$('#wpo_wcpdf-data-input-box .notice').show();
```
Spreeuw, line 159: instead of preloading the success notice, you could also add a query param to the URL? That would also allow for more specific notices.

Spreeuw (changes requested):
very nice UI improvements 👍 
Some suggestions for minor tweaks.

assets/js/order-script.js
```diff
@@ -78,48 +96,94 @@ jQuery(document).ready(function($) {
...
+						});
+					} else if( action == 'regenerate' ) {
+						$form.find('.document-action-success').show();
```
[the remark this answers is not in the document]
alexmigf: The problem is, I added a notice because if choose to show the green check mark `.document-action-success`, it will display it in every document.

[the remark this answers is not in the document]
Spreeuw: I don't think we're talking about the same thing. I meant reloading the document data pane as a whole. The success notice could be handled with the same functions you use to handle the save notice (see other comment)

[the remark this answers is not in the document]
alexmigf: Should I use the notice also for the regenerate? Because if I apply it to the save it will show for every document at the same time.

Spreeuw:
I just noticed the **Notes (printed in the invoice):** label got left out in edit mode!

alexmigf:
@Spreeuw I see, I will fix

includes/class-wcpdf-admin.php
```diff
@@ -816,7 +809,21 @@ public function ajax_save_regenerate_document() {
 				}
 			}
 		}
-		$form_data     = stripslashes_deep( $form_data );
+		$form_data       = stripslashes_deep( $form_data );
+
+		// prepare notice message
+		$notice_types    = array(
+			'saved'       => array(
+				'success' => __( 'saved', 'woocommerce-pdf-invoices-packing-slips' ),
```
Spreeuw, line 811: for good translatability, these should be the full strings

includes/class-wcpdf-admin.php
```diff
@@ -782,44 +788,90 @@ public function regenerate_document() {
...
+
+		// notice messages
+		$notice_messages = array(
+			'saved'       => array(
+				'success' => __( 'Document data saved!', 'woocommerce-pdf-invoices-packing-slips' ),
+				'error'   => __( 'An error ocurred while saving the document data!', 'woocommerce-pdf-invoices-packing-slips' ),
+			),
+			'regenerated' => array(
+				'success' => __( 'Document data regenerated!', 'woocommerce-pdf-invoices-packing-slips' ),
+				'error'   => __( 'An error ocurred while regenerating the document data!', 'woocommerce-pdf-invoices-packing-slips' ),
```
Spreeuw, line 816: It's not the data that is regenerated but the document, so you can leave out 'data'.

Spreeuw (approved):
One small change and and 🚀 ⭐

YordanSoares:
✔️ @alexmigf I just tested your latest changes, and it works well.

In this screenshot you can see that I save a new note, and then delete it:

document-notes

alexmigf:
@YordanSoares did you test the other actions? Edit, regenerate, etc

YordanSoares:
Sorry @alexmigf I didn't realized that you also added changes to those action too.

✔️ I tested the invoices notes addition and removal, and the document data editing and regeneration, and all appears to be working well:

https://user-images.githubusercontent.com/38109855/116892373-c4b1e300-abfd-11eb-8f6a-c452dd81514b.mp4

alexmigf:
@YordanSoares can you also test other documents like Packing Slip, Proforma and Credit Notes?

I did a test and all seems good! It's just to make sure everything is fine.

YordanSoares:
@alexmigf sure!

✔️ All appear to be fine with other documents too :)

Here is a screenshot recording with my latest tests:

https://user-images.githubusercontent.com/38109855/116901649-2ecf8580-ac08-11eb-85cf-dc68c08a980f.mp4


## Some remarks/suggestions:

- I just realized that we don't have a regenerate button for packing slips. Does this mean that packing slips are always updated? (if so, I didn't know).
- Don't you think it's a good idea to add a notice to confirm the document deletion too?

Spreeuw (approved):
Nice work!

alexmigf:
@YordanSoares can you do a final test? (delete should work the same now too)

YordanSoares (approved):
✔️ All appear to be working fine :)

save-cancel-button-test

PR #176, opened 2021-05-05
fix missing directory separator in distributed fonts path

Spreeuw:
When the font folder in the temporary wcpdf folder is cleared (but not deleted) or when the font cache files (`.ufm.php`  and `.afm.php`) are deleted for whatever reason (most likely as a result of a false positive in a virus scan), the fonts can no longer be properly loaded and texts appear stacked.

dompdf actually provides a built-in fallback to 'distributed' (bundled) fonts if fonts cannot be loaded from the fonts path, but our version was missing a slash (`DIRECTORY_SEPARATOR`) in the path, so that this fallback would work either.

This PR fixes both the fallback and also adds the extra fonts we bundle by default to it


Spreeuw:
some example issues:
https://wordpress.org/support/topic/pdf-rendering-broken/
https://wordpress.org/support/topic/several-lines-of-text-in-one-line/
https://wordpress.org/support/topic/overlapping-text-in-pdf/
https://wordpress.org/support/topic/overlapping-text-8/

YordanSoares (approved):
✔️ l performed the tests and it works very well :)

https://user-images.githubusercontent.com/38109855/117306104-64b17b80-ae4d-11eb-8c03-86049d42d800.mp4

Issue #177, opened 2021-05-13
Improve internationalization of the plugin (i18n)


YordanSoares:
- [ ] Add numbered placeholders.
- [ ] Add comments for translators.

Spreeuw:
@YordanSoares I closed this but I was preparing a new release and realized that the POT and bundled PO/MO translations have not been updated yet. I think we should do this before releasing, can you do that? (perhaps open a new issue/PR)

YordanSoares:
Sure @Spreeuw! It's done: #181

PR #178, opened 2021-05-13
Improve internationalization of the plugin (i18n)

YordanSoares:
Closes: #177

### DIFF
--- a/includes/views/attachment-settings-hint.php
+++ b/includes/views/attachment-settings-hint.php
@@ -19,6 +19,7 @@ $invoice_settings_url = add_query_arg( array(
 </style>
 <!-- <div id="message" class="updated woocommerce-message"> -->
 <div class="wcpdf-attachment-settings-hint">
-	<?php printf(__( "It looks like you haven't setup any email attachments yet, check the settings under <b>%sDocuments > Invoice%s</b>", 'woocommerce-pdf-invoices-packing-slips' ), '<a href="'.$invoice_settings_url.'">', '</a>'); ?><br>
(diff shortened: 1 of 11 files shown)


includes/views/attachment-settings-hint.php
```diff
@@ -19,6 +19,7 @@
 </style>
 <!-- <div id="message" class="updated woocommerce-message"> -->
 <div class="wcpdf-attachment-settings-hint">
-	<?php printf(__( "It looks like you haven't setup any email attachments yet, check the settings under <b>%sDocuments > Invoice%s</b>", 'woocommerce-pdf-invoices-packing-slips' ), '<a href="'.$invoice_settings_url.'">', '</a>'); ?><br>
+<?php /* translators: <a> tags */ ?>
+	<?php printf(__( 'It looks like you haven\'t setup any email attachments yet, check the settings under <b>%1$sDocuments > Invoice%2$s</b>', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="'.$invoice_settings_url.'">', '</a>'); ?><br>
```
Spreeuw, line 23: I prefer using the double quotes if there's a single quote in the text TBH

YordanSoares: The issue in this case is that the numbered placeholders uses the dollar sign. Do you prefer using double quotes and escape the dollar sign instead?

Spreeuw: I see, I didn't consider that. Let's leave it as you have it now then.

PR #190, opened 2021-06-16
Displays the bulk actions for enabled documents only.

### LINKED ISSUE #189: Display bulk actions for enabled documents only
Currently we are displaying the bulk actions for every document, even if it's disabled.

alexmigf:
closes #189

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -562,7 +562,9 @@ class Admin {
 		$actions = array();
 		$documents = WPO_WCPDF()->documents->get_documents();
 		foreach ($documents as $document) {
-			$actions[$document->get_type()] = "PDF " . $document->get_title();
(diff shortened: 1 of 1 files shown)


Spreeuw:
I'm not sure why this is not working already for you, because it works fine with current master for me!

Here we call the documents:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/6faba46c20c70109b16a342bc47fee28976afbd5/includes/class-wcpdf-admin.php#L563-L566

and the call to `WPO_WCPDF()->documents->get_documents()` not having the argument set, defaults to `enabled`:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/6faba46c20c70109b16a342bc47fee28976afbd5/includes/class-wcpdf-documents.php#L69-L86

alexmigf:
You're right!

PR #195, opened 2021-06-18
Open PDF on My Account page in a new browser tab/window

### LINKED ISSUE #194: Follow the 'View PDF' option on the My Account page
When the '_Open the PDF in a new browser tab/window_' option is selected in the '_How do you want to view the PDF_' dropdown the PDF opens in a new tab/window in the backend but not on the customer's My Account page. We can use some jQuery to add a `target="_blank"` attribute.

Terminator-Barbapapa:
Closes #194

### DIFF
--- a/includes/class-wcpdf-frontend.php
+++ b/includes/class-wcpdf-frontend.php
@@ -16,6 +16,7 @@ class Frontend {
 	function __construct()	{
 		add_filter( 'woocommerce_my_account_my_orders_actions', array( $this, 'my_account_pdf_link' ), 10, 2 );
 		add_filter( 'woocommerce_api_order_response', array( $this, 'woocommerce_api_invoice_number' ), 10, 2 );
+		add_action( 'wp_enqueue_scripts', array( $this, 'open_my_account_pdf_link_on_new_tab' ), 999 );
 		add_shortcode( 'wcpdf_download_invoice', array($this, 'download_invoice_shortcode') );
 	}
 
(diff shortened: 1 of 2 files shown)


includes/class-wcpdf-frontend.php
```diff
@@ -63,6 +64,28 @@ public function my_account_pdf_link( $actions, $order ) {
...
 
+	/**
+	 * Open PDF on My Account page in a new browser tab/window
+	 */
+	public function open_my_account_pdf_link_on_new_tab() {
+		if ( function_exists( 'is_account_page' ) && is_account_page() ) {
+			if ( $general_settings = get_option( 'wpo_wcpdf_settings_general' ) ) {
+				if ( isset( $general_settings['download_display'] ) && $general_settings['download_display'] == 'display' ) {
+					wp_add_inline_script( 'jquery', "
+						jQuery( function( $ ) {
```
Spreeuw, line 75: While I think it's beneficial for the site speed to include this inline, for development purposes it's maybe nicer to keep it in a file. How about using `file_get_contents()` (or `ob_start()` + `include()` + `ob_get_clean()`) instead?

Terminator-Barbapapa: Should be resolved in [this commit](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/195/commits/23635ea03740cca971bbc070ecefb9f4f1f0f099).

includes/class-wcpdf-frontend.php
```diff
@@ -63,6 +64,21 @@ public function my_account_pdf_link( $actions, $order ) {
...
 	}
 
+	/**
+	 * Open PDF on My Account page in a new browser tab/window
+	 */
+	public function open_my_account_pdf_link_on_new_tab() {
+		if ( function_exists( 'is_account_page' ) && is_account_page() ) {
+			if ( $general_settings = get_option( 'wpo_wcpdf_settings_general' ) ) {
+				if ( isset( $general_settings['download_display'] ) && $general_settings['download_display'] == 'display' ) {
+					if ( function_exists( 'file_get_contents' ) && !empty( $file = file_get_contents( WPO_WCPDF()->plugin_url() . '/assets/js/my-account-link.js' ) ) ) {
```
Spreeuw, line 74: This doesn't work with `allow_url_fopen` to false (or other restrictions), and is slow too (loading via the interwebs instead of locally). Using `plugin_path()` instead of `plugin_url()` fixes that
```php
if ( function_exists( 'file_get_contents' ) && $script = file_get_contents( WPO_WCPDF()->plugin_path() . '/assets/js/my-account-link.js' ) ) {
	wp_add_inline_script( 'jquery', $script );
}
```

Terminator-Barbapapa: Fixed in last commit.

PR #207, opened 2021-08-23
Returns the setting first key if 'default' is empty

alexmigf:
closes https://github.com/wpovernight/woocommerce-pdf-ips-pro/issues/98

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -621,11 +621,10 @@ abstract class Order_Document {
 	public function get_settings_text( $settings_key, $default = false, $autop = true ) {
 		if( array_key_exists( 'default', $this->settings[$settings_key] ) ) {
 			$text = wptexturize( trim( $this->settings[$settings_key]['default'] ) );
-		} else {
+		} elseif( ! empty( $this->settings[$settings_key] ) && is_array( $this->settings[$settings_key] ) ) {
 			$text = wptexturize( trim( reset( $this->settings[$settings_key] ) ) );
-			if( ! $text ) {
-				$text = $default;
-			}
+		} else {
+			$text = $default;
 		}
 
 		if ( $autop === true ) {
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -619,14 +619,19 @@ abstract class Order_Document {
 	}
 
 	public function get_settings_text( $settings_key, $default = false, $autop = true ) {
-		if ( !empty( $this->settings[$settings_key]['default'] ) ) {
+		if( array_key_exists( 'default', $this->settings[$settings_key] ) ) {
 			$text = wptexturize( trim( $this->settings[$settings_key]['default'] ) );
-			if ($autop === true) {
-				$text = wpautop( $text );
-			}
 		} else {
-			$text = $default;
+			$text = wptexturize( trim( reset( $this->settings[$settings_key] ) ) );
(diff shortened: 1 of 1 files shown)


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -619,14 +619,19 @@ public function header_logo() {
...
 	public function get_settings_text( $settings_key, $default = false, $autop = true ) {
-		if ( !empty( $this->settings[$settings_key]['default'] ) ) {
+		if( array_key_exists( 'default', $this->settings[$settings_key] ) ) {
 			$text = wptexturize( trim( $this->settings[$settings_key]['default'] ) );
-			if ($autop === true) {
-				$text = wpautop( $text );
-			}
 		} else {
-			$text = $default;
+			$text = wptexturize( trim( reset( $this->settings[$settings_key] ) ) );
```
Spreeuw, line 625: you should check if `$this->settings[$settings_key]` exists at all, and if it's an `array` before calling `reset`, otherwise this might throw warnings.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -619,14 +619,19 @@ public function header_logo() {
 	}
 
 	public function get_settings_text( $settings_key, $default = false, $autop = true ) {
-		if ( !empty( $this->settings[$settings_key]['default'] ) ) {
+		if( array_key_exists( 'default', $this->settings[$settings_key] ) ) {
```
Spreeuw, line 622: this changes the current behavior of using the default when nothing is filled in. While that could be an improvement, it's a breaking change so I'd prefer to stick to the current method. You'll need both.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -619,14 +619,19 @@ public function header_logo() {
...
 			$text = wptexturize( trim( $this->settings[$settings_key]['default'] ) );
-			if ($autop === true) {
-				$text = wpautop( $text );
-			}
 		} else {
-			$text = $default;
+			$text = wptexturize( trim( reset( $this->settings[$settings_key] ) ) );
+			if( ! $text ) {
+				$text = $default;
```
Spreeuw, line 627: `$default` should be the fallback for both options (it works like that now, but no longer if you change the first condition, so it's safer as a third `else` condition.

Spreeuw (changes requested):

includes/documents/abstract-wcpdf-order-document.php
Spreeuw: you marked this as resolved but you didn't reintroduce the `!empty( $this->settings[$settings_key]['default'] )` condition.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -621,11 +621,10 @@ public function header_logo() {
 	public function get_settings_text( $settings_key, $default = false, $autop = true ) {
 		if( array_key_exists( 'default', $this->settings[$settings_key] ) ) {
 			$text = wptexturize( trim( $this->settings[$settings_key]['default'] ) );
-		} else {
+		} elseif( ! empty( $this->settings[$settings_key] ) && is_array( $this->settings[$settings_key] ) ) {
```
Spreeuw, line 624: if `'default'` is `empty`, it will get to this step, and still meet both requirements. But to keep the current behavior it should then use `$default`.

Spreeuw (changes requested):

Spreeuw (approved):

PR #209, opened 2021-09-24
Template path portability

Spreeuw:
This PR introduces a new method of storing the selected template in the database - instead of a relative path (e.g. `wp-content/plugins/woocommerce-pdf-invoices-packing-slips/templates/Simple`, it stores an "id" (e.g. `default/Simple`). When the template path is requested, this id is then matched with a list of paths that is maintained separately.
The "id" is in the form of `{$group}/{$template_name}`, where `group` is the array key that was used when registering the template base path (the folder that holds the actual template subfolders).

This should improve portability by solving a number of edge-case issues:

- Templates are stored outside of the `WP_CONTENT_DIR` (or `ABSPATH`) and conversion from relative path to absolute path fails
- Templates are symlinks (in certain setups using Bedrock)
- Template location is stored (for some reason) as an absolute path and the base path changes (versioning etc).

Working with a 'live' list of paths does come with the downside that when the path is needed early on in the process, any (third party) functions hooking into our filters may not have been applied yet, and the path is therefor not available yet. To solve this problem, I have implemented a 'cached' list of paths that is updated when:

- The settings are saved/updated ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L45))
- Later retrieval of the actual 'live' list returns a different list than the cached one. ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L295-L297))
- One or more paths in the cached list no longer exists ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L306-L329))

Migration of old paths occurs when:
- The plugin is updated to the new version (Lifecycle methods, [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-install.php#L350-L355))
- The settings page is loaded ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L47))

As a safety precaution/fallback the plugin also still accepts paths as the setting ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L227-L241))

Known (minor) issues:

- Premium Templates before 2.13.1 registered the template search path relatively late (`init` with priority `9`), while there was also a relatively early call for the template-functions.php file ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/60516f1c081404cf95d9dc5e03754941cfbf4e65/includes/class-wcpdf-main.php#L31-L35), which is at `plugins_loaded` with priority `9`). This should be corrected by the cached path list update routines outlined above.
- Premium Templates before 2.13.1 uses the value of the selected template to dynamically change the hints for creating a custom template. Since the value no longer holds the path but an "id", the result is that rather than the actual path to copy from, that "id" is shown, which could lead to confusion.

Both of these issues are addressed in 2.13.1 but we should take into account that people may be using older versions.

A **new feature** that I have introduced is the possibility to use the same name "Simple" when copying the template to a (child-) theme. This will show up as "Simple (Theme)" in the list.

## Testing
Because this change touches on one of the most crucial settings of the plugin (telling it where to find the template files), we should very carefully test as many scenarios as possible. This includes (but is certainly not limited to):

- Simply upgrading the old plugin to the new version - checking that the path is migrated to the id, and the PDF can still be generated everywhere
- Upgrading with a setup that was using the official Simple template and had a copy of the Simple template in the theme (this would not be used in the current release because it required a unique name).
- Usage in combination with the Premium Templates extension (both current version 2.13.1 as well as the previous 2.13.0)
-  ... any possible edge case you can think of!

**Note:** There is no 'downgrade' routine available in a current release for the settings migration, so once migrated to the "id", when you go back from the version from this PR/branch to the current release, the template setting will be lost (and should reset to "Simple").

### DIFF
--- a/includes/class-wcpdf-assets.php
+++ b/includes/class-wcpdf-assets.php
@@ -124,7 +124,8 @@ class Assets {
 				'wpo-wcpdf-admin',
 				'wpo_wcpdf_admin',
 				array(
-					'ajaxurl'		=> admin_url( 'admin-ajax.php' ),
+					'ajaxurl'        => admin_url( 'admin-ajax.php' ),
+					'template_paths' => WPO_WCPDF()->settings->get_installed_templates(),
 				)
 			);
 
--- a/includes/class-wcpdf-documents.php
+++ b/includes/class-wcpdf-documents.php
@@ -59,6 +59,8 @@ class Documents {
 
 		// Allow plugins to add their own documents
 		$this->documents = apply_filters( 'wpo_wcpdf_document_classes', $this->documents );
+
+		do_action( 'wpo_wcpdf_init_documents' );
 	}
 
 	/**
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -347,6 +347,12 @@ class Install {
 			update_option( 'wpo_wcpdf_settings_debug', $debug_settings );
 		}
 
+		// 2.10.0-dev: migrate template path to template ID
+		if ( version_compare( $installed_version, '2.10.0-dev', '<' ) ) {
+			if ( ! empty( WPO_WCPDF()->settings ) && is_callable( array( WPO_WCPDF()->settings, 'maybe_migrate_template_paths' ) ) ) {
+				WPO_WCPDF()->settings->maybe_migrate_template_paths();
+			}
+		}
 	}
 
 	/**
--- a/includes/class-wcpdf-settings-general.php
+++ b/includes/class-wcpdf-settings-general.php
@@ -64,7 +64,7 @@ class Settings_General {
 				'args'		=> array(
 					'option_name'	=> $option_name,
 					'id'			=> 'template_path',
-					'options' 		=> $this->find_templates(),
+					'options' 		=> $this->get_installed_templates_list(),
 					/* translators: 1,2. template paths */
 					'description'	=> sprintf( __( 'Want to use your own template? Copy all the files from <code>%1$s</code> to your (child) theme in <code>%2$s</code> to customize them' , 'woocommerce-pdf-invoices-packing-slips' ), $plugin_template_path, $theme_template_path),
 				)
@@ -272,6 +272,29 @@ class Settings_General {
 				}
 			}
 		}
+	}
+
+	public function get_installed_templates_list() {
+		$installed_templates = WPO_WCPDF()->settings->get_installed_templates();
+		$template_list = array();
+		foreach ( $installed_templates as $path => $template_id ) {
+			$template_name = basename( $template_id );
+			$group = dirname( $template_id );
+			switch ( $group ) {
+				case 'theme':
+					$template_name = sprintf( '%s (%s)', $template_name, __( 'Theme', 'woocommerce-pdf-invoices-packing-slips' ) );
+					break;
+				case 'default':
+				case 'premium_plugin':
+					// no suffix
+					break;
+				default:
+					$template_name = sprintf( '%s (%s)', $template_name, __( 'Custom', 'woocommerce-pdf-invoices-packing-slips' ) );
+					break;
+			}
+			$template_list[$template_id] = $template_name;
+		}
+		return $template_list;
 	}
 
 	/**
--- a/includes/class-wcpdf-settings-general.php
+++ b/includes/class-wcpdf-settings-general.php
@@ -281,13 +281,11 @@ class Settings_General {
 			$template_name = basename( $template_id );
 			$group = dirname( $template_id );
 			switch ( $group ) {
-				case 'theme':
-					$template_name = sprintf( '%s (%s)', $template_name, __( 'Theme', 'woocommerce-pdf-invoices-packing-slips' ) );
-					break;
 				case 'default':
 				case 'premium_plugin':
 					// no suffix
 					break;
+				case 'theme':
 				default:
 					$template_name = sprintf( '%s (%s)', $template_name, __( 'Custom', 'woocommerce-pdf-invoices-packing-slips' ) );
 					break;
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -283,7 +283,7 @@ class Settings {
 
 		if ( empty( $installed_templates ) ) {
 			// fallback to Simple template for servers with glob() disabled
-			$simple_template_path = $template_paths['default'] . 'Simple';
+			$simple_template_path = $this->normalize_path( $template_paths['default'] . 'Simple' );
 			$installed_templates[$simple_template_path] = 'default/Simple';
 		}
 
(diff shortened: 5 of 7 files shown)


alexmigf (approved):
### My tests

- Simply upgrading the old plugin to the new version - checking that the path is migrated to the id, and the PDF can still be generated everywhere :heavy_check_mark: 
- Upgrading with a setup that was using the official Simple template and had a copy of the Simple template in the theme (this would not be used in the current release because it required a unique name). :heavy_check_mark: 
- Usage in combination with the Premium Templates extension (both current version 2.13.1 as well as the previous 2.13.0) :heavy_check_mark: 
- Symlinked Simple theme on upgrade :heavy_check_mark:

dpeyou (approved):
- [x] Simply upgrading the old plugin to the new version - checking that the path is migrated to the id, and the PDF can still be generated everywhere 

- [x] Upgrading with a setup that was using the official Simple template and had a copy of the Simple template in the theme (this would not be used in the current release because it required a unique name).

- [x] Usage in combination with the Premium Templates extension (both current version 2.13.1 as well as the previous 2.13.0)

YordanSoares:
### Testing results:

- ✔️ Simply upgrading the old plugin to the new version - checking that the path is migrated to the id, and the PDF can still be generated everywhere.
- ✔️ Upgrading with a setup that was using the official Simple template and had a copy of the Simple template in the theme (this would not be used in the current release because it required a unique name).
- ✔️ Usage in combination with the Premium Templates extension (both current version 2.13.1 as well as the previous 2.13.0).

### Some remarks:

1. Don't you think it's clearer to use **Custom template** instead of **Theme**, after the custom templates names, to avoid confusion?

image

2. I realized that the template description that indicates the instruction to create a custom template works great for the **Premium Templates**, but still appear when a custom template is already selected. What do you think about hiding this text in these cases or adding a phrase that makes more sense in this context, such as: "This is a custom template. Do you want to create another custom template?", followed by the same instructions or a link to the documentation article?

In addition, if you select **Simple** (default), the description refers to **Simple Premium**, when **Premium Templates** extension is enabled:

pdf-templates-description

alexmigf:
I think remark number 2 is because you have the Premium Templates extension enable to force you to copy that instead of the Simple, because this way is compatible with the customizer.

YordanSoares:
@alexmigf That makes sense! 👍

Terminator-Barbapapa (commented):
Right now when you upgrade the plugin and do not change and save any settings, there is no cached list stored in the DB. Is this the desired behaviour?

Terminator-Barbapapa:
### My tests

- Simply upgrading the old plugin to the new version - checking that the path is migrated to the id, and the PDF can still be generated everywhere :heavy_check_mark: 
- Upgrading with a setup that was using the official Simple template and had a copy of the Simple template in the theme (this would not be used in the current release because it required a unique name). :heavy_check_mark: 
- Usage in combination with the Premium Templates extension (both current version 2.13.1 as well as the previous 2.13.0) :heavy_check_mark: 
- Loading custom PDF document template :heavy_check_mark:

Spreeuw:
> Right now when you upgrade the plugin and do not change and save any settings, there is no cached list stored in the DB. Is this the desired behaviour?

This should, in theory (!), be done by the function hooked to the option update ([`update_option_{$option}`](https://github.com/WordPress/WordPress/blob/70f1109c9fdc5ba8dd67664db404453ffeddffee/wp-includes/option.php#L515-L527)):
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L45

Since we use `update_option()` to write the changes:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings.php#L395

So if that's not happening for some reason, I'm not sure why... It won't hurt to add it to the migration procedure anyway (worst case scenario it simply saves twice), but I'm a little worried it may not catch settings updates fired from other contexts such as WP CLI, so it would be nice to find the root cause.

Did this happen with custom templates as well as premium templates and free templates? Or did you only test one of these 3?

Spreeuw:
@YordanSoares 
> Don't you think it's clearer to use **Custom template** instead of **Theme**, after the custom templates names, to avoid confusion?

I wanted to distinguish between templates that were found in the theme (which could theoretically have been shipped with the theme, although I haven't seen that in the wild) and ones that were really custom:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/84ba7b8e8d92fc7516d89a0bb9bfacc96172b30c/includes/class-wcpdf-settings-general.php#L284-L293

But I agree that's not a very important different and "Custom" may be a nicer catchall, so I merged them in 6d8e43b :
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/6d8e43b00454327ebc4c68cd18903ed042ff549c/includes/class-wcpdf-settings-general.php#L288-L291


> What do you think about hiding this text in these cases or adding a phrase that makes more sense in this context, such as: "This is a custom template. Do you want to create another custom template?", followed by the same instructions or a link to the documentation article?

That's indeed a nice UX improvement! For this PR I want to focus primarily on the newly introduced portability from this PR. (the issue you point out has in fact been present since the very first version of the plugin).

Terminator-Barbapapa:
@Spreeuw It looks like I made a mistake while testing. I was switching back and forth between branches, reloading the general settings page and was removing the `wpo_wcpdf_installed_template_paths` option in between from the DB. 

For some reason the `template_path` value in the `wpo_wcpdf_settings_general` option was empty in my setup. Not sure how that happened exactly. But I believe this prevented the `wpo_wcpdf_installed_template_paths` option from being saved.

I did the whole routine again (after lunch which probably makes a lot of difference) and it seems to work without issue this time.

YordanSoares:
> I wanted to distinguish between templates that were found in the theme (which could theoretically have been shipped with the theme, although I haven't seen that in the wild) and ones that were really custom. But I agree that's not a very important different and "Custom" may be a nicer catchall, so I merged them in 6d8e43b

👍 

> That's indeed a nice UX improvement! For this PR I want to focus primarily on the newly introduced portability from this PR. (the issue you point out has in fact been present since the very first version of the plugin).

That's right. I just took advantage to mentioning it, since this PR touch this description :)

YordanSoares (approved):

Terminator-Barbapapa (approved):

Spreeuw:
> For some reason the `template_path` value in the `wpo_wcpdf_settings_general` option was empty in my setup. Not sure how that happened exactly. But I believe this prevented the `wpo_wcpdf_installed_template_paths` option from being saved.

This happens when you deactivate the Premium Templates extension, to prevent premium templates to be loaded (the file paths still exist after all) only to crash when the PDF is being generated (because the template functions are not loaded because the plugin itself isn't). There may be other reasons the setting becomes empty, but I think this is the most likely one.
I guess we could improve this when this branch is merged, by setting it to `default/Simple` (now that it is no longer tied to paths, yay!).
I just realized the reactivating the Premium Templates, it will not switch from 'empty' to 'Simple Premium' (only from 'Simple' to 'Simple Premium') and that the plugin install procedure is also still path based. I'll open a separate issue for that.

Terminator-Barbapapa:
> This happens when you deactivate the Premium Templates extension

That makes sense. This is probably what happened as I indeed deactivated PT to test with the `2.13.0` version.

Issue #210, opened 2021-09-27
Undefined variable $order when using [wcpdf_download_invoice] shortcode in PHP 8


YordanSoares:
Testing the `[wcpdf_download_invoice]` shortcode in a staging site with PHP 8, I received this warning when I put it in the **Shop** page:

> ( ! ) Warning: Undefined variable $order in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips-template-path-portability\includes\class-wcpdf-frontend.php on line 127

This is not the place where this shortcode should be used, but I'm testing its behavior in a place outside **My Account > View order** and **Checkout > Order Received** page (or when a specific order ID isn't passed).

PR #211, opened 2021-09-27
Undefined variable $order when using [wcpdf_download_invoice] shortcode in PHP 8

YordanSoares:
Closes: #210

### DIFF
--- a/includes/class-wcpdf-frontend.php
+++ b/includes/class-wcpdf-frontend.php
@@ -116,7 +116,6 @@ class Frontend {
 		global $wp;
 
 		// Get $order
-		$order = '';
 		if( is_checkout() && !empty(is_wc_endpoint_url('order-received')) && empty($values['order_id']) && isset($wp->query_vars['order-received']) ) {
 			$order = wc_get_order( $wp->query_vars['order-received'] );
 		} elseif( is_account_page() && !empty(is_wc_endpoint_url('view-order')) && empty($values['order_id']) && isset($wp->query_vars['view-order']) ) {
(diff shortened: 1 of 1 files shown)


includes/class-wcpdf-frontend.php
```diff
@@ -116,6 +116,7 @@ public function download_invoice_shortcode( $atts ) {
 		global $wp;
 
 		// Get $order
+		$order = '';
```
Spreeuw, line 119: `$order` is never a string, so setting it to an empty string can lead to type-related issues down the road.

The real problem is in this check:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c9f467de65f49165007a10c9d5039b4be4e7c017/includes/class-wcpdf-frontend.php#L127

That is also the line that the error was referring to. The code is checking `is_object` before checking if it exists at all. Reversing that will solve the error. Furhermore, `!isset()` is not necessary since `!isset()` is a subset of `empty()`.

Spreeuw (changes requested):

includes/class-wcpdf-frontend.php
YordanSoares: Is this what you were referring to: 27915ee9d110e874e73c7ea70c39e701cd4bc204?

Spreeuw (approved):

includes/class-wcpdf-frontend.php
Spreeuw: Yes!

PR #223, opened 2021-10-29
Tweak: improved mb-string fallback using Symfony polyfill

Spreeuw:
Replacing our incomplete mb-string replacement (from the legacy dompdf 0.6) with the more complete Symfony polyfill.

This will prevent issues with special characters not showing when the MB String PHP extension is not installed.

For testing, make sure to disable all plugins except WooCommerce & WCPDF, because this is a popular polyfill used in several plugins (including our own WooCommerce Ultimate Barcodes).


alexmigf (approved):
### With this branch (with mbstring installed and uninstalled)

Captura de ecrã de 2021-11-10 15-05-28

### Using current master with mbstring uninstalled

Captura de ecrã de 2021-11-10 15-06-58

Spreeuw:
@YordanSoares can you test this too?

YordanSoares (approved):
✔️  **With this branch (with mbstring installed and uninstalled)**

image

❌  **With current master with mbstring uninstalled**

image

Issue #230, opened 2021-11-18
Load custom fonts in DOMPDF merging with local or vendor fonts


alexmigf:
Currently we pass a cache file `dompdf_font_family_cache.php` directly to the `vendor/dompdf/dompdf/lib/fonts` directory and also our custom fonts as well. This is not ideal because every time we need to upgrade DOMPDF we need to copy this.

#### Things we should address:

* Move our custom fonts to a dedicated plugin directory, I suggest: `assets/fonts`
* Copy the custom fonts to the local (uploads folder) fonts
* Load the uploads fonts directory as the `fontDir`

Spreeuw:
More specifically, this is referring to the following procedure, which is triggered 1) when the tmp folder is initiated ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1b97c0589293bc36167f5ae99d9f8684cdabdc21/includes/class-wcpdf-main.php#L580-L583)) and 2) when the plugin is updated ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1b97c0589293bc36167f5ae99d9f8684cdabdc21/includes/class-wcpdf-install.php#L206)):
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1b97c0589293bc36167f5ae99d9f8684cdabdc21/includes/class-wcpdf-main.php#L704-L730

dompdf actually has merging code built-in:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1b97c0589293bc36167f5ae99d9f8684cdabdc21/vendor/dompdf/dompdf/src/FontMetrics.php#L123-L151

And I think we should use this as much as possible. In theory (I have not verified this with tests), with the above dompdf function, our own merge function shouldn't even be necessary because dompdf will merge it automatically.

However, when we introduce a third font source (our own 'custom' fonts that are not distributed with dompdf), we still need to merge them with the local fonts.

I think it could work like this:

1. We load our own custom font list in the in the `FontMetrics` class, then change the font folder to the local WCPDF tmp folder (using `setOptions()`) and call `saveFontFamilies()` so that the merged result is saved there.
2. Then, when we load dompdf as we do now, dompdf will automatically merge those fonts with the distributed fonts.

[...]

PR #232, opened 2021-11-19
Testing custom fonts load in DOMPDF

alexmigf:
#230


includes/class-wcpdf-main.php
```diff
@@ -588,6 +590,34 @@ public function init_tmp () {
...
+			'isFontSubsettingEnabled'	=> true,
+			'isHtml5ParserEnabled'		=> true,
+		);
+
+		// save custom fonts
+		$dompdf = new Dompdf( new Options( $options ) );
+		$dompdf->getFontMetrics()->saveFontFamilies();
+
+		// save local fonts
+		$options['fontDir'] = $options['fontCache'] = $this->get_tmp_path( 'fonts' );
```
Spreeuw, line 612: would this result in an infinite loop if `load_custom_and_local_fonts()` is called by `init_tmp()`? The goal is to make this method replace the current `copy_fonts()` method.

includes/class-wcpdf-install.php
```diff
@@ -80,6 +80,9 @@ protected function install() {
 			WPO_WCPDF()->main->init_tmp();
 		}
 
+		// load fonts
+		WPO_WCPDF()->main->load_custom_and_local_fonts();
```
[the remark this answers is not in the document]
alexmigf: I'm just testing

includes/class-wcpdf-main.php
alexmigf: Well, `copy_fonts` calls `get_tmp_path` also: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/2ebe7cb9be7248bfe1cb7a4ea81cb6037ac7404a/includes/class-wcpdf-main.php#L696-L697

Spreeuw: I see, so it's probably safe, but something to keep in mind.

assets/fonts/dompdf_font_family_cache.php
```diff
@@ -1,4 +1,5 @@
-<?php return array (
+<?php return function ($fontDir, $rootDir) {
```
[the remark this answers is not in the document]
alexmigf: Yep, this was the culprit

assets/fonts/lib/fonts/dompdf_font_family_cache.dist.php
```diff
@@ -0,0 +1,107 @@
+<?php
+$fontDir = trailingslashit( WPO_WCPDF()->plugin_path() ) . 'assets' . DIRECTORY_SEPARATOR . 'fonts';
```
Spreeuw, line 2: I would follow the dompdf naming and use `$distFontDir` instead, this prevents mixups
https://github.com/dompdf/dompdf/blob/8768448244967a46d6e67b891d30878e0e15d25c/lib/fonts/dompdf_font_family_cache.dist.php#L2

includes/class-wcpdf-main.php
```diff
@@ -588,6 +590,50 @@ public function init_tmp () {
...
 
+	public function load_custom_and_local_fonts() {
+		require WPO_WCPDF()->plugin_path() . '/vendor/autoload.php';
+
+		$options = array(
+			'defaultFont'				=> 'dejavu sans',
+			'tempDir'					=> $this->get_tmp_path( 'dompdf' ),
+			'logOutputFile'				=> $this->get_tmp_path( 'dompdf' ) . "/log.htm",
+			'fontDir'					=> $this->get_custom_fonts_path(),
+			'fontCache'					=> $this->get_custom_fonts_path(),
```
Spreeuw, line 601: I think the above two should be `$this->get_tmp_path( 'fonts' )`? After all that's what we want to write to!

includes/class-wcpdf-main.php
```diff
@@ -588,6 +590,50 @@ public function init_tmp () {
...
+			'isRemoteEnabled'			=> true,
+			'isFontSubsettingEnabled'	=> true,
+			'isHtml5ParserEnabled'		=> true,
+		);
+
+		if ( ! @is_dir( $this->get_dompdf_fonts_path() ) || ! wp_is_writable( $this->get_dompdf_fonts_path() ) ) {
```
Spreeuw, line 607: why are you checking this? We shouldn't write to this path anyway, only to our temp folder!

includes/class-wcpdf-main.php
```diff
@@ -588,6 +590,50 @@ public function init_tmp () {
...
+			$rootDir = $this->get_custom_fonts_path();
+		}
+
+		// save custom fonts
```
Spreeuw, line 611: we're not saving fonts, we're merging custom fonts cache with local fonts cache

includes/class-wcpdf-main.php
```diff
@@ -588,6 +590,50 @@ public function init_tmp () {
...
+		$custom_fonts_options = new Options( $options );
+		if( isset( $rootDir ) ) {
+			$custom_fonts_options->setRootDir( $rootDir );
+		}
+		$dompdf = new Dompdf( $custom_fonts_options );
+		$dompdf->getFontMetrics()->saveFontFamilies();
+
+		// save local fonts
```
Spreeuw, line 619: I'm not sure why this step is necessary. dompdf should merge the dist fonts (from `$rootDir`) with the local fonts (from `$fontDir`) automatically when it's creating the PDF, no?

includes/class-wcpdf-main.php
```diff
@@ -588,6 +590,50 @@ public function init_tmp () {
...
+		$dompdf->getFontMetrics()->setOptions( $local_fonts_options );
+		$dompdf->getFontMetrics()->saveFontFamilies();
+	}
+
+	public function get_custom_fonts_path() {
+		return trailingslashit( WPO_WCPDF()->plugin_path() ) . 'assets' . DIRECTORY_SEPARATOR . 'fonts';
+	}
+
+	public function get_dompdf_fonts_path() {
+		return trailingslashit( WPO_WCPDF()->plugin_path() ) . 'vendor' . DIRECTORY_SEPARATOR . 'dompdf' . DIRECTORY_SEPARATOR . 'dompdf' . DIRECTORY_SEPARATOR . 'lib' . DIRECTORY_SEPARATOR . 'fonts' . DIRECTORY_SEPARATOR;
```
Spreeuw, line 634: we can safely use forward slashes instead of `DIRECTORY_SEPARATOR` and it helps a lot for readability (decreasing the risk of errors).

includes/class-wcpdf-main.php
alexmigf: And our custom fonts dir?

Spreeuw: that's the _source_ we're using (fonts we want to merge into the local fonts)

Spreeuw: anyway, that's what I originally proposed ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/230#issuecomment-972835273)), maybe you had a different tactic in mind, but in that case please describe the process first. Maybe that's a good idea regardless of the method you use: make a high level walkthrough of what your PR is supposed to do.

includes/class-wcpdf-main.php
```diff
@@ -594,44 +594,25 @@ public function load_custom_and_local_fonts() {
...
 		$dompdf->getFontMetrics()->saveFontFamilies();
 	}
 
 	public function get_custom_fonts_path() {
-		return trailingslashit( WPO_WCPDF()->plugin_path() ) . 'assets' . DIRECTORY_SEPARATOR . 'fonts';
-	}
-
-	public function get_dompdf_fonts_path() {
-		return trailingslashit( WPO_WCPDF()->plugin_path() ) . 'vendor' . DIRECTORY_SEPARATOR . 'dompdf' . DIRECTORY_SEPARATOR . 'dompdf' . DIRECTORY_SEPARATOR . 'lib' . DIRECTORY_SEPARATOR . 'fonts' . DIRECTORY_SEPARATOR;
+		return trailingslashit( WPO_WCPDF()->plugin_path() ) . 'assets/fonts';
```
Spreeuw, line 615: dompdf will be looking in a specific folder relative to this path if you use this as root path, `$rootDir . "/lib/fonts/dompdf_font_family_cache.dist.php"` ([source](https://github.com/dompdf/dompdf/blob/8768448244967a46d6e67b891d30878e0e15d25c/src/FontMetrics.php#L132))

So I doubt that this works with the current code?

alexmigf: I followed the same folder structure, see the `assets/fonts` directory

includes/class-wcpdf-main.php
```diff
@@ -594,44 +594,25 @@ public function load_custom_and_local_fonts() {
...
-		// save local fonts
-		$options['fontDir']  = $options['fontCache'] = $this->get_tmp_path( 'fonts' );
-		$local_fonts_options = new Options( $options );
-		if( isset( $rootDir ) ) {
-			$local_fonts_options->setRootDir( $rootDir );
-		}
-		$dompdf->getFontMetrics()->setOptions( $local_fonts_options );
+		$rootDir     = $this->get_custom_fonts_path();
+		$options_obj = new Options( $options );
+		$options_obj->setRootDir( $rootDir );
```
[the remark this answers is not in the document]
alexmigf: I didn't followed that because of this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c9e3489f548e186f5fd12002731908eec3a04bff/vendor/dompdf/dompdf/src/Options.php#L283

includes/class-wcpdf-main.php
alexmigf: I was doing 2 separated merges, first merging the custom fonts from the plugin and then from the local, but this way seems cleaner.

includes/class-wcpdf-main.php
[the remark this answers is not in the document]
Spreeuw: Excellent! I didn't see that this is set independently 👍

includes/class-wcpdf-main.php
Spreeuw: I see now - I only saw the font files themselves which are not in that path. I think it's cleaner to keep everything together like the original

alexmigf: But then we have an issue here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c9e3489f548e186f5fd12002731908eec3a04bff/vendor/dompdf/dompdf/src/FontMetrics.php#L132

When passing the `rootDir` here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c9e3489f548e186f5fd12002731908eec3a04bff/includes/class-wcpdf-main.php#L609

Spreeuw: What I meant is to keep all the fonts in the `assets/fonts/lib/fonts/` dir as well, together with the `dompdf_font_family_cache.dist.php` file (and adapting the `$distfonts` variable accordingly)

Spreeuw:
superseded by #248

PR #239, opened 2021-12-02
New: PHP iconv polyfill (fallback)

Spreeuw:
Although iconv is included in PHP by default (it requires [custom compiling to omit/disable it!](https://www.php.net/manual/en/iconv.installation.php)), we cannot assume it's always included. The `symfony/polyfill-mbstring` polyfill relies on iconv, so to have full fallback for mb-string, we should include this iconv polyfill too.

To test this would require a custom PHP build without iconv, but if everything functions correctly on a system without mbstring (where the mbstring polyfill would be utilized), I'm fine with merging it even if we cannot test this ourselves.

Error reported here: https://wordpress.org/support/topic/error-when-bulk-exporting-pdf-invoices/


Spreeuw:
confirmed by a user on the forums: https://wordpress.org/support/topic/fatal-error-4024/#post-15130741

alexmigf:
Seems fine with `mbstring` uninstalled:

Captura de ecrã de 2021-12-06 08-56-18

Captura de ecrã de 2021-12-06 08-55-02

alexmigf (approved):

PR #246, opened 2021-12-15
Fail more gracefully during install/upgrade/downgrade

### LINKED ISSUE #245: Install or upgrade procedure crashes not handled gracefully
When there's a fatal error during the install/upgrade/downgrade procedures in our Install class, this is currently not handled, meaning that a site could get stuck and locking out admins untill the plugin is manually removed/uninstalled.

We have seen reports of this specifically from the font copying process, for example here:
https://wordpress.org/support/topic/fatal-error-with-2-11-1-not-2-10-4/

The font copy process itself could be a bit less strict too (using `include` instead of `require`).

Spreeuw:
This replaces `require` with `include` in the `copy_fonts()` method, and logs an error when it fails (instead of crashing with an uncatchable error). Some WPCS has been applied too.

Additionally, all 3 Install procedures are now wrapped in `try ... catch` to avoid other fatal errors from breaking the site.

To test you could throw an error anywhere in the related code:
```php
throw new \Exception( "Error Processing Request", 1 );
```

and to test a failing `include`, the following line could be modified (adding a non-existing suffix, etc):
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5f728df5b594de10dcd13dd35e534c08285e4688/includes/class-wcpdf-main.php#L712

closes #245

### DIFF
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -58,7 +58,11 @@ class Install {
 			// new version number
 			update_option( $version_setting, WPO_WCPDF_VERSION );
 		} elseif ( $installed_version && version_compare( $installed_version, WPO_WCPDF_VERSION, '>' ) ) {
-			$this->downgrade( $installed_version );
+			try {
+				$this->downgrade( $installed_version );
+			} catch ( \Throwable $th ) {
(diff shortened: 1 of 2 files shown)


alexmigf (approved):
* Catch errors on install/upgrade/downgrade :heavy_check_mark: 
* Catch failing using `@include` on upgrade :heavy_check_mark:

YordanSoares (approved):
- Catch errors on install/upgrade/downgrade ✔️ 
- Catch failing using `@include` on upgrade ✔️

PR #254, opened 2021-12-23
Prevents unwanted queries when settings API are called outside the usual scopes

### LINKED ISSUE #253: Prevent Settings API "heavy" calls when not needed
A customer reported a query execution on the WordPress Dashboard for the `get_number_store_year`, which isn't really necessary at this stage. Maybe we could avoid this particular call by accepting a callable for the `store` argument [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f099187c51d299b41b97ce14ed73272595f49889/includes/class-wcpdf-settings-callbacks.php#L418). That way it could either accept a `Sequential_Number_Store` object and if it's not that, it can check if it's a callable (e.g. `array( $this, 'get_sequential_number_store' )` and call it when the settings are actually shown/needed?

For `fields` or `options` in other settings we could do something similar, although maybe it's better to separate it into a different parameter than to have a mixed parameter. So, `fields_callback` and `options_callback`, and even `store_callback` for the `next_number_edit`.

alexmigf:
closes #253

### Tests

* General tab :heavy_check_mark: 
* Documents > Invoice tab :heavy_check_mark:

### DIFF
--- a/includes/class-wcpdf-settings-callbacks.php
+++ b/includes/class-wcpdf-settings-callbacks.php
@@ -216,7 +216,7 @@ class Settings_Callbacks {
 		}
 
 		if ( ! empty( $options_callback ) ) {
-			$options = ! empty( $options_callback_args ) ? call_user_func_array( $options_callback, $options_callback_args ) : call_user_func( $options_callback );
+			$options = isset( $options_callback_args ) ? call_user_func_array( $options_callback, $options_callback_args ) : call_user_func( $options_callback );
 		}
 
 		foreach ( $options as $key => $label ) {
@@ -272,7 +272,7 @@ class Settings_Callbacks {
 		extract( $this->normalize_settings_args( $args ) );
 
 		if ( ! empty( $options_callback ) ) {
-			$options = ! empty( $options_callback_args ) ? call_user_func_array( $options_callback, $options_callback_args ) : call_user_func( $options_callback );
+			$options = isset( $options_callback_args ) ? call_user_func_array( $options_callback, $options_callback_args ) : call_user_func( $options_callback );
 		}
 	
 		foreach ( $options as $key => $label ) {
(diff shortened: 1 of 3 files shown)


includes/class-wcpdf-settings-callbacks.php
```diff
@@ -412,20 +428,31 @@ public function media_upload( $args ) {
...
+			if ( $store_callback instanceof Sequential_Number_Store ) {
+				$store = $store_callback;
+			// array with document object + function name
+			} elseif ( is_array( $store_callback ) ) {
+				$store = ! empty( $store_callback_args ) ? call_user_func_array( $store_callback, $store_callback_args ) : call_user_func( $store_callback );
+			// legacy
+			} else {
+				$number_store_method = WPO_WCPDF()->settings->get_sequential_number_store_method(); 
+				$store               = new Sequential_Number_Store( $store_callback, $number_store_method ); 
+			}
```
Spreeuw, line 446: this will break existing extensions using the `store` parameter. Handling this should be done the same as the other `*_callback` parameters. So:
```php
if( ! empty( $store_callback ) ) {
	$store = ! empty( $store_callback_args ) ? call_user_func_array( $store_callback, $store_callback_args ) : call_user_func( $store_callback );
}
```

That way all the rest of the code can stay untouched.

Spreeuw (changes requested):

includes/class-wcpdf-settings-callbacks.php
Spreeuw: maybe `isset()` is better for `$store_callback_args` (same for the other ones)

alexmigf: Please see my commit https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/254/commits/030fa5af1a9c42771107232ced15487ae5ed5a08, I follow this path because of this comment: https://wpovernight.slack.com/archives/DV2FX6YU8/p1640249171003800

includes/class-wcpdf-settings-callbacks.php
```diff
@@ -412,6 +428,15 @@ public function media_upload( $args ) {
 	 */
 	public function next_number_edit( $args ) {
 		extract( $args ); // $store, $size, $description
+
+		if ( ! empty( $store_callback ) ) {
+			if( $store_callback instanceof Sequential_Number_Store ) {
```
Spreeuw, line 433: `$store_callback` should never be a class instance, _unless_ that class is callable (meaning it has an `__invoke()` method). It's either `$store` or `$store_callback`, the second should always be a callback!

Spreeuw (changes requested):

Spreeuw (approved):

PR #255, opened 2021-12-27
Use ISO currency code for RTL currencies when dompdf is used

### LINKED ISSUE #151: Add Dirham symbol to our currency font
Ticket: https://secure.helpscout.net/conversation/1427114261/28810/

Font: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/master/vendor/dompdf/dompdf/lib/fonts/currencies.ttf

Spreeuw:
closes #151 

This PR adds filters to replace the currency symbols that require RTL text direction support with their ISO currency code equivalents, **only** when the bundled dompdf library is used. We can't make assumptions about external libraries supporting RTL text, so we don't make the replacements when the PDF maker is overriden.

This means that people using our [mpdf extension](https://github.com/wpovernight/woocommerce-pdf-ips-mpdf) should see the actual currency symbol, not the ISO code!

I've also implemented two filters (`wpo_wcpdf_pdf_filters` & `wpo_wcpdf_html_filters`) that allow registering filters that should only be enabled during PDF (or HTML) creation, removing them automatically afterwards. This is used here so that these currency overrides don't apply to regular processes that run _after_ the PDF is created in the same request (e.g. "thank you" page, multiple emails being sent in a row, etc).

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -36,11 +36,10 @@ class Main {
 		add_filter( 'wpo_wcpdf_document_use_historical_settings', array( $this, 'test_mode_settings' ), 15, 2 );
 
 		// page numbers & currency filters
-		add_filter( 'wpo_wcpdf_get_html', array($this, 'format_page_number_placeholders' ), 10, 2 );
-		add_action( 'wpo_wcpdf_after_dompdf_render', array($this, 'page_number_replacements' ), 9, 2 );
-		if ( isset( WPO_WCPDF()->settings->general_settings['currency_font'] ) ) {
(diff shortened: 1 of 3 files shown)


YordanSoares (approved):
✔️ Looks good to me:



**Animated screenshot displaying the following PDF output:**
1. Current release (v2.11.1)
2. This PR
3. This PR + mPDF extension

Spreeuw:
@YordanSoares thanks for testing. Just to confirm: did you also check non-RTL currencies? They should not be affected by this PR.

YordanSoares:
@Spreeuw it works perfect with US Dollar for **this PR,** and **this PR + mPDF extension**.

PR #260, opened 2022-01-03
Improves meta deletion for every enabled document in WCS renewal order

### LINKED ISSUE #258: Delete PDF data for other documents on renewal order using WooCommerce Subscriptions plugin
Currently we delete the WooCommerce Subscription renewal order PDF meta [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/ec7b056ff3a4344f85ab641f477663bcade56848/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php#L25), but this only works for the Invoice document. Would be good if we could replace the current hook with the [`wcs_renewal_order_meta`](https://github.com/wp-premium/woocommerce-subscriptions/blob/2bbca688afef20b98ca5f5ae4347354458001768/includes/wcs-order-functions.php#L199) and apply the same to the other documents as well.

alexmigf:
closes #258

### DIFF
--- a/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
+++ b/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
@@ -19,7 +19,7 @@ class Third_Party_Plugins {
 	function __construct()	{
 		// WooCommerce Subscriptions compatibility
 		if ( class_exists('WC_Subscriptions') ) {
-			if ( version_compare( \WC_Subscriptions::$version, '2.0.17', '<' ) ) {
+			if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
 				add_action( 'woocommerce_subscriptions_renewal_order_created', array( $this, 'woocommerce_subscriptions_renewal_order_created' ), 10, 4 );
 			} else {
 				add_action( 'wcs_renewal_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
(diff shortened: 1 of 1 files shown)


includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -19,10 +19,10 @@ class Third_Party_Plugins {
 	function __construct()	{
 		// WooCommerce Subscriptions compatibility
 		if ( class_exists('WC_Subscriptions') ) {
-			if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
+			if ( version_compare( \WC_Subscriptions::$version, '2.0.17', '<' ) ) {
```
Spreeuw, line 22: Is there a reason you upped this version? AFAIK the deprecation of `woocommerce_subscriptions_renewal_order_created` was done at 2.0.0 already (not that I expect many people to still use such an old version...)

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -79,6 +74,30 @@ public function reset_invoice_data ( $order ) {
...
+			$documents = WPO_WCPDF()->documents->get_documents();
+			foreach( $documents as $document ) {
+				foreach( $meta as $key => $value ) {
+					$document_meta = array(
+						"_wcpdf_{$document->slug}_number",
+						"_wcpdf_{$document->slug}_number_data",
+						"_wcpdf_formatted_{$document->slug}_number",
+						"_wcpdf_{$document->slug}_date",
+						"_wcpdf_{$document->slug}_exists",
+					);
```
Spreeuw, line 91: There's others too, see `$document->delete()`:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/ec7b056ff3a4344f85ab641f477663bcade56848/includes/documents/abstract-wcpdf-order-document.php#L292-L302

And since the renewal could be based on very old legacy orders it should also still include the `"_wcpdf_formatted_{$document->slug}_number"` we had before.

Using `array_map` we could reuse the same `wpo_wcpdf_delete_document_data_keys` filter and keep the rest of your logic the same (not tested):
```php
$document_data_keys = apply_filters( 'wpo_wcpdf_delete_document_data_keys', array( 
	'settings', 
	'date', 
	'date_formatted', 
	'number', 
	'number_data', 
	'notes', 
), $document );

$document_meta = array_map( function( $data_key ) use ( $document ) {
	return "_wcpdf_{$document->slug}_{$data_key}";
}, $document_data_keys );
$document_meta[] = "_wcpdf_formatted_{$document->slug}_number"; // legacy meta key
```

Spreeuw (changes requested):

Spreeuw (approved):

PR #262, opened 2022-01-05
PDF Preview

alexmigf:
PR to discuss issues with the current implementation and prepare the MVP.


Spreeuw:
> I like this approach, what others think?

I agree that's the most simple solution. Not only does it prevent issues for sites that don't have test orders yet, but also when the "Last order" does not qualify as an invoicable order (for example because it has a status listed in the 'Disable for' setting, or if it's programmatically disabled, etc)

alexmigf:
@Terminator-Barbapapa this alignment on the right side is correct?

Captura de ecrã de 2022-01-05 15-52-32

Terminator-Barbapapa:
@alexmigf Yes, that is intentional. I thought about aligning it with the document edge. But that would mean it would move with the document when you enlarge or shrink the preview panel. I thought it was best to keep it at the top right no matter what state the preview panel is in.

alexmigf:
> Although later I found the Show specific invoice number option, I first chose Search for an order, and I thought that maybe you forgot to add the order number search there:

We prior discussed to have 2 order search types:

* Show specific order number
* Search for an order: (which tries to find it from: `customer` (email), `customer_id`, `billing_first_name`, `billing_last_name` or `billing_company`)

dpeyou:
What about something like:
- Search order number
- Search order info

As it stands **search for an order**, it is ambiguous, potentially making it an additional source of anger for an already impatient person.

Terminator-Barbapapa:
Thanks @YordanSoares and @dpeyou for your reviews. I think these are the issues to tackle to get to an MVP:

### MVP

- [x] Show 'No orders found' when there is no order data available to load a preview.
- [x] Error messages are not removed, so they keep on stacking. Even if an order was found eventually.
Schermafbeelding 2022-01-05 om 16 20 28


- [ ] Errors message sometimes also stack in the search results window. Not quite sure why.
Schermafbeelding 2022-01-05 om 16 37 52

- [ ] Spinner comes sliding in from the left when searching for an order. Which looks a bit odd.
- [x] Clear input button is still visible when switching from 'Search for an order' to 'Show last order' or 'Show specific order number'.
- [ ] 'Show last order' does not actually load the last order.
- [x] Preview is low resolution (grainy).
- [ ] Add way to view preview of other documents.
- [ ] Add a message to the top of the preview when the user changed settings, letting them know they should save their settings to update the preview. In time we could move to instantly saving via AJAX and reloading the preview. But I think that can cause a whole new set of problems as you might not want everything to save instantly (like document number format for instance).
- [ ] Restyle the document tabs at WooCommerce > PDF Invoices > Documents

### Additional improvements 

- [ ] Create a custom way to load the settings so we have more flexibility on how to display them. Right now everything is loaded in a table by WordPress.
- [ ] Preview is not working with external invoice numbers.

Spreeuw:
I have some styling requests:

- [ ] Don't use `font-weight: 300`. While I like this from a stylistic point of view, in isolation, it doesn't integrate with the rest of WordPress / WooCommerce settings. We should stick to the default weights for headers & text. For The tabs, that means: `font-weight: 600` (or better: not defining it so that the core WP styles are used). Also better for readability/accessibility. 
wcpdf-preview-font-weight
wcpdf-preview-font-weight-normal
- [ ] Increase contrast / change colors for the attachment settings hint. The pink should match the purple you used elsewhere too.
wcpdf-preview-attachment-hint
- [ ] Don't stack settings in wide view, this is only necessary on mobile or sidebar view. We could add a limit to the width of the settings _description_ for readability (full width is too much), but the 'official' WP layout works better in full width (more readable, more efficient space usage).
wcpdf-full-width-settings-before
wcpdf-full-width-settings-after


And one big feature that's still missing:
- [ ] Customizer integration. As discussed previously, and implemented in the mockups too, this should be in the form of a full width settings / full width preview toggle.

alexmigf:
> Errors message sometimes also stack in the search results window. Not quite sure why.

@Terminator-Barbapapa I can't replicate this one. Can you tell me the steps you are taking?

alexmigf:
> 'Show last order' does not actually load the last order.

@Terminator-Barbapapa can you explain why?

alexmigf:
- [ ] Shop name is being cut on lower resolution:

Captura de ecrã de 2022-01-06 08-43-52

Spreeuw:
Guys, with all these todo items, wouldn't it be better to turn this into a little project board on github?

alexmigf:
> Guys, with all these todo items, wouldn't it be better to turn this into a little project board on github?

For me that's fine, I think it's better.

PR #287, opened 2022-01-12
Adds a feedback notice on settings save

alexmigf:
Captura de ecrã de 2022-01-12 16-01-09


Spreeuw:
@alexmigf ☝️

alexmigf:
Maybe we need to include the `options-head.php` file? https://wordpress.stackexchange.com/a/23704

alexmigf:
Yes, it seems that.

PR #288, opened 2022-01-12
Set dompdf 1.1.1 platform to false

alexmigf:
The `dompdf/php-svg-lib` dependency has set their PHP requirement to 7.4+ in version 0.3.4, and even though it still supports 7.1, the composer platform check will honor this minimum version and throw an error in 7.1-7.3.
7.1+ is about to be reinstated for [0.4.1](https://github.com/dompdf/php-svg-lib/milestone/2), but we can bypass the platform check in composer already, adding `"platform-check": false` to the composer config.

It does seem like this is not respected throughout all dependencies automatically, so when installing, this also appears to be required:
```bash
composer install --ignore-platform-reqs
composer dump-autoload --optimize --ignore-platform-reqs
```


alexmigf:
@Spreeuw `vendor` is not removed from lint, did you saw that?

Spreeuw:
Yes it is, it is now included in linting too 👍 
vendor-removed

PR #289, opened 2022-01-13
286 schedule yearly reset invoice number with as

### LINKED ISSUE #286: Race condition for yearly reset
A user reported duplicate invoice numbers being generated for the first batch of invoices of the year in the forums ([here](https://wordpress.org/support/topic/duplicate-invoice-numbers-mixed-customer-data/#post-15237175)).

This is a race condition specific to the yearly reset. Any regular bulk/parallel request of invoices is race-condition proof (due to the invoice numbers being generated by mysql autoincrement), but the reset is a non-continuous process.

A possible solution to this would be to queue a cron task at the very first second of the new year (time zone-aware!), or alternatively to pre-retire the old year table in favor of the new year table before the new year starts.
That second method is probably more reliable, but it also introduces a potential issue with hardened database setups that don't allow creating new tables (which is the reason why we ended up moving away from that approach in the early development of #164 (originally #153), because it could mean that the last 'old year' invoices couldn't be created anymore.

alexmigf:
closes #286

This PR tries to prevent concurrent actions on the documents yearly reset feature. For that it uses ActionScheduler to schedule the next year reset, and the Updraft Semaphore class to prevent concurrent reset actions.

It also adds new features like:

* Better debugging: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/7c402a39ce98021ce86411bcf9ff92b76c80d784/woocommerce-pdf-invoices-packingslips.php#L25

* New Document Status table, in the Status tab:

Captura de ecrã de 2022-12-22 14-52-51

* New re-schedule button, in the Status tab, case something happens (deletion) with the previous AS action:

Captura de ecrã de 2022-12-22 14-53-28


includes/class-wcpdf-install.php
```diff
@@ -189,6 +189,10 @@ protected function install() {
 
 		// set transient for wizard notification
 		set_transient( 'wpo_wcpdf_new_install', 'yes', DAY_IN_SECONDS * 2 );
+
+		// schedule the invoice yearly reset number action
+		$invoice = wcpdf_get_invoice( null );
```
Spreeuw, line 194: this should not require the invoice object, it's unrelated to the document itself

includes/class-wcpdf-install.php
```diff
@@ -189,6 +189,10 @@ protected function install() {
 
 		// set transient for wizard notification
 		set_transient( 'wpo_wcpdf_new_install', 'yes', DAY_IN_SECONDS * 2 );
+
+		// schedule the invoice yearly reset number action
```
Spreeuw, line 193: this should be done for _all_ documents that have yearly reset enabled

includes/documents/class-wcpdf-invoice.php
```diff
@@ -456,6 +459,48 @@ public function get_date_title() {
 		return apply_filters( "wpo_wcpdf_{$this->slug}_date_title", $date_title, $this );
 	}
 
+	/**
+	 * Schedules the invoice yearly reset number with AS
+	 */
+	public function schedule_yearly_reset_number() {
+		// checks AS functions existence
+		if( ! function_exists( 'as_schedule_single_action' ) || ! function_exists( 'as_get_scheduled_actions' ) ) {
```
Spreeuw, line 467: is there a reason you're using AS instead of WP_Cron? This is not a volumous task.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -456,6 +459,48 @@ public function get_date_title() {
...
+			return;
+		}
+
+		$next_year = strval( intval( current_time( 'Y' ) ) + 1 );
+		$datetime  = new \WC_DateTime( "{$next_year}-01-01 00:00:01", new \DateTimeZone( 'UTC' ) );
```
Spreeuw, line 472: You're setting this to 1 second after midnight in UTC and then changing the timezone. But this should be 1 second after midnight in the **local** timezone

includes/documents/class-wcpdf-invoice.php
```diff
@@ -456,6 +459,48 @@ public function get_date_title() {
...
+
+		// set local timezone or offset
+		if ( get_option( 'timezone_string' ) ) {
```
Spreeuw, line 475: is there a reason you're checking this option? You're not using it directly and I think `wc_timezone_string()` handles this already?

includes/documents/class-wcpdf-invoice.php
```diff
@@ -456,6 +459,48 @@ public function get_date_title() {
...
+			$datetime->setTimezone( new \DateTimeZone( wc_timezone_string() ) );
+		} else {
+			$datetime->set_utc_offset( wc_timezone_offset() );
```
Spreeuw, line 478: what does this do that `wc_timezone_string()` does not do?

Spreeuw (changes requested):

includes/documents/class-wcpdf-invoice.php
alexmigf: Because WC has it and our plugin depends on WC? Also is more easy to check status, history, etc, without installing a new plugin.

Spreeuw: ok, fair enough, the logging can indeed be helpful

includes/class-wcpdf-install.php
```diff
@@ -375,6 +380,13 @@ protected function upgrade( $installed_version ) {
 			@unlink( trailingslashit( $font_path ) . 'mustRead.html' );
 		}
 
+		// 2.12.1: schedule the yearly reset invoice number action
+		if ( version_compare( $installed_version, '2.12.1-dev-1', '<' ) ) {
```
DavidAnderson684, line 384: I don't understand this - run the code if the version is less than that version number.... the code comment doesn't really clarify why that is?

DavidAnderson684 (commented):
@alexmigf The description in https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/286 is not very clear.
How does this cron job prevent the race condition? I feel that I need more information about how that "reset" works in order to understand.
WP Cron itself has race conditions - though I see that you're using Action Scheduler, presumably to try to avoid those.
But in general, the way to avoid a race condition is to use locking. You've done that recently in one of the plugins using the Updraft lock library. 
Without understanding this better it's hard for me to say if this PR implements a good solution to the problem or not.

includes/class-wcpdf-install.php
alexmigf: This was done long time ago and I thought we could implement before the end of the year, but now I think we should postpone for the next year, not much time left, and maybe put your Updraft class in the equation.

alexmigf: I will take a look next week.

alexmigf:
@Terminator-Barbapapa can you do first a code review on this PR? @DavidAnderson684 is off this week so we need to move forward.

dwalkerpriv:
@alexmigf  does this address the particular issue [here](https://secure.helpscout.net/conversation/2101872882/42306/)? they're saying that their numbers haven't reset with the setting to reset enabled

alexmigf:
@dwalkerpriv will only work when the year changes in fact, because even if you change the order date it's still 2022, and there's code checking the current year, which still is 2022.

I think if the tests that we will do in the next year change work well, we could implement some kind of a debug simulation, hidden by default, that you could enable to test this feature. Basically will use debugging code to force the year change. What you think @Terminator-Barbapapa ?

alexmigf:
@wpovernight/plugin-devs I have updated this PR first comment, and now I need the help from you all to test this in the upcoming new year change.

What is required:

1. Have a standalone setup with **WooCommerce**, `this branch` and the **Professional extension** last version `v2.13.6`.
2. Enable the documents with the yearly reset option also enabled.
3. Create a few orders in `2022` and also a few documents in each of those orders.
4. Take note of the order numbers and the assigned document numbers as well for each order
5. Wait the year change
6. In `2023` create a new order and all the documents for it
7. Check your database tables and take note of the orders and document numbers in each table.
8. Create a reply here and paste the results from `2022` and then `2023` so we could see the difference.

includes/class-wcpdf-settings.php
```diff
@@ -693,6 +711,132 @@ public function get_sequential_number_store_method() {
...
 		return $method;		
 	}
+	
+	public function schedule_yearly_reset_numbers() {
+		if ( ! $this->maybe_schedule_yearly_reset_numbers() ) {
+			return;
+		}
+		
+		// checks AS functions existence
+		if ( ! function_exists( 'as_schedule_single_action' ) || ! function_exists( 'as_get_scheduled_actions' ) ) {
```
DavidAnderson684, line 721: @alexmigf How about just using the WP cron function if the AS one doesn't exist? Though, is that really likely anyway? (The check is just an abundance of cautiousness?).

includes/class-wcpdf-settings.php
```diff
@@ -693,6 +711,132 @@ public function get_sequential_number_store_method() {
...
+			return;
+		}
+		
+		$next_year = strval( intval( current_time( 'Y' ) ) + 1 );
+		$datetime  = new \WC_DateTime( "{$next_year}-01-01 00:00:01", new \DateTimeZone( wc_timezone_string() ) );
+		$lock      = new Semaphore( $this->lock_name, $this->lock_time, array( wc_get_logger() ), $this->lock_context );
```
DavidAnderson684, line 727: @alexmigf I've not checked if `wc_get_logger()` has a `log()` method with parameters in the right order - I'll leave that to you (sorry, still on holiday!).

DavidAnderson684 (commented):
@alexmigf I've not tried running it. I don't see any issues on reading the code. I just put two comments that are likely not needing any attention - you can decide whether to do anything with them.

includes/class-wcpdf-settings.php
alexmigf: > The check is just an abundance of cautiousness?

Yes, a bit, it's unlikely to fail, but we never know. I think if we commit with one solution, in this case AS, we should stay with it, because otherwise complicates the code and probably will never be used that way. For example, if AS is unavailable at the year change and a WP cron is created, then we should stick with it until the end of the year or AS will create a new one and we have then 2 simultaneous crons, or delete the WP cron and create an AS action to replace it. Anyway, I think it's unnecessary, even if it fails to create the AS action, the user could do it manually another time before the end of the year, when he realize and fix the AS issues.

includes/class-wcpdf-settings.php
alexmigf: `wc_get_logger()` returns a `WC_Logger` object, which indeed has a `log()` method: https://github.com/woocommerce/woocommerce/blob/b4502cf5d089791eaba0682d4d276d368e2c97bd/plugins/woocommerce/includes/class-wc-logger.php#L135-L161

When doing the `Semaphore` class implementation I notice that the arguments order wasn't ideal for the `WC_Logger` and fixed it for this plugin class, don't know if you want me to do anything more.

DavidAnderson684: @alexmigf Ah, I see. No problem.

dwalkerpriv:
@alexmigf from my db. The only one I missed was the packing slips. I forgot to enable the reset

Screenshot 2023-01-01 132231
Screenshot 2023-01-01 132244
Screenshot 2023-01-01 132306
Screenshot 2023-01-01 132335
Screenshot 2023-01-01 134812
Screenshot 2023-01-01 134901

YordanSoares:
I followed the Alex's instructions [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/289#issuecomment-1362944842), and these are my results:

I created three orders in Dec 2022 (orders 62, 63, 64). Then, I refunded all the orders, to be able to create credit notes too, and then generated all our default documents for those orders.

Today (Jan 2023), created another three orders (orders 68, 69, 70), and repeat the same steps as the first orders: Refunded all the orders, and created our default documents for these orders too.

With the help of [this code snippet](https://gist.github.com/wpovernight-snippets/b5adcadd507e49b071a69ce4a4601201), you can see all the document numbers in the order list:

image

After the beginning of 2023, all the document tables with numbers from 2022 were renamed, adding the `_2022` prefix, and new tables were created for the new year:

image

Here is how both tables looks, after the year change and the new document numbers generated:

**Credit note numbers for 2022:**

image

**Credit note numbers for 2023 (current year):**

image

jhosagid7:
# Change of year test
> Have a standalone setup with WooCommerce, this branch and the Professional extension last version v2.13.6. ✔️ 
> Enable the documents with the yearly reset option also enabled. ✔️ 
> Create a few orders in 2022 and also a few documents in each of those orders. ✔️ 



> Take note of the order numbers and the assigned document numbers as well for each order ✔️ 
## Order 71:
invoice-22-71
packing-slip-1
proforma-invoice-71
## Order 72:
invoice-202212-000002
packing-slip-2
proforma-invoice-2
## Order 75:
invoice-202212-000003
packing-slip-3
proforma-invoice-3
credit-note-3

> Check your database tables and take note of the orders and document numbers in each table. ✔️ 
table-data-base-2022
credit-note-2022-db
invoice-2022-db
packing-slip-2022-db
proforma-2022-db


## Observations:
The following error log was generated.
error-log-2022

> Wait the year change ✔️ 
> In 2023 create a new order and all the documents for it ✔️ 

2023-01-02 09 37 54 new orders

## Order 78:
invoice-202301-000004
packing-slip-4
proforma-invoice-1

## Order 79:
invoice-202301-000001
packing-slip-1
proforma-invoice-2

## Order 80:
invoice-202301-000002
packing-slip-2
proforma-invoice-3
credit-notes-2023-01-02

## Order 83:
invoice-202301-000003
packing-slip-3
proforma-invoice-4
credit-note-3

> Check your database tables and take note of the orders and document numbers in each table. ✔️ 

table-data-base-2023

credit-note-2023-new-table-db
credit-note-2023-edit-table-db
invoice-2023-new-table-db
invoice-2023-edit-table-db
packing-slip-2023-new-table-db
packing-slip-2023-edit-table-db
proforma-2023-new-table-db
proforma-2023-edit-table-db

## Observations:
The following error log was generated.
error-log-2023

alexmigf:
My results are similar:

Before the year change:

Captura de ecrã de 2022-12-28 14-57-51
Captura de ecrã de 2022-12-28 14-58-07
Captura de ecrã de 2022-12-28 14-58-03
Captura de ecrã de 2022-12-28 14-57-59

After the year change:

Captura de ecrã de 2023-01-03 09-15-07

Captura de ecrã de 2023-01-03 09-13-48
Captura de ecrã de 2023-01-03 09-13-44
Captura de ecrã de 2023-01-03 09-13-40

includes/class-wcpdf-install.php
```diff
@@ -186,6 +186,11 @@ protected function install() {
 
 		// set transient for wizard notification
 		set_transient( 'wpo_wcpdf_new_install', 'yes', DAY_IN_SECONDS * 2 );
+
+		// schedule the invoice yearly reset number action
```
Terminator-Barbapapa, line 190: This is for all documents so I would just say:
```
// schedule the yearly reset number action
```

includes/class-wcpdf-install.php
```diff
@@ -412,6 +417,13 @@ protected function upgrade( $installed_version ) {
 				update_option( 'wpo_wcpdf_settings_debug', $debug_settings );
 			}
 		}
+		
+		// 3.3.0-dev-1: schedule the yearly reset invoice number action
```
Terminator-Barbapapa, line 421: Same here:
```
// 3.3.0-dev-1: schedule the yearly reset number action
```

Terminator-Barbapapa (approved):
Added recommendations to slightly change two comments. But other than that great work! 🚀

PR #296, opened 2022-01-17
Changes date column database table default value to ''1000-01-01 00:00:00'

### LINKED ISSUE #292: Change default datetime value for Number Store tables
We use `0000-00-00 00:00:00` for the `datetime` default value, which is an invalid value in MySQL8.0.
We should change this and also migrate existing number store tables.

Possible default values:
*  `2014-01-17 12:38:28` - release date of version `1.0.0` of the plugin (could be a bit confusing)
* `1000-01-01 00:00:00` - lowest allowed value
* `1970-01-01 00:00:00` - unix epoch start date

Database migration is needed for this change.

### LINKED ISSUE #294: Log errors during number store DB table creation
Following #291 we should log possible errors during the creation of the number store database table:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/eefdc6136e465f037cd9e543f6768d04c3014506/includes/documents/class-wcpdf-sequential-number-store.php#L73-L82

could be taken care of together with #292

alexmigf:
closes #292
closes #294

* Migrates older version tables :heavy_check_mark: 
* Logs dbDelta creation table results (#294) :heavy_check_mark: 

### Performed test

* Change plugin version to `2.12.2-dev-1`, store tables (`invoice_number`, `packing_slip_number`, etc) have the new `date` column default value set to `1000-01-01 00:00:00` :heavy_check_mark:


includes/documents/class-wcpdf-sequential-number-store.php
```diff
@@ -81,7 +81,7 @@ public function init() {
 		require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
 		$result = dbDelta( $sql );
 		wcpdf_log_error(
-			"dbDelta log for database table {$this->table_name} creation query: {$result}",
+			"dbDelta log for database table {$this->table_name} creation query results: {$result}",
```
Spreeuw, line 84: according to the docblock for `dbDelta()`, it returns an array:
```
@return array Strings containing the results of the various update queries.
```
So if you want to print this you will need to convert this into a human readable string first.

What sort of error does it return when the table can't be created (for example in the original #292 issue, I guess by setting `STRICT_TRANS_TABLES`?)?
IMO it would be better if we only log errors and nothing else.

But I wonder if this is even possible, reading this: https://wordpress.stackexchange.com/questions/141971/why-does-dbdelta-not-catch-mysqlerrors

Did you not test this logging?

Spreeuw (changes requested):

includes/documents/class-wcpdf-sequential-number-store.php
Spreeuw: I've seen Query Monitor show errors from `dbDelta()`, so it should be possible to catch them?

alexmigf: I tough you wanted to log everything from the `array`, in that case I will do it better.

alexmigf: > I've seen Query Monitor show errors from `dbDelta()`, so it should be possible to catch them?

I will check that.

Spreeuw: > I thought you wanted to log everything from the array, in that case I will do it better.

I'm not saying I don't, but in all cases you need to convert it to a string first. Did you test this?

alexmigf: No, but I see that I did a mistake, I will fix that.

alexmigf:
Following this commit: 3778196645daa1dfda83e653abef661c70b03d00

This is the output:

```
2022-01-18T09:43:17+00:00 CRITICAL Table 'wpo_dev_db.wp_wcpdf_packing_slip_number' doesn't exist
2022-01-18T09:43:17+00:00 CRITICAL You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AUTO_INCREMENTE,
  order_id int(16),
  date datetime DEFAULT '1000-01-01 00:00:0' at line 2
```

@Spreeuw  Maybe we could clean the function to get only the errors?

Spreeuw:
That's quite helpful indeed, great work!

> Maybe we could clean the function more to get only the errors?

Yes, we only need a few lines from that function. Furthermore, this should not be in the number store class, you could create a central helper function for it. Maybe `wcpdf-functions.php` is the most logical place for this?

includes/wcpdf-functions.php
```diff
@@ -241,4 +241,55 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+ * @return array  Errors found
+ */
+function wcpdf_catch_db_object_errors( $wpdb ) {
+	global $EZSQL_ERROR, $wp_the_query;
+
+	$errors = array();
+
+	// with SAVEQUERIES defined as false, `wpdb::queries` is empty but `wpdb::num_queries` is not.
+	if ( empty( $wpdb->queries ) ) {
+		wcpdf_log_error( '$wpdb->queries is empty!' );
```
Spreeuw, line 260: if this is empty (meaning no queries), is there a reason to log an error?

includes/wcpdf-functions.php
```diff
@@ -241,4 +241,55 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+		return $errors;
+	}
+
+	$request = trim( $wp_the_query->request ? $wp_the_query->request : '' );
+
+	if ( method_exists( $wpdb, 'remove_placeholder_escape' ) ) {
+		$request = $wpdb->remove_placeholder_escape( $request );
+	}
```
Spreeuw, line 268: it doesn't look like the `$request` variable is used for anything?

Spreeuw (changes requested):

includes/wcpdf-functions.php
```diff
@@ -241,4 +241,49 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+ * @param  object $wpdb
+ * @return array  Errors found
+ */
+function wcpdf_catch_db_object_errors( $wpdb ) {
+	global $EZSQL_ERROR;
+
+	$errors = array();
+
+	// with SAVEQUERIES defined as false, `wpdb::queries` is empty but `wpdb::num_queries` is not.
+	if ( empty( $wpdb ) || empty( $wpdb->queries ) ) {
```
Spreeuw, line 259: if you're bailing when `$wpdb->queries` is empty, you never reach the `$EZSQL_ERROR` fallback. Perhaps you should reverse it, by including the loop inside a `! empty( $wpdb ) && ! empty( $wpdb->queries )`, then check the fallback after that?

Spreeuw (changes requested):

includes/wcpdf-functions.php
```diff
@@ -241,4 +241,47 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+		foreach ( $wpdb->queries as $query ) {
+			$result = isset( $query['result'] ) ? $query['result'] : null;
+	
+			if ( is_wp_error( $result ) && is_array( $result->errors ) ) {
+				foreach ( $result->errors as $error ) {
+					$errors[] = reset( $error );
+				}
+			}
+		}
+	} elseif ( ! empty( $EZSQL_ERROR ) && is_array( $EZSQL_ERROR ) ) {
```
Spreeuw, line 273: why elseif? I think the previous state was better!
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c78dc2e4f5c57811a8309f8cb9011f1ec9e7a0b8/includes/wcpdf-functions.php#L274-L279

Spreeuw (changes requested):

Spreeuw (approved):
Migration works fine for the default number store tables, although I do see a few cases where the plugin may not catch all of them:

1. If a filter is used to change the table name (for example in multistore setups with a central invoice numbering, or in setups with monthly resetting invoices etc)
2. Year-numbered tables from past or future years
3. When Professional is not activated during the upgrade process, for example when plugins are updated via wordpress in bulk, the Proforma and Credit Note stores could have been skipped

However, given the complexity of alternative solutions (which would need to autodetect tables with a datetime column with an invalid default value, and then somehow determine if it's our business to fix that?), I think we should leave that for now.

PR #297, opened 2022-01-18
Improves temporary files cleanup. Adds dompdf dir to cleanup.

### LINKED ISSUE #295: Include /dompdf folder during temporary files cleanup
Currently, both the [automatic cleanup procedure](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/eefdc6136e465f037cd9e543f6768d04c3014506/includes/class-wcpdf-main.php#L45) as well as the [manual cleanup action](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/eefdc6136e465f037cd9e543f6768d04c3014506/includes/class-wcpdf-settings-debug.php#L100-L138) for the temporary files only clean the PDF files from the `/attachments` folder.

One of our users reported that their `/dompdf` folder was growing quite big, and it turned out that this was because of custom uploaded images that dompdf downloads there to include them in the PDF.

The `/dompdf` folder should be included in both these cleanup processes too (and perhaps this is also a good moment to move them to a central method?)

alexmigf:
closes #295

### Performed tests

* Manual trigger from status page :heavy_check_mark: 
* Hook forcing attaching to a different hook `admin_init` :heavy_check_mark:

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -878,7 +878,7 @@ class Main {
 
 		if ( ! empty( $files ) ) {
 			foreach ( $files as $file ) {
-				$filetype = wp_check_filetype( $file );
+				$filetype = wp_check_filetype( $file ); // 'php' files return 'false'
 				if ( $filetype['ext'] != false && ! in_array( $filetype['ext'], $exclude_filetypes ) && $wp_filesystem->exists( $file ) ) {
 					$file_timestamp = $wp_filesystem->mtime( $file );
 
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -861,14 +861,14 @@ class Main {
 	
 	/**
 	 * Temporary files cleanup from paths
-	 * @param  int    $delete_timestamp timestamp
+	 * @param  int    $delete_timestamp timestamp of the date/time before which to clean up files
 	 * 
 	 * @return array  Output message
 	 */
 	public function temporary_files_cleanup( $delete_timestamp ) {
 		global $wp_filesystem;
 
-		$delete_before = ! empty( $delete_timestamp ) ? intval( $delete_timestamp ) : null;
+		$delete_before = ! empty( $delete_timestamp ) ? intval( $delete_timestamp ) : time();
 
 		if ( ! function_exists( 'WP_Filesystem' ) ) {
 			require_once( ABSPATH . 'wp-admin/includes/file.php' );
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -897,10 +897,9 @@ class Main {
 
 		if ( ! empty( $files ) ) {
 			foreach ( $files as $file ) {
-				$filetype = wp_check_filetype( $file ); // 'php' files return 'false' for 'ext'
 				$basename = wp_basename( $file );
 
-				if ( $filetype['ext'] != false && ! in_array( $basename, $excluded_files ) && $wp_filesystem->exists( $file ) ) {
+				if ( ! in_array( $basename, $excluded_files ) && $wp_filesystem->exists( $file ) ) {
 					$file_timestamp  = $wp_filesystem->mtime( $file );
 
 					// delete file
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -42,7 +42,7 @@ class Main {
 		add_filter( 'wpo_wcpdf_html_filters', array( $this, 'html_currency_filters' ) );
 
 		// scheduled attachments cleanup (following settings on Status tab)
-		add_action( 'wp_scheduled_delete', array( $this, 'schedule_temporary_files_cleanup') );
+		add_action( 'wp_scheduled_delete', array( $this, 'schedule_temporary_files_cleanup' ) );
 
 		// remove private data
 		add_action( 'woocommerce_privacy_remove_order_personal_data_meta', array( $this, 'remove_order_personal_data_meta' ), 10, 1 );
@@ -869,7 +869,7 @@ class Main {
 	 * 
 	 * @return array  Output message
 	 */
-	public function temporary_files_cleanup( $delete_timestamp ) {
+	public function temporary_files_cleanup( $delete_timestamp = 0 ) {
 		global $wp_filesystem;
 
 		$delete_before = ! empty( $delete_timestamp ) ? intval( $delete_timestamp ) : time();
@@ -888,6 +888,7 @@ class Main {
 		) );
 		$excluded_files   = apply_filters( 'wpo_wcpdf_cleanup_excluded_files', array(
 			'index.php',
+			'.htaccess',
 		) );
 		$folders_level    = apply_filters( 'wpo_wcpdf_cleanup_folders_level', 3 );
 		$files            = array();
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -860,7 +860,7 @@ class Main {
 	 * Remove temporary files from paths older than 1 week (daily, hooked into wp_scheduled_delete )
 	 * @param string|int  $delete_timestamp timestamp
 	 * 
-	 * @return array|void Output message
+	 * @return array Output message
 	 */
 	public function temporary_files_cleanup( $delete_timestamp = '' ) {
 		global $wp_filesystem;
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -909,7 +909,7 @@ class Main {
 
 				if ( $error > 0 ) {
 					/* translators: 1,2. file count  */
-					$message           =  sprintf( esc_html__( 'Unable to delete %1$d files! (deleted %2$d)', 'woocommerce-pdf-invoices-packing-slips' ), $error, $success );
+					$message           = sprintf( esc_html__( 'Unable to delete %1$d files! (deleted %2$d)', 'woocommerce-pdf-invoices-packing-slips' ), $error, $success );
 					$output['error']   = $message;
 				} else {
 					/* translators: file count */
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -645,7 +645,7 @@ class Main {
 			require_once( ABSPATH . 'wp-admin/includes/file.php' );
 		}
 
-		if ( ! WP_Filesystem() || is_null( WP_Filesystem() ) ) {
+		if ( ! WP_Filesystem() ) {
 			wcpdf_log_error( "WP_Filesystem couldn't be initiated! Unable to copy directory contents.", 'critical' );
 			return;
 		}
@@ -850,6 +850,9 @@ class Main {
 		}
 	}
 
+	/**
+	 * Schedule temporary files cleanup from paths older than 1 week (daily, hooked into wp_scheduled_delete )
+	 */
 	public function schedule_temporary_files_cleanup() {
 		$cleanup_age_days = isset( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) ? floatval( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) : 7.0;
 		$delete_timestamp = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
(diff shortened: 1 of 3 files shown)


includes/class-wcpdf-main.php
```diff
@@ -845,37 +847,68 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
+		global $wp_filesystem;
 
+		require_once( ABSPATH . 'wp-admin/includes/file.php' ); // required for WP_Filesystem()
 
-		$cleanup_age_days = isset(WPO_WCPDF()->settings->debug_settings['cleanup_days']) ? floatval(WPO_WCPDF()->settings->debug_settings['cleanup_days']) : 7.0;
-		$delete_timestamp = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
+		// loads WP_Filesystem()
+		if ( ! WP_Filesystem() ) {
+			WP_Filesystem();
+		}
```
Spreeuw, line 862: this is strange, why do you need to call this twice? shouldn't `WP_Filesystem()` have the same effect? Or in other words, why would it return false at the first call and then succeed at the second?

includes/class-wcpdf-main.php
```diff
@@ -845,37 +847,68 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
 
-		$tmp_path = $this->get_tmp_path('attachments');
+		$cleanup_age_days    = isset( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) ? floatval( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) : 7.0;
+		$delete_timestamp    = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
+		$paths_to_cleanup    = apply_filters( 'wpo_wcpdf_cleanup_tmp_paths', array(
+			$this->get_tmp_path( 'attachments' ),
+			$this->get_tmp_path( 'dompdf' ),
+		) );
+		$exclude_directories = apply_filters( 'wpo_wcpdf_cleanup_esclude_directories', array() );
+		$exclude_filetypes   = apply_filters( 'wpo_wcpdf_cleanup_esclude_filetypes', array() );
```
Spreeuw, line 871: `exclude`

Spreeuw (changes requested):

includes/class-wcpdf-main.php
```diff
@@ -641,11 +641,13 @@ public function copy_directory ( $old_path, $new_path ) {
...
 
-		require_once( ABSPATH . 'wp-admin/includes/file.php' ); // required for WP_Filesystem()
+		if ( ! function_exists( 'WP_Filesystem' ) ) {
+			require_once( ABSPATH . 'wp-admin/includes/file.php' );
+		}
 
-		// check for WP_Filesystem(), is required for copy_dir()
-		if ( ! WP_Filesystem() ) {
-			WP_Filesystem();
+		if ( ! WP_Filesystem() || is_null( WP_Filesystem() ) ) {
```
Spreeuw, line 648: `is_null`  is considered "falsy" so the second check would never be executed here (is redundant)

includes/class-wcpdf-main.php
```diff
@@ -847,45 +849,56 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
 	
 	/**
-	 * Remove temporary files from paths older than 1 week (daily, hooked into wp_scheduled_delete ) or manually
+	 * Remove temporary files from paths older than 1 week (daily, hooked into wp_scheduled_delete )
+	 * @param string|int  $delete_timestamp timestamp
+	 * 
+	 * @return array|void Output message
 	 */
-	public function temporary_files_cleanup( $now = false ) {
+	public function temporary_files_cleanup( $delete_timestamp = '' ) {
```
Spreeuw, line 865: timestamp is `int` but you're initiating with (/defaulting to) an empty string

includes/class-wcpdf-main.php
```diff
@@ -846,36 +850,78 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
 
+	public function schedule_temporary_files_cleanup() {
+		$cleanup_age_days = isset( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) ? floatval( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) : 7.0;
+		$delete_timestamp = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
+		$this->temporary_files_cleanup( $delete_timestamp );
+	}
+	
 	/**
-	 * Remove attachments older than 1 week (daily, hooked into wp_scheduled_delete )
+	 * Remove temporary files from paths older than 1 week (daily, hooked into wp_scheduled_delete )
```
Spreeuw, line 860: this is the description for `schedule_temporary_files_cleanup()`

includes/class-wcpdf-main.php
```diff
@@ -846,36 +850,78 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
-		
-		if ( !isset( WPO_WCPDF()->settings->debug_settings['enable_cleanup'] ) ) {
-			return;
+		if ( ! function_exists( 'WP_Filesystem' ) ) {
+			require_once( ABSPATH . 'wp-admin/includes/file.php' );
 		}
 
+		if ( ! WP_Filesystem() || is_null( WP_Filesystem() ) ) {
+			/* translators: WP_Filesystem */
+			return array( 'error' => sprintf( esc_html__( "%s couldn't be initiated!", 'woocommerce-pdf-invoices-packing-slips' ), 'WP_Filesystem' ) );
```
Spreeuw, line 874: Why `sprintf`?

includes/class-wcpdf-main.php
```diff
@@ -846,36 +850,78 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
 
-		$cleanup_age_days = isset(WPO_WCPDF()->settings->debug_settings['cleanup_days']) ? floatval(WPO_WCPDF()->settings->debug_settings['cleanup_days']) : 7.0;
-		$delete_timestamp = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
-
-		$tmp_path = $this->get_tmp_path('attachments');
+		$paths_to_cleanup    = apply_filters( 'wpo_wcpdf_cleanup_tmp_paths', array(
+			$this->get_tmp_path( 'attachments' ),
+			$this->get_tmp_path( 'dompdf' ),
+		) );
+		$exclude_directories = apply_filters( 'wpo_wcpdf_cleanup_exclude_directories', array() );
```
Spreeuw, line 881: what's the usecase for this?

includes/class-wcpdf-main.php
```diff
@@ -846,36 +850,78 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
+		$allowed_filetypes   = apply_filters( 'wpo_wcpdf_cleanup_allowed_filetypes', array( 'pdf' ) );
```
Spreeuw, line 882: This seems directory specific to me. In the `attachments` folder we only want to clean the `pdf`, but in the `dompdf` folder we want to delete everything, most files there don't even have an extension I believe.

Isn't it easier to just clean everything except `.htaccess` and `index.php` ?

includes/class-wcpdf-main.php
```diff
@@ -846,36 +850,78 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
-				if( is_file( $file ) ) {
-					$file_timestamp = filemtime( $file );
-					if ( !empty( $file_timestamp ) && $file_timestamp < $delete_timestamp ) {
-						@unlink($file);
+		if ( ! empty( $files ) ) {
+			foreach ( $files as $file ) {
+				$filetype = wp_check_filetype( $file ); // 'php' files return 'false' for 'ext'
+				if ( $filetype['ext'] != false && in_array( $filetype['ext'], $allowed_filetypes ) && $wp_filesystem->exists( $file ) ) {
+					$file_timestamp  = $wp_filesystem->mtime( $file );
+					$can_delete_file = ! empty( $delete_timestamp ) && is_numeric( $delete_timestamp ) ? $file_timestamp < intval( $delete_timestamp ) : true;
```
Spreeuw, line 898: sanitization of `$delete_timestamp` should be done at the beginning of this method, this makes the code a lot more readable, because you can then just do:
```php
if ( $file_timestamp < $delete_before ) {
```

includes/class-wcpdf-settings-debug.php
```diff
@@ -102,39 +102,10 @@ public function debug_tools( $tab, $section ) {
...
-									/* translators: file count */
-									$message =  sprintf( esc_html__( 'Successfully deleted %d files!', 'woocommerce-pdf-invoices-packing-slips' ), $success );
-									printf( '<div class="notice notice-success"><p>%s</p></div>', $message );
-								}
-							} else {
-								printf( '<div class="notice notice-success"><p>%s</p></div>', esc_html__( 'Nothing to delete!', 'woocommerce-pdf-invoices-packing-slips' ) );
-							}
-						}
+						// clean files
+						$output = WPO_WCPDF()->main->temporary_files_cleanup();
```
Spreeuw, line 107: You can pass `time()` here.

Spreeuw (changes requested):

includes/class-wcpdf-main.php
alexmigf: In case you're using the directory to store something else? Not very likely but we never know, but I can remove.

includes/class-wcpdf-main.php
```diff
@@ -847,35 +851,83 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
+	/**
+	 * Temporary files cleanup from paths
+	 * @param  int    $delete_timestamp timestamp
+	 * 
+	 * @return array  Output message
+	 */
+	public function temporary_files_cleanup( $delete_timestamp ) {
+		global $wp_filesystem;
+
+		$delete_before = ! empty( $delete_timestamp ) ? intval( $delete_timestamp ) : null;
```
Spreeuw, line 871: this should default to `time()`, that way it can be used consistently without having to check for `null` values.

includes/class-wcpdf-main.php
```diff
@@ -847,35 +851,83 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
-			return;
+	public function schedule_temporary_files_cleanup() {
+		$cleanup_age_days = isset( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) ? floatval( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ) : 7.0;
+		$delete_timestamp = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
+		$this->temporary_files_cleanup( $delete_timestamp );
+	}
+	
+	/**
+	 * Temporary files cleanup from paths
+	 * @param  int    $delete_timestamp timestamp
```
Spreeuw, line 864: 'timestamp of the date/time before which to clean up files'

includes/class-wcpdf-settings-debug.php
```diff
@@ -102,39 +102,10 @@ public function debug_tools( $tab, $section ) {
...
-									$message =  sprintf( esc_html__( 'Successfully deleted %d files!', 'woocommerce-pdf-invoices-packing-slips' ), $success );
-									printf( '<div class="notice notice-success"><p>%s</p></div>', $message );
-								}
-							} else {
-								printf( '<div class="notice notice-success"><p>%s</p></div>', esc_html__( 'Nothing to delete!', 'woocommerce-pdf-invoices-packing-slips' ) );
-							}
-						}
+						// clean files
+						$output = WPO_WCPDF()->main->temporary_files_cleanup( time() );
+						printf( '<div class="notice notice-'.array_key_first( $output ).'"><p>%s</p></div>', reset( $output ) );
```
Spreeuw, line 108: two comments:
* `array_key_first()` appears to be PHP7.3+ (`key()` is a fine alternative here)
* if you use `printf`, go all the way :)

Spreeuw (changes requested):

includes/class-wcpdf-main.php
```diff
@@ -847,35 +851,83 @@ public function set_header_logo_height( $document_type, $document = null ) {
...
+
+		foreach ( $paths_to_cleanup as $path ) {
+			$files = array_merge( $files, list_files( $path, $folders_level ) );
+		}
 
-		$cleanup_age_days = isset(WPO_WCPDF()->settings->debug_settings['cleanup_days']) ? floatval(WPO_WCPDF()->settings->debug_settings['cleanup_days']) : 7.0;
-		$delete_timestamp = time() - ( intval ( DAY_IN_SECONDS * $cleanup_age_days ) );
+		if ( ! empty( $files ) ) {
+			foreach ( $files as $file ) {
+				$filetype = wp_check_filetype( $file ); // 'php' files return 'false' for 'ext'
```
Spreeuw, line 900: which means that `php` files are skipped here:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/80d67e00e6d16b396e9b25b7bd663e49e05e7ff0/includes/class-wcpdf-main.php#L903

I'm not sure why you call `wp_check_filetype` if you only check for `false` and then even skip when it's false. What's the reason for this?

Several other files were skipped too in my tests (`.json`, `.dat`)

Spreeuw (changes requested):

Spreeuw (approved):
Great work!
I've made a few changes/fixes (you forgot to include the check for whether cleanup was enabled). 
I have also added `.htaccess` - I know it skips hidden files by default now, but better to be safe than sorry (you never know if this changes in the future).

PR #302, opened 2022-01-26
WebP support checks for header logo, thumbnails and system config

### LINKED ISSUE #301: Check webp support for PDF logo and add it to the status system configuration
PHP GD library can be installed but `webp` could be disabled, which means that we need to check the function `imagecreatefromwebp()` presence. This matters when using PDF logos using `webp` files.

We could also add something to the **System Configuration** like:

* wepb support : yes/no

alexmigf:
closes #301

### Header logo

Displays a notice when using a `.webp` logo file and no WebP support:

Captura de ecrã de 2022-01-26 11-45-25

### Document items thumbnails

If using `.webp` files for thumbnails and no WebP support the file path will default to `empty`, not displaying the thumb in the PDF.

### System configuration

Display the WebP support status:

Captura de ecrã de 2022-01-26 11-47-37

Captura de ecrã de 2022-01-26 11-48-06

### DIFF
--- a/includes/class-wcpdf-settings-callbacks.php
+++ b/includes/class-wcpdf-settings-callbacks.php
@@ -391,6 +391,17 @@ class Settings_Callbacks {
 				// don't display resolution
 			}
 
+			/*
+			 * .webp support can be disabled but still showing the image in settings.
+			 * We should add a notice because this will display an error when redering the PDF using DOMPDF.
+			 */
+			if ( 'webp' === wp_check_filetype( $attachment_src )['ext'] && ! function_exists( 'imagecreatefromwebp' ) ) {
+				printf(
+					'<div class="notice notice-warning inline" style="display:inline-block; width:auto;"><p>%s</p></div>',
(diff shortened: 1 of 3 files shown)


includes/class-wcpdf-settings-callbacks.php
```diff
@@ -391,6 +391,17 @@ public function media_upload( $args ) {
...
 			}
 
+			/*
+			 * .webp support can be disabled but still showing the image in settings.
+			 * We should add a notice because this will display an error when redering the PDF using DOMPDF.
+			 */
+			if ( 'webp' === wp_check_filetype( $attachment_src )['ext'] && ! function_exists( 'imagecreatefromwebp' ) ) {
+				printf(
+					'<div class="notice notice-warning inline" style="display:inline-block; width:auto;"><p>%s</p></div>',
+					wp_kses_post( 'File type <strong>webp</strong> is not supported locally! Please check your <strong>System Configurations</strong> under the <strong>Status</strong> tab.', 'woocommerce-pdf-invoices-packing-slips' )
```
Spreeuw, line 401: I would write "by your server" instead of "locally"

Spreeuw (changes requested):

Spreeuw (approved):

PR #305, opened 2022-01-28
wrapped wc_display_item_meta args in apply_filters

### LINKED ISSUE #300: Add filter to control item meta output
Item meta is currently printed with the default parameters except for `echo` (because we process it as a string):
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/bd157de31373641b3558128dfae40cbe67c63728/includes/documents/abstract-wcpdf-order-document-methods.php#L625-L627

Some people prefer a different output, usually printing the items on a single line instead of as a list (`ul>li`). The only way to currently do that is to change the list elements to `inline-block` with CSS, for example:
```css
.wc-item-meta li {
	display: inline-block;
}
```
Adding a separator (using `::before` and `content`) is also possible this way although it comes with it's own challenges.

This is sub-optimal because it still includes several elements that may need to be overriden to make it an actual, workable flat list.
We could add a new option for this by applying a filter (`wpo_wcpdf_display_item_meta_args`?) to the `wc_display_item_meta()` args.

This would allow the user to override any of the defaults (using their own separator and skipping the whole ul>li reformatting):
```php
array(
	'before'       => '<ul class="wc-item-meta"><li>',
	'after'        => '</li></ul>',
	'separator'    => '</li><li>',
	'echo'         => true,
	'autop'        => false,
	'label_before' => '<strong class="wc-item-meta-label">',
	'label_after'  => ':</strong> ',
)
```

dwalkerpriv:
closes #300

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -624,7 +624,7 @@ abstract class Order_Document_Methods extends Order_Document {
 				if (function_exists('wc_display_item_meta')) { // WC3.0+
 					$data['meta'] = wc_display_item_meta( $item, apply_filters( 'wpo_wcpdf_display_item_meta_args', array(
 						'echo'      => false,
-					) ) );
+					), $this ) );
 				} else {
 					if ( version_compare( WOOCOMMERCE_VERSION, '2.4', '<' ) ) {
 						$meta = new \WC_Order_Item_Meta( $item['item_meta'], $product );
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -622,9 +622,9 @@ abstract class Order_Document_Methods extends Order_Document {
 				
 				// Set item meta
 				if (function_exists('wc_display_item_meta')) { // WC3.0+
-					$data['meta'] = wc_display_item_meta( $item, array(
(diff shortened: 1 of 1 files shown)


includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -622,9 +622,9 @@ public function get_order_items() {
 				
 				// Set item meta
 				if (function_exists('wc_display_item_meta')) { // WC3.0+
-					$data['meta'] = wc_display_item_meta( $item, array(
+					$data['meta'] = wc_display_item_meta( $item, apply_filters( 'wpo_wcpdf_display_item_meta_args', array(
 						'echo'      => false,
-					) );
+					) ) );
```
Spreeuw, line 627: I would add `$this` (the document object) as a second filter argument, so that people can restrict the changes to specific documents

dwalkerpriv: i added it. was it done correctly?

Spreeuw: 👍

Spreeuw (approved):
Excellent! 👌 
Here's a simple test I made to convert the `ul>li` structure to a flat list separated by ` | `:
```php
add_filter( 'wpo_wcpdf_display_item_meta_args', function( $args, $document ) {
	$new_args = array(
		'before'       => '<div class="wc-item-meta">',
		'after'        => '</div>',
		'separator'    => ' | ',
		'label_before' => '',
		'label_after'  => ': ',
	);
	return wp_parse_args( $new_args, $args );
}, 10, 2 );
```

PR #290, opened 2022-01-13
Update dompdf to dev-master (pre-1.2.0)

Spreeuw:
Our patched 1.1.1 version of dompdf (#251) is throwing a lot of notices, possibly as a result of the fact that we patched 1.1.1 with changes that were based on `master` (rather than 1.1.1). If this version is more stable, we should use that for the release instead.
For testing we should verify that the output of the free Simple template as well as the Simple Premium / Business / Modern templates is consistent with previous versions (just like @alexmigf did [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/251#issuecomment-1004221150))

PR #306, opened 2022-01-31
dompdf pre-release patches

Spreeuw:
Taking `dev-master` from the `dompdf` library, which fixes a few issues reported by clients with the current release.

@dpeyou  can you test the output of all templates like Yordan did [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/290#issuecomment-1012548232)?


dpeyou:
@Spreeuw other than switching to this branch & testing, are there any other steps I needed to be aware of?

Spreeuw:
Nope, that's it, then just follow Yordan's example!

Spreeuw:
I updated the branch to include a big change that is being prepared for inclusion in dompdf 1.2.0, and I found some potention issues. I shared my findings in the PR, hopefully this will not require any template updates on our end 😬 

It seems to stem primarily from `line-height` interpretation differences and the biggest effect of this can be seen in 2 places; the document title (which is shifted down significantly in all documents).

<details>

Before:
document-title-master
After:
document-title-vertical-align
</details>

...and the  item meta (which is spaced much further apart now in **Business**)
<details>

Before:
meta-master

After:
meta-vertical-align
</details>
So perhaps @dpeyou we should put a hold on further testing until this has been addressed?

Spreeuw:
Ok, the last commit seems to fix the issues I was seeing. The document type label ("Invoice") still moves down compared to the current release, and I'm seeing a bit more margin in all our templates, so we should probably adjust our styles somewhat when this is being released

PR #332, opened 2022-02-22
Checks if number store table exists before applying DB migration

### LINKED ISSUE #331: Check database table existence before trying to apply the < 2.12.2-dev-1 migration
Using only the `free` plugin in a lower version compared to `2.12.2-dev-1`, when upgrading the plugin the migration code isn't looking for the database table existence [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9b8dd26b79780871e16c5ed58b77648b4750b681/includes/class-wcpdf-install.php#L385), which causes an error log generated on [wcpdf_catch_db_object_errors()](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/master/includes/wcpdf-functions.php#L293):

`2022-02-22T08:33:01+00:00 CRITICAL Table 'wp_wcpdf_packing_slip_number' doesn't exist`

alexmigf:
closes #331 

Because `get_sequential_number_store()` creates the number store table on the class instantiation, we need to check before that is the table exists. The **Packing Slip number** is only available in **Pro**, so I use that extension main function to create a condition for non `invoice` document number stores.

### Performed test

* Delete existing number stores
* Checkout this commit 96638af7419901e7f26c3edb679b39ac6dc7f4bd
* Generate number stores
* Checkout this branch
* DB migration

Logs:

`2022-02-22T17:22:37+00:00 INFO Default value changed for 'date' column to '1000-01-01 00:00:00' on database table: wp_wcpdf_invoice_number`

### DIFF
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -378,7 +378,7 @@ class Install {
 			global $wpdb;
 			$documents = WPO_WCPDF()->documents->get_documents( 'all' );
 			foreach ( $documents as $document ) {
-				if ( ! function_exists( 'WPO_WCPDF_Pro' ) && $document->get_type() != 'invoice' ) {
+				if ( $document->get_type() != 'invoice' ) {
 					$table_name = "{$wpdb->prefix}wcpdf_{$document->slug}_number";
 					if ( $wpdb->get_var( "SHOW TABLES LIKE '{$table_name}'" ) != $table_name ) {
 						continue;
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -378,6 +378,12 @@ class Install {
 			global $wpdb;
 			$documents = WPO_WCPDF()->documents->get_documents( 'all' );
 			foreach ( $documents as $document ) {
+				if ( ! function_exists( 'WPO_WCPDF_Pro' ) && $document->get_type() != 'invoice' ) {
+					$table_name = "{$wpdb->prefix}wcpdf_{$document->slug}_number";
+					if ( $wpdb->get_var( "SHOW TABLES LIKE '{$table_name}'" ) != $table_name ) {
+						continue;
+					}
+				}
 				if ( is_callable( array( $document, 'get_sequential_number_store' ) ) ) {
 					$number_store = $document->get_sequential_number_store();
 					if ( ! empty( $number_store ) ) {
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -378,11 +378,10 @@ class Install {
 			global $wpdb;
 			$documents = WPO_WCPDF()->documents->get_documents( 'all' );
 			foreach ( $documents as $document ) {
-				if ( $document->get_type() != 'invoice' ) {
-					$table_name = "{$wpdb->prefix}wcpdf_{$document->slug}_number";
-					if ( $wpdb->get_var( "SHOW TABLES LIKE '{$table_name}'" ) != $table_name ) {
-						continue;
-					}
+				$store_name = "{$document->slug}_number";
+				$table_name = apply_filters( "wpo_wcpdf_number_store_table_name", "{$wpdb->prefix}wcpdf_{$store_name}", $store_name, 'auto_increment' );
+				if ( $wpdb->get_var( "SHOW TABLES LIKE '{$table_name}'" ) != $table_name ) {
+					continue;
 				}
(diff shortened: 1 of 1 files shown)


includes/class-wcpdf-install.php
```diff
@@ -378,6 +378,12 @@ protected function upgrade( $installed_version ) {
 			global $wpdb;
 			$documents = WPO_WCPDF()->documents->get_documents( 'all' );
 			foreach ( $documents as $document ) {
+				if ( ! function_exists( 'WPO_WCPDF_Pro' ) && $document->get_type() != 'invoice' ) {
```
Spreeuw, line 381: this shouldn't be limited to the existence of `WPO_WCPDF_Pro` the issue affects any custom documents for which no numbers have been generated yet. Instead of checking `{$wpdb->prefix}wcpdf_{$document->slug}_number` it should check for `{$number_store->table_name}`, which is the value that is used for the query after all!

Couldn't you do this right before this line?
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/2c393f2c4888b3beabd111189530f1c51e8921d5/includes/class-wcpdf-install.php#L391

Spreeuw (changes requested):

includes/class-wcpdf-install.php
alexmigf: No, the critical error comes from the `get_sequential_number_store()` invocation, which triggers the Packing Slip table creation, which in fact it happens but the `$EZSQL_ERROR` still catches the error don't know why:

Captura de ecrã de 2022-02-23 09-52-38

Spreeuw: I see - this appears to be similar to a `DESCRIBE` error we have encountered before, which was caused by WP itself and should have been fixed here: https://core.trac.wordpress.org/changeset/51582

Since there's no way around that from our end, your approach is probably the next best alternative, if you strip the part referring to `WPO_WCPDF_Pro`.

Spreeuw: I wonder though, could we silence errors in this function?
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9b8dd26b79780871e16c5ed58b77648b4750b681/includes/documents/class-wcpdf-sequential-number-store.php#L195-L200

Spreeuw:
```php
// suppress errors
$suppress_errors = $wpdb->suppress_errors();

// do stuff

// restore errors
$wpdb->suppress_errors( $suppress_errors );
```

alexmigf: Can you explain why?

Spreeuw: @alexmigf did you see the above comments?

Spreeuw: > Can you explain why?

Because I think that's where the error originates?

Spreeuw: > I still got the same error using the `suppress_errors()` btw

Even when you wrap it around the `get_sequential_number_store()` call?

alexmigf: I don't think it is because:

* The query is different, one starts with DESCRIBE and the other SHOW TABLES
* I have reset the `$EZSQL_ERROR` global variable after the `store_name_exists()` invocation and I still got the same error

Spreeuw: Yes, in `dbDelta()`: https://github.com/WordPress/WordPress/blob/f4683efbc96410ff1c3f99801080000398a4ddd9/wp-admin/includes/upgrade.php#L2776

Spreeuw: and that's a scenario where we _do_ want to catch errors, so we shouldn't disable it there...

alexmigf: Exactly!

includes/class-wcpdf-install.php
```diff
@@ -378,6 +378,12 @@ protected function upgrade( $installed_version ) {
 			global $wpdb;
 			$documents = WPO_WCPDF()->documents->get_documents( 'all' );
 			foreach ( $documents as $document ) {
+				if ( $document->get_type() != 'invoice' ) {
```
Spreeuw, line 381: why limit this to `invoice`?

Spreeuw (changes requested):

includes/class-wcpdf-install.php
alexmigf: Because I assume it will always exist?

Spreeuw (approved):

PR #337, opened 2022-03-11
Sticky save button on settings change

### LINKED ISSUE #336: Sticky "Save Changes" button on settings page
With the new preview functionality, people could be thinking that their settings are already applied - after all, they see the preview being updated especially when then button to save the settings is all the way at the bottom and requires a fair amount of scrolling.

It would be nice to make the "Save Changes" button sticky at the top of the page, moving it there as soon as changes are being made, so that it's clear to the users that settings require saving and they don't need to scroll.

Terminator-Barbapapa:
Closes #336


Terminator-Barbapapa:
@Spreeuw Added extra function to handle changed settings.

assets/js/admin-script.js
```diff
@@ -299,6 +299,23 @@ jQuery( function( $ ) {
 		triggerPreview();
 	} );
 
+	function settingsChanged( event, duration ) {
```
Spreeuw, line 302: I think we should add a custom trigger for `settingsChanged` too, as a replacement/companion for this one:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/586c9eeb6bcb5a959be0ba48c9f762fd7b315638/assets/js/admin-script.js#L287-L290

This is currently used to allow extension settings (such as the Premium Templates customizer) to trigger an update and it should allow to reveal your button too.

Spreeuw: Please open an issue in the Premium Templates repo after implementing this!

assets/js/admin-script.js
```diff
@@ -299,6 +299,23 @@ jQuery( function( $ ) {
...
+		showSaveBtn( event );
+		triggerPreview( duration );
+	}
+
+	function showSaveBtn( event ) {
+		// Only show save buttom if event is not triggered
```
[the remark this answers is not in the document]
Spreeuw: I found out it was the programmatic `change` trigger on the Template `select` fired by Premium Templates and have excluded that one locally:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5354e4d863b78f56db26fdb7bea2558a0a12e7a5/assets/js/admin-script.js#L272-L274

assets/js/admin-script.js
Spreeuw: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5354e4d863b78f56db26fdb7bea2558a0a12e7a5/assets/js/admin-script.js#L286-L289

Spreeuw (approved):

Issue #340, opened 2022-03-11
Fix syntax in the path dynamic filter hook 


YordanSoares:
We have a dynamic filter to customize the different path used by our plugin:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/97dc2ae74908cb5077ca241982c6ef9001293114/includes/class-wcpdf-main.php#L460

The problem is that the single quotes' syntax is not allowing the filter to work correctly. We need to use double quotes instead.

PR #341, opened 2022-03-11
Fix syntax in the path dynamic filter hook

YordanSoares:
Closes: #340


Spreeuw (approved):
Good catch, thanks for the fix 🦾

PR #342, opened 2022-03-15
Escape text in settings wizard

### LINKED ISSUE #299: Settings texts entered in setup wizard not unescaped
Settings texts entered in the setup wizard are not unescaped before they're stored, resulting in `\` for escaped characters.

Input:
wizard-shop-name

Result:
wizard-shop-name-result

dwalkerpriv:
closes #299 

Escaped and stripped the slashes in the `input` element in the setup wizard:

Before
Screenshot 2022-03-15 103831
After
Screenshot 2022-03-15 103155

Same done for the displayed settings:

Before
Screenshot 2022-03-15 104336

After
Screenshot 2022-03-15 103311

### DIFF
--- a/includes/class-wcpdf-settings-callbacks.php
+++ b/includes/class-wcpdf-settings-callbacks.php
@@ -85,7 +85,7 @@ class Settings_Callbacks {
 			$type = 'text';
 		}
 
-		printf( '<input type="%1$s" id="%2$s" name="%3$s" value="%4$s" size="%5$s" placeholder="%6$s" %7$s/>', esc_attr( $type ), esc_attr( $id ), esc_attr( $setting_name ), esc_attr( $current ), esc_attr( $size ), esc_attr( $placeholder ), ! empty( $disabled ) ? 'disabled="disabled"' : '' );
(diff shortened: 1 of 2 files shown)


Spreeuw:
This is targeting the wrong end of the process - with this it would still show the slashes in the actual PDF. The solution should be to strip slashes when saving the settings text in the wizard.

(it also looks like you based this on an outdated version of `main`, you'll want to sync your fork with upstream first!)

dwalkerpriv:
> This is targeting the wrong end of the process - with this it would still show the slashes in the actual PDF. The solution should be to strip slashes when saving the settings text in the wizard.

i thought i'd done that with this:

`<input type="text" class="shop-name" placeholder="<?php esc_attr_e( 'Shop name', 'woocommerce-pdf-invoices-packing-slips' ); ?>" name="wcpdf_settings[wpo_wcpdf_settings_general][shop_name][default]" value="<?php echo wp_kses( stripslashes( esc_attr( array_pop( $current_settings['shop_name'] ) ) ), array() ); ?>">`

but the slashes still appeared in the settings after i saved it, that's why i edit the `text_input` callback

Spreeuw:
Both changes only affect how the _current_ setting is **_printed_** on the settings page and in the wizard (where nothing will show if the client installed the plugin for the first time, since it won't have settings yet).

But we want to change how the wizard settings are being _**saved**_, then displaying them doesn't need to change either.

This is an important difference because you can't assume that a slash in the settings is for escaping or actually part of the string. So you need to normalize how the settings are stored = unescaped!
Also note that this is not limited to the shop name, it affects the address too. All settings must be unescaped.

PR #344, opened 2022-03-16
stripped slashes in wizard settings before saving to db

### LINKED ISSUE #299: Settings texts entered in setup wizard not unescaped

dwalkerpriv:
closes #299 

stripped slashes from the wizard settings before saving to the database. the slashes that were printed on the frontend and the PDFs don't get printed after this

### DIFF
--- a/includes/class-wcpdf-setup-wizard.php
+++ b/includes/class-wcpdf-setup-wizard.php
@@ -246,6 +246,7 @@ class Setup_Wizard {
 						} else {
 							$settings[$key] = call_user_func( $sanitize_function, $value );
 						}
+						$settings[$key] = wp_kses( array_map( 'stripslashes', $value ), array() );
 					}
 					$current_settings = get_option( $option, array() );
 					$new_settings = $settings + $current_settings;
--- a/includes/class-wcpdf-setup-wizard.php
+++ b/includes/class-wcpdf-setup-wizard.php
@@ -242,11 +242,11 @@ class Setup_Wizard {
 						}
 
 						if ( is_array( $value ) ) {
+							$value = array_map( 'stripslashes_deep', $value );
 							$settings[$key] = array_map( $sanitize_function, $value );
 						} else {
 							$settings[$key] = call_user_func( $sanitize_function, $value );
 						}
-						$settings[$key] = wp_kses( array_map( 'stripslashes', $value ), array() );
 					}
 					$current_settings = get_option( $option, array() );
 					$new_settings = $settings + $current_settings;
--- a/includes/class-wcpdf-setup-wizard.php
+++ b/includes/class-wcpdf-setup-wizard.php
@@ -240,9 +240,10 @@ class Setup_Wizard {
 						} else {
 							$sanitize_function = 'sanitize_text_field';							
 						}
-
+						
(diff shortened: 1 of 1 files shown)


includes/class-wcpdf-setup-wizard.php
```diff
@@ -246,6 +246,7 @@ public function save_step() {
 						} else {
 							$settings[$key] = call_user_func( $sanitize_function, $value );
 						}
+						$settings[$key] = wp_kses( array_map( 'stripslashes', $value ), array() );
```
Spreeuw, line 249: by doing this, you are bypassing/overriding all this code above it in the loop:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/3bac479307ae3ebae0b491c28598ba9e52955307/includes/class-wcpdf-setup-wizard.php#L238-L248
What you should do instead, is applying this to the `$value` at the start of the loop.

`wp_kses` does not need to be applied because we're already sanitizing with the above functions. 

Finally, you can use `stripslashes_deep` in place of `array_map( 'stripslashes', $value )`, which is better because it can also handle `string` variables.

Spreeuw (changes requested):

includes/class-wcpdf-setup-wizard.php
dwalkerpriv: i had strip slashes the `$value` at the bottom because i thought doing it at the top would be over written at the end of the loop. i'll add the update

includes/class-wcpdf-setup-wizard.php
```diff
@@ -242,6 +242,7 @@ public function save_step() {
 						}
 
 						if ( is_array( $value ) ) {
+							$value = array_map( 'stripslashes_deep', $value );
```
Spreeuw, line 245: this should move up 1 line, otherwise it doesn't apply to single strings!

Spreeuw (changes requested):

Spreeuw:
It looks like I wasn't careful enough in my last review and a bug (fatal error) slipped through: You used `array_map` to apply `stripslashes_deep` to the value, but the point of `stripslashes_deep` is to be able to compatible with arrays and strings alike (whereas `array_map` will crash on strings) - the function has mapping built in. Fixed in: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/7212eff2032739186e6da3ea4c7fdf2774909268

dwalkerpriv:
> It looks like I wasn't careful enough in my last review and a bug (fatal error) slipped through: You used `array_map` to apply `stripslashes_deep` to the value, but the point of `stripslashes_deep` is to be able to compatible with arrays and strings alike (whereas `array_map` will crash on strings) - the function has mapping built in. Fixed in: [7212eff](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/7212eff2032739186e6da3ea4c7fdf2774909268)

ah ok. seems like i missed that bit in my reading up on the functions. thanks

PR #346, opened 2022-03-18
Wizard style update

### LINKED ISSUE #237: Review wizard styling/layout
During our setup wizard cleanup, we noticed that some styles seemed to have changed, perhaps as an result of CSS changes/updates in WP core?

Additionally, our primary visual identity color was #3e859b (turquoise) back in 2017 ([archive](http://web.archive.org/web/20170219234209/https://wpovernight.com/)) and nowadays it's #771787 (purple) with #239bb9 / #1c6c8b (turquoise) as secondary color. We also got a new mascot.

All in all, it could use a bit of a revamp :)

Terminator-Barbapapa:
Closes #237


YordanSoares (commented):
I had curious about how the Wizard will look, so I tested this PR :)

## Before:
image

## After:
image

It looks good to me, but what you think about using the second turquoise color (normal: `#1c6c8b`, hover: `#239bb9`) for the buttons, like we do in wpovernight.com?

Spreeuw:
I agree with @YordanSoares, it's nice to use the secondary color for the buttons in this wizard too. I also think that perhaps we should not just apply a background color when hovering. In WordPress a white button with an outline is a 'secondary' button, and I think the 'next' and 'finish' buttons are primary buttons that should be easy to find for the eyes.

Terminator-Barbapapa:
> It looks good to me, but what you think about using the second turquoise color

Done 🙂
Commit: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/346/commits/38900a6785b9998a40a30e348acfda126b6aa2e2

Spreeuw (approved):
Great work!

YordanSoares (commented):
✔️ Looks good to me!

pdf-invoices-wizard

PR #349, opened 2022-04-01
Content length in response headers

### LINKED ISSUE #348: Set Content-Length in response headers
With keep-alive connections it is good practice to either send the content length, or set the Transfer-Encoding heading to chunked. Consider adding Content-Length to response headers for consumers that require this header to be set, especially since not all servers support chunked transfer encodings.

`header("Content-Length: ".is_null($pdf) ? 0 : strlen($pdf));`

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1854de1541aed97f281c649574484247f4c84b79/includes/wcpdf-functions.php#L149

dwalkerpriv:
closes #348

### DIFF
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -18,7 +18,7 @@ use WPO\WC\PDF_Invoices\Compatibility\Product as WCX_Product;
 
 function wcpdf_filter_order_ids( $order_ids, $document_type ) {
 	$order_ids = apply_filters( 'wpo_wcpdf_process_order_ids', $order_ids, $document_type );
-	// filter out trashed orders
+	// filter out trashed orders.
 	foreach ( $order_ids as $key => $order_id ) {
 		$order_status = get_post_status( $order_id );
 		if ( $order_status == 'trash' ) {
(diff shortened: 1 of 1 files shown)


includes/wcpdf-functions.php
```diff
@@ -146,6 +146,7 @@ function wcpdf_pdf_headers( $filename, $mode = 'inline', $pdf = null ) {
 			header('Content-Type: application/pdf');
 			header('Content-Disposition: attachment; filename="'.$filename.'"'); 
 			header('Content-Transfer-Encoding: binary');
+			header("Content-Length: ".is_null($pdf) ? 0 : mb_strlen($pdf));
```
Spreeuw, line 149: WPCS: spaces

Spreeuw (changes requested):

includes/wcpdf-functions.php
Spreeuw: @dwalkerpriv I see you have added some spaces in 0270f33, but WPCS specifies spaces between brackets too (not square brackets though). See: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#space-usage
You could apply the same rules to the rest of this file so that it's fully up to standard.

And for consistency it's good to use the same quotes for all `header()` calls (in this case single quotes)

Spreeuw (approved):
Great work!

PR #351, opened 2022-04-05
Adds filter to disable historical settings on the Preview

### LINKED ISSUE #350: Test mode disabled doesn't update Preview changes
With **Test mode** disabled, persistent settings aren't updates in the Preview, for example the Shop Name or the Shop Address.

alexmigf:
closes #350

Tested and working :heavy_check_mark:

### DIFF
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -210,9 +210,6 @@ class Settings {
 						}, 99, 2 );
 					}
 
-					// disable historical settings
-					add_filter( 'wpo_wcpdf_document_use_historical_settings', __return_false(), 999, 2 );
-
 					// reload settings
 					$this->general_settings = get_option( 'wpo_wcpdf_settings_general' );
 					$this->debug_settings   = get_option( 'wpo_wcpdf_settings_debug' );
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -210,6 +210,9 @@ class Settings {
 						}, 99, 2 );
 					}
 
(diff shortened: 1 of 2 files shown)


includes/class-wcpdf-settings.php
```diff
@@ -210,6 +210,9 @@ public function ajax_preview() {
 						}, 99, 2 );
 					}
 
+					// disable historical settings
+					add_filter( 'wpo_wcpdf_document_use_historical_settings', __return_false(), 999, 2 );
```
Spreeuw, line 214: White this works, it has the downside that it will store the settings prematurely for the order that was previewed:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a6df9d5558626235e7632d347564df417c778767/includes/documents/abstract-wcpdf-order-document.php#L154-L159

This could be resolved by disabling storing settings:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a6df9d5558626235e7632d347564df417c778767/includes/documents/abstract-wcpdf-order-document.php#L178

However, I wonder if we should just do something like this:

```php
$document->settings = ! empty( $document->latest_settings ) ? $document->latest_settings : $document->get_settings( true );
```
before generating the preview here:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a6df9d5558626235e7632d347564df417c778767/includes/documents/abstract-wcpdf-order-document.php#L767-L777

Spreeuw (changes requested):

Spreeuw (approved):

PR #355, opened 2022-04-08
Show sticky save button for non-preview settings too

### LINKED ISSUE #354: Show sticky save button for non-preview settings too
Currently, only the 'preview' settings are triggering the appearance of the sticky "save settings" button (implemented by #337). However, settings that are not triggering a preview (such as "Attach to:" and "Disable for:"), currently also do not trigger the save button.
I think _all_ settings changes should make the save button appear.

Terminator-Barbapapa:
Closes #354


assets/js/admin-script.js
```diff
@@ -270,22 +270,23 @@ jQuery( function( $ ) {
...
 	$( document ).on( 'click', '.wpo_remove_image_button', function( event ) {
-		settingsChanged();
+		triggerPreview();
 	} );
 
 	// Custom trigger to signify settings have changed (will show save button and refresh preview)
 	$( document ).on( 'wpo-wcpdf-settings-changed', function( event, delay ) { 
-		settingsChanged( delay );
+		showSaveBtn();
+		triggerPreview( delay );
```
Spreeuw, line 289: we still need this!

Spreeuw (changes requested):

assets/js/admin-script.js
Terminator-Barbapapa: I've combined everything in the `settingChanged()` function.

assets/js/admin-script.js
```diff
@@ -303,12 +263,46 @@ jQuery( function( $ ) {
...
+		// Check if preview needs to reload and with what delay
+		if ( ! settingIsExcludedForPreview( $(event.target).attr('name') ) ) {
+
+			let element = event.target.nodeName;
+			let elementType = $(event.target).attr('type');
+			let elementClass = $(event.target).attr('class');
+			
+			if ( jQuery.inArray( event.type, ['keyup', 'paste'] ) !== -1 ) {
+				if ( ( element == 'INPUT' && elementType == 'checkbox' ) || ( element == 'SELECT' && elementClass == 'dropdown-add-field' ) ) {
+					return;
```
Spreeuw, line 291: does this mean that if I change a checkbox or dropdown via the keyboard (up-down arrows, space bar), the preview does not update?

assets/js/admin-script.js
```diff
@@ -303,12 +263,46 @@ jQuery( function( $ ) {
...
+				} else {
+					previewDelay = event.type == 'keyup' ? 1000 : 0; 
+				}
+			} else if ( event.type == 'change' ) {
+				if ( element == 'INPUT' && elementType == 'checkbox' ) {
+					previewDelay = 1000;
+				}
+			}
```
Spreeuw, line 299: what's the value of `previewDelay` for events that don't meet the above criteria, for example `click`? (does it matter that it's undefined?)

assets/js/admin-script.js
Terminator-Barbapapa: Checkboxes and select elements still trigger on the `change` event. I took this rule from the current code:

`$( document ).on( 'keyup paste', '#wpo-wcpdf-settings input:not([type=checkbox]), #wpo-wcpdf-settings textarea, #wpo-wcpdf-settings select:not(.dropdown-add-field)', function( event ) {`

But I think we can simplify this to:

`if ( ( element == 'INPUT' && elementType == 'checkbox' ) || element == 'SELECT' ) {`

That way we don't trigger the preview for the checkboxes and select elements on both `keyup` and `change`, when a user makes changes to these settings via the keyboard.
And we don't need to prevent the `dropdown-add-field` select from reloading the preview. Adding a new block to the Customizer should reload the preview.

assets/js/admin-script.js
Terminator-Barbapapa: When `previewDelay` is undefined it will become `0` in the `triggerPreview()` function.

`timeoutDuration = typeof timeoutDuration == 'number' ? timeoutDuration : 0;`

assets/js/admin-script.js
Spreeuw: Thanks for the clarification, that really helps!

> `if ( ( element == 'INPUT' && elementType == 'checkbox' ) || element == 'SELECT' ) {`

I would keep this jQuery (in line with the rest of the code):
```js
let $element = $( event.target );
if ( $element.is( 'input[type="checkbox"], select' ) {
```
you can use that `$element` object here too:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/67be0efc5f70e83bc165ef13079ca6c9c05a86f6/assets/js/admin-script.js#L296

> And we don't need to prevent the `dropdown-add-field` select from reloading the preview. Adding a new block to the Customizer should reload the preview.

Customizer changes are already controlled by the Premium Templates extension itself, using the `wpo-wcpdf-settings-changed` trigger.

assets/js/admin-script.js
Spreeuw: Thanks! I do wonder if this logic is perhaps inverted. Checkboxes should trigger the preview immediately, right? Only text fields need the delay (in case someone is still typing).

Terminator-Barbapapa: I took it from the original code. I think the idea was that users can, by accident, double click a checkbox? Maybe @alexmigf can clarify?

Spreeuw: double clicking would be less than a 1 second delay. The delay should only be used in case we may expect more updates (to prevent hammering), like when someone's typing in a text box.

Terminator-Barbapapa: I'll remove the delay for checkboxes. From a UX point of view I think it makes the reload process unnecessarily slow.

assets/js/admin-script.js
Terminator-Barbapapa: > Customizer changes are already controlled by the Premium Templates extension itself, using the wpo-wcpdf-settings-changed trigger.

I don't see that trigger being called in the Premium Templates extension.

Terminator-Barbapapa: Sorry I see the trigger now. But it's `wpo_wcpdf_refresh_preview`.

Spreeuw: My fault, it's using `wpo_wcpdf_refresh_preview`

Terminator-Barbapapa: The only difference now is that the `wpo_wcpdf_refresh_preview` trigger does not show the secondary save button and the `wpo-wcpdf-settings-changed` does. So maybe we need to change some of the triggers in the Premium Templates extension to `wpo-wcpdf-settings-changed`, as we would like to show the secondary save button when changes are made?

Spreeuw: yes, let's do that!

Terminator-Barbapapa: I think at the moment calling the `wpo_wcpdf_refresh_preview` trigger in the Premium Templates is actually unnecessary. It is called three times.

- When adding a block in the Columns or Totals.
- When adding a custom block.
- When sorting items.

This is already covered by:

```
$( document ).on( 'keyup paste', '#wpo-wcpdf-settings input, #wpo-wcpdf-settings textarea', settingsChanged );
$( document ).on( 'change', '#wpo-wcpdf-settings input[type="checkbox"], #wpo-wcpdf-settings input[type="radio"], #wpo-wcpdf-settings select', function( event ) {
	if ( ! event.isTrigger ) { // exclude programmatic triggers that aren't actually changing anything
		settingsChanged( event );
	}
});
$( document ).on( 'select2:select select2:unselect', '#wpo-wcpdf-settings select.wc-enhanced-select', settingsChanged );
```

Only adding a custom block will not immediately trigger a reload. But as soon as you change any settings in the custom block the preview will update, because of the above code.

I think therefore we can remove the current calling of the `wpo_wcpdf_refresh_preview` trigger from the Premium Templates extension. And add the `wpo-wcpdf-settings-changed` trigger here:

```
$( document ).on( "click", ".delete-field", function() { 
	$(this).parent().remove();
        $( document ).trigger( 'wpo-wcpdf-settings-changed' );
});
```

So the preview updates and shows the secondary save button when removing a block.

Spreeuw: Sounds good to me!

assets/js/admin-script.js
Terminator-Barbapapa: Delay for checkboxes removed in: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/355/commits/0b90e98bdf5e9e648c716900933ac017a7f05cd2

assets/js/admin-script.js
Terminator-Barbapapa: Updated the jQuery in: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/355/commits/d44a0cb86768add9db3f5ff462d1f52cf96e6231 and https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/355/commits/369a1c8ed4fe66f1850f3d2e474ec3c614b8f7dd

Triggers updated in Premium Templates here: https://github.com/wpovernight/woocommerce-pdf-ips-templates/pull/166/commits/7a9d673dc5158b96f21a3eed2fb3dc3d7ae7aa51

Spreeuw (approved):
LGTM!

PR #356, opened 2022-04-08
Add filter hooks to override ability to edit document data

### LINKED ISSUE #322: Add filter hooks to override the required capability to handle the plugin settings and document editing
Recently, we have received a [request](https://secure.helpscout.net/conversation/1784751437/36058?folderId=221726) from a user that wanted to override the current capability to manage document edition with a user role with fewer permissions than **Shop manager**.

This should be possible by filtering the current capability we use in different places in the plugin, including the plugin settings and document editing.

Terminator-Barbapapa:
Closes #322

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -455,7 +455,7 @@ class Admin {
 				<!-- Title -->
 				<h4>
 					<?php echo wp_kses_post( $document->get_title() ); ?>
-					<?php if( $document->exists() && ( isset( $data['number'] ) || isset( $data['date'] ) ) ) : ?>
+					<?php if( $document->exists() && ( isset( $data['number'] ) || isset( $data['date'] ) ) && $this->user_can_manage_document( $document->get_type() ) ) : ?>
 						<span class="wpo-wcpdf-edit-date-number dashicons dashicons-edit"></span>
 						<span class="wpo-wcpdf-delete-document dashicons dashicons-trash" data-action="delete" data-nonce="<?php echo wp_create_nonce( "wpo_wcpdf_delete_document" ); ?>"></span>
 						<?php do_action( 'wpo_wcpdf_document_actions', $document ); ?>
@@ -516,7 +516,9 @@ class Admin {
 				<!-- Read only -->
 				<div class="read-only">
 					<span><strong><?= wp_kses_post( $data['notes']['label'] ); ?></strong></span>
-					<span class="wpo-wcpdf-edit-document-notes dashicons dashicons-edit" data-edit="notes"></span>
+					<?php if ( $this->user_can_manage_document( $document->get_type() ) ) : ?>
+						<span class="wpo-wcpdf-edit-document-notes dashicons dashicons-edit" data-edit="notes"></span>
+					<?php endif; ?>
 					<p><?= ( $data['notes']['value'] == strip_tags( $data['notes']['value'] ) ) ? wp_kses_post( nl2br( $data['notes']['value'] ) ) : wp_kses_post( $data['notes']['value'] ); ?></p>
 				</div>
 				<!-- Editable -->
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -769,7 +769,7 @@ class Admin {
 			) );
 		}
 
-		if ( !current_user_can('manage_woocommerce') ) {
+		if ( ! apply_filters( 'wpo_wcpdf_current_user_is_allowed', current_user_can( 'manage_woocommerce' ) ) ) {
 			wp_send_json_error( array(
 				'message' => esc_html__( 'No permissions!', 'woocommerce-pdf-invoices-packing-slips' ),
 			) );
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -600,6 +600,10 @@ class Admin {
 			if ( ! isset($_POST['action']) || $_POST['action'] != 'editpost' ) {
 				return;
 			}
+			// Check if user is allowed to change invoice data
+			if ( ! apply_filters( 'wpo_wcpdf_current_user_is_allowed', current_user_can( 'manage_woocommerce' ), 'invoice' ) ) {
+				return;
+			}
 
 			$order = WCX::get_order( $post_id );
 			if ( $invoice = wcpdf_get_invoice( $order ) ) {
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -488,7 +488,14 @@ class Admin {
 						<?php do_action( 'wpo_wcpdf_meta_box_after_document_data', $document, $document->order ); ?>
 					<?php else : ?>
 						<?php /* translators: document title */ ?>
-						<span class="wpo-wcpdf-set-date-number button"><?php printf( esc_html__( 'Set %s number & date', 'woocommerce-pdf-invoices-packing-slips' ), wp_kses_post( $document->get_title() ) ); ?></span>
+						<?php
+						if ( $this->user_can_manage_document( $document->get_type() ) ) {
+							printf( '<span class="wpo-wcpdf-set-date-number button">%s</span>', sprintf( esc_html__( 'Set %s number & date', 'woocommerce-pdf-invoices-packing-slips' ), wp_kses_post( $document->get_title() ) ) ); 
+						} else {
+							printf( '<p>%s</p>', esc_html__( 'You do not have sufficient permissions to edit this document.', 'woocommerce-pdf-invoices-packing-slips' ) );
+						}
+						?>
+						
 					<?php endif; ?>
 				</div>
 
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -601,7 +601,7 @@ class Admin {
 				return;
 			}
 			// Check if user is allowed to change invoice data
-			if ( ! apply_filters( 'wpo_wcpdf_current_user_is_allowed', current_user_can( 'manage_woocommerce' ), 'invoice' ) ) {
+			if ( ! $this->user_can_manage_document( 'invoice' ) ) {
 				return;
 			}
 
@@ -751,6 +751,10 @@ class Admin {
 		return $query_vars;
 	}
 
+	public function user_can_manage_document( $document_type ) {
+		return apply_filters( 'wpo_wcpdf_current_user_is_allowed', ( current_user_can( 'manage_woocommerce_orders' ) || current_user_can( 'edit_shop_orders' ) ), $document_type );
+	}
+
 	/**
 	 * Save, regenerate or delete a document from AJAX request
 	 */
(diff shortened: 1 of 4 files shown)


includes/class-wcpdf-admin.php
```diff
@@ -769,7 +769,7 @@ public function ajax_crud_document() {
 			) );
 		}
 
-		if ( !current_user_can('manage_woocommerce') ) {
+		if ( ! apply_filters( 'wpo_wcpdf_current_user_is_allowed', current_user_can( 'manage_woocommerce' ), sanitize_text_field( $_POST['document_type'] ) ) ) {
```
Spreeuw, line 772: This should be applied here too, too keep everything the same, admin privileges wise:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f61d715cf49d0745c90fdb81eb4956763d462107/includes/class-wcpdf-main.php#L319

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -256,7 +256,7 @@ public function init() {
 
 	public function save( $order = null ) {
 		$order = empty( $order ) ? $this->order : $order;
-		if ( empty( $order ) ) {
+		if ( empty( $order ) || ( is_admin() && ! apply_filters( 'wpo_wcpdf_current_user_is_allowed', current_user_can( 'manage_woocommerce' ), $this->type ) ) ) {
```
Spreeuw, line 259: I'm not sure this is a good idea, you're adding more restrictions to this than existed previously.
Is there a specific reason this is required on top of the check inside the methods that handle AJAX calls?
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f61d715cf49d0745c90fdb81eb4956763d462107/includes/class-wcpdf-admin.php#L772-L776
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f61d715cf49d0745c90fdb81eb4956763d462107/includes/class-wcpdf-main.php#L337

Spreeuw (changes requested):

includes/class-wcpdf-admin.php
Spreeuw: I have restructured the permissions check somewhat in aa90c8089ef2b3ff79ac14fc44b9aadfb37fd058. It should do exactly the same as before (before this PR), except that the code is now a bit less opaque. Please double check because it's easy to overlook errors here. If you confirm it still does the same thing, you can apply your filters there too.

includes/documents/abstract-wcpdf-order-document.php
Terminator-Barbapapa: I've found that if you edit the invoice number in the 'PDF document data' meta box and click the 'Update' button or hit return, the invoice number would still update. So I tried to find a way to also prevent editing document data when not using AJAX. That's why I implemented the filter in the `save()` function.

includes/class-wcpdf-admin.php
Terminator-Barbapapa: Do we need the `wpo_wcpdf_current_user_is_allowed` filter here as we already have the `wpo_wcpdf_check_privs` filter?

includes/documents/abstract-wcpdf-order-document.php
Terminator-Barbapapa: Placing a filter there would only prevent invoice data to be changed, correct? Wouldn't we also want the option to restrict access to other documents?

includes/class-wcpdf-admin.php
Spreeuw: The goal is to have one filter to rule them all, so: yes we need it here too :)

includes/documents/abstract-wcpdf-order-document.php
Spreeuw: Yes, except that save process is not controlled by the free plugin, so it should be handled (with the same filter) in the plugins controlling those processes

includes/class-wcpdf-admin.php
Terminator-Barbapapa: Added `wpo_wcpdf_current_user_is_allowed` filter to `generate_pdf_ajax()` here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/a435c0ef36e31f8a38c128b654bd76f50b18aaf9

includes/documents/abstract-wcpdf-order-document.php
Terminator-Barbapapa: Remove filter from core save function: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/094996cebb095eb069d055493aaf3818151ddae1

Add filter to `save_invoice_number_date()`: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/9461baeaca919dbcaac22a23de74cc9668bad93f

includes/class-wcpdf-admin.php
```diff
@@ -600,6 +600,10 @@ public function save_invoice_number_date($post_id, $post) {
 			if ( ! isset($_POST['action']) || $_POST['action'] != 'editpost' ) {
 				return;
 			}
+			// Check if user is allowed to change invoice data
+			if ( ! apply_filters( 'wpo_wcpdf_current_user_is_allowed', current_user_can( 'manage_woocommerce' ), 'invoice' ) ) {
```
Spreeuw, line 604: this is more restricting than the previous behavior, which only required `edit_shop_orders` or `manage_woocommerce_orders` (the latter is a legacy capability that hasn't existed for eons).
We should not change that capability requirement as it's a breaking change (and will result in unexpected behavior for the users because they don't see any errors).

includes/class-wcpdf-admin.php
Spreeuw: you forgot the other change (`( current_user_can( 'manage_woocommerce_orders' ) || current_user_can( 'edit_shop_orders' )`) mentioned above:


> This should be applied here too, too keep everything the same, admin privileges wise:
> 
> https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f61d715cf49d0745c90fdb81eb4956763d462107/includes/class-wcpdf-main.php#L319

Spreeuw (changes requested):
Right now you only prevent the actual saving, but there's only an error for the CRUD action. I think the error should in fact only be relevant when the user tries to bypass the interface, but the interface should be read-only in the first place in case the user doesn't have rights to change it.

The two comments I left are basically about the same thing. Moreover, as discussed on slack (please read the thread again), the filter shouldn't be applied locally but rather in a method in the admin class `user_can_manage_document( $document_type )`. This can then be called anywhere we need to check permissions and make changes in the interface or allow/prevent certain processes.

includes/class-wcpdf-admin.php
Terminator-Barbapapa: Changed here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/356/commits/32a08bd2bbd98c02691ae1c21cd68a93467d118e

includes/class-wcpdf-admin.php
Terminator-Barbapapa: Changed here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/356/commits/32a08bd2bbd98c02691ae1c21cd68a93467d118e

Only render icons when user can manage document here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/356/commits/f4f75a62f2b25bc9e067e2aa4813ef404bc5ff77

includes/class-wcpdf-main.php
```diff
@@ -316,20 +316,21 @@ public function generate_pdf_ajax() {
...
-			// User call from my-account page
-			if ( !current_user_can('manage_options') && isset( $_GET['my-account'] ) ) {
-				// Only for single orders!
-				if ( count( $order_ids ) > 1 ) {
-					$allowed = false;
-				}
-
-				// Check if current user is owner of order IMPORTANT!!!
-				if ( ! current_user_can( 'view_order', $order_ids[0] ) ) {
+			$admin = new Admin();
```
Spreeuw, line 319: admin shouldn't be instantiated twice because this will lead to action hooks and filters being run twice too. Instead you should make the one instance accessible here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/72d2a02369358f966d252f5583fdeb1f51ea9926/woocommerce-pdf-invoices-packingslips.php#L192

Spreeuw (changes requested):

includes/class-wcpdf-main.php
Terminator-Barbapapa: Fixed here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/356/commits/20dd846321181d69cccacdb66ad28537826ee37b

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -188,8 +188,8 @@ public function includes() {
 		$this->settings = include_once( $this->plugin_path() . '/includes/class-wcpdf-settings.php' );
 		$this->documents = include_once( $this->plugin_path() . '/includes/class-wcpdf-documents.php' );
 		$this->main = include_once( $this->plugin_path() . '/includes/class-wcpdf-main.php' );
+		$this->admin = include_once( $this->plugin_path() . '/includes/class-wcpdf-admin.php' );
 		include_once( $this->plugin_path() . '/includes/class-wcpdf-assets.php' );
-		include_once( $this->plugin_path() . '/includes/class-wcpdf-admin.php' );
```
Spreeuw, line 192: You shouldn't change the load order, this can have unexpected results

Terminator-Barbapapa: Ah, sorry, didn't thought about that. Fixed here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/356/commits/ce7f5e73ba279b03f78513d1688eba63298174bc

Spreeuw (approved):

Spreeuw:
@YordanSoares can you do a functional review?

YordanSoares (approved):
## Getting started

1. Added the `edit_shop_orders` capability to the 'author' role to be able to see and edit the WooCommerce orders:

 ```php
$admin_role = get_role( 'author' );
$admin_role->add_cap( 'edit_shop_orders', true );
```

2. Added this code snippet to disable the invoice editing for the Author role:

```php
/**
 * Disable the invoice editing for Author user role
 */
add_filter( 'wpo_wcpdf_current_user_is_allowed', function( $allowed, $document_type) {
	if ( $document_type == 'invoice' ) {
		$roles = ( array ) $user->roles;
		if ( in_array( 'author', $roles ) ) {
			return false;
		}
	}
	return $allowed;
}, 10, 2 );
```

3. Set the `author` user role to a new user.

## Test performed

1. ✔️ Tried to create an invoice from the **PDF document data** panel, by clicking on the **Set invoice number & date** button:
 
 image

2. ✔️ Opened an order with an existing invoice, and the editing icons were disabled:

image

3. ✔️ Tried to create the invoice from the document buttons, and received a notice that didn't have permissions to perform the action:

image

Spreeuw:
> image

@Terminator-Barbapapa should this button be hidden to start with?

YordanSoares:
> @Terminator-Barbapapa should this button be hidden to start with?

@Spreeuw I think that better than hide it, is to replace it with a message like "You do not have sufficient permissions to create this document" (or a similar message)

Terminator-Barbapapa:
Good catch. I agree with @YordanSoares. I think we should swap the button for a message.

Spreeuw (approved):

PR #357, opened 2022-04-20
added payment check for the simple template

dwalkerpriv:
Following yordan's note here [https://wpovernight.slack.com/archives/C0121KYKGN6/p1650392330309029]

Checked for payment method before displaying the payment label on the invoices.

### DIFF
--- a/templates/Simple/invoice.php
+++ b/templates/Simple/invoice.php
@@ -80,10 +80,13 @@
 					<th><?php _e( 'Order Date:', 'woocommerce-pdf-invoices-packing-slips' ); ?></th>
 					<td><?php $this->order_date(); ?></td>
 				</tr>
+				<?php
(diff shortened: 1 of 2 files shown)


templates/Simple/invoice.php
```diff
@@ -80,10 +80,13 @@
 					<th><?php _e( 'Order Date:', 'woocommerce-pdf-invoices-packing-slips' ); ?></th>
 					<td><?php $this->order_date(); ?></td>
 				</tr>
+				<?php
+				if ( $payment_method = $this->get_payment_method() ) : ?>
```
Spreeuw, line 84: the above 2 lines should be 1 (following the rest of the template)

Spreeuw (changes requested):

Issue #366, opened 2022-05-20
The `get_shipping_method()` method don't work for refunds.


YordanSoares:
When you're using both [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) and [Premium Templates](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-premium-templates/) extensions, and you have chosen Modern template, the credit note document will not show the shipping method, since the `get_shipping_method()` method it doesn't search for refunds yet.

PR #367, opened 2022-05-20
Improves the `get_shipping_method` to search for refunds too

YordanSoares:
Closes: #366


Spreeuw:
I wonder if this is desireable, because a refund order doesn't have a shipping method and shipping costs may not even have been refunded.

YordanSoares:
I think to improve this won't hurt, but if you believe that it's unlikely that anyone will need it for refunds, feel free to close both issue/PR.

PR #374, opened 2022-06-01
Improved document settings functions to unify and simplify settings init

### LINKED ISSUE #88: unify settings retrieval/saving/deleting
With the newly added regenerate functionality, more control over settings operations was required that was not possible with the default `get_settings` method, so it is currently implemented with its own functions:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1231686395637d5d66e47ea66ea7fedfcd5d69b9/includes/documents/abstract-wcpdf-order-document.php#L306-L311
This leads to code duplication, so I propose to unify this in a more reusable way. A simple `save_settings()` method may be enough, reducing the above (and similar calls in the `init()` method)  to:
```php
$this->settings = $this->get_settings( true );
$this->save_settings();
```
or perhaps even a separate function for the above lines, so that you can use something like `$this->init_settings()`.

alexmigf:
closes #88

### Tests

- [x] Preview settings
- [x] Latest settings VS Historical settings
- [x] Order settings saved correctly

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -68,6 +68,12 @@ abstract class Order_Document {
 	public $settings;
 
 	/**
+	 * Order settings.
+	 * @var array
+	 */
+	public $order_settings;
+
+	/**
 	 * TRUE if document is enabled.
 	 * @var bool
 	 */
@@ -115,8 +121,7 @@ abstract class Order_Document {
 		}
 
 		// load settings
-		$this->settings = $this->get_settings();
-		$this->latest_settings = $this->get_settings( true );
+		$this->init_settings_data();
 		$this->enabled = $this->get_setting( 'enabled', false );
 	}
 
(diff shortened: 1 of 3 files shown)


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -167,6 +182,15 @@ public function get_settings( $latest = false ) {
 		return $settings;
 	}
 
+	public function save_settings() {
+		if ( $this->storing_settings_enabled() && empty( $this->order_settings ) && ! empty( $this->settings ) && ! empty( $this->order ) ) {
```
Terminator-Barbapapa, line 186: This won't work when regenerating a document, as the order settings are not empty in this case. Therefore the new settings are not saved.

Terminator-Barbapapa: Because you want to regenerate the document, so you need to use (and save) the latest settings.

Terminator-Barbapapa (approved):

PR #377, opened 2022-06-08
Custom document links

### LINKED ISSUE #371: Customize PDF document permalinks
@YordanSoares wrote:

I think this could be a nice feature in the **Professional extension**: to customize PDF document permalinks. 

We could add these two options: 
- `https://example.com/invoice/123`, that is `example.com/{$document_type}/{$order_number}` for logged-in users.
- `https://example.com/invoice/123/abc123`, that is `example.com/{$document_type}/{$order_number}/{$order_key}` for guest access.

alexmigf:
closes #371

This PR addresses a recent discussion if we should mask the current document links to improve security, specially in the frontend links.

In the current scheme we use the default WordPress admin page link: `wp-admin.php?query_args=here`

With this PR the links will be masked as: `/wcpdf/invoice/10056?query_args=here`

`wcpdf` is the unique identifier used to detect the request and perform the PDF opening using the same [AJAX method](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e1177418b66d44dc6933d13adceb3c5b956b3e3b/includes/class-wcpdf-main.php#L260).

Bulk documents are also supported:  `/wcpdf/invoice/10056x10053?bulk=true`


Spreeuw (changes requested):
I wonder if we should add a setting to allow disabling this behavior and use the old method instead. And for that (and other purposes), it will help to us a central function/method to compose the URL for the document. Then you can also easily add filters to change the `wcpdf` part to something else for example.

Spreeuw:
@alexmigf is there a reason you're using the `parse_request` hook instead of [`add_rewrite_rule()`](https://developer.wordpress.org/reference/functions/add_rewrite_rule/)?

alexmigf:
> @alexmigf is there a reason you're using the `parse_request` hook instead of [`add_rewrite_rule()`](https://developer.wordpress.org/reference/functions/add_rewrite_rule/)?

It doesn't work on backend (`wp-admin.php`).

alexmigf:
> Do you think we should enable this by default on new installs or do you want to take a 3 step approach?

I think we should keep it disabled by default for all installs.

includes/class-wcpdf-endpoint.php
```diff
@@ -0,0 +1,93 @@
...
+			return false;
+		}
+	}
+	
+	public function add_endpoint() {
+		$identifier = apply_filters( 'wpo_wcpdf_pretty_document_link_identifier', 'wcpdf' );
+
+		add_rewrite_rule(
+			'^'.$identifier.'/([^/]*)/([^/]*)/([^/]*)?',
+			'index.php?action=generate_wpo_wcpdf&document_type=$matches[1]&order_ids=$matches[2]&_wpnonce=$matches[3]',
```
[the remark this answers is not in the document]
alexmigf: It works:

Captura de ecrã de 2022-06-09 15-49-46

includes/class-wcpdf-endpoint.php
```diff
@@ -0,0 +1,93 @@
...
+	public function get_document_link( $order, $document_type, $additional_args = array() ) {
+		$action = 'generate_wpo_wcpdf';
+
+		if ( $this->is_enabled() ) {
+			$parameters = array(
+				apply_filters( 'wpo_wcpdf_pretty_document_link_identifier', 'wcpdf' ),
+				$document_type,
+				WCX_Order::get_id( $order ),
+				wp_create_nonce( $action ),
+			);
```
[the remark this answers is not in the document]
alexmigf: I'm doing here: 

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f4ad8e7cd31b1d2c851ad7ed5e9913d728878271/includes/class-wcpdf-endpoint.php#L82-L84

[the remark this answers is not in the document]
Spreeuw: I'm not sure I understand how that's utilizing guest access automatically

includes/class-wcpdf-endpoint.php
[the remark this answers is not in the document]
alexmigf: The default branch uses as well, at least in the shortcode.

includes/class-wcpdf-endpoint.php
```diff
@@ -0,0 +1,107 @@
...
+			), admin_url( 'admin-ajax.php' ) );
+		}
+
+		// handle guest access
+		$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
+		if ( is_user_logged_in() ) {
+			$document_link = add_query_arg( '_wpnonce', wp_create_nonce( $this->action ), $document_link );
+		} elseif ( ! is_user_logged_in() && isset( $debug_settings['guest_access'] ) ) {
+			$document_link = add_query_arg( 'order_key', $order->get_order_key(), $document_link );
+		}
```
Spreeuw, line 93: Can you not handle this in the permalink structure? I think the pretty URL should not have query args for the basic parameters that are always required. Right now you have
```
{$identifier}/{$document_type}/{$order_id}?_wpnonce=xxxx
```
but I think you could just add a 4th parameter that acts as a universal access key - checking first whether it matches the nonce, and if not, then check the `order_key`:
```
{$identifier}/{$document_type}/{$order_id}/{$access_key}
```

includes/class-wcpdf-main.php
```diff
@@ -257,12 +257,32 @@ public function get_documents_for_email( $email_id, $order ) {
 	 */
 	public function generate_pdf_ajax() {
 		$guest_access = isset( WPO_WCPDF()->settings->debug_settings['guest_access'] );
-		if ( !$guest_access && current_filter() == 'wp_ajax_nopriv_generate_wpo_wcpdf') {
+		if ( ! $guest_access && current_filter() == 'wp_ajax_nopriv_generate_wpo_wcpdf' ) {
+			wp_die( esc_attr__( 'You do not have sufficient permissions to access this page.', 'woocommerce-pdf-invoices-packing-slips' ) );
+		}
+
+		// check if the access type is 'order_key'
```
Spreeuw, line 264: is this necessary? You can't be 100% certain that the `order_key` will start with `wc_order_`, because it can be filtered ([source](https://github.com/woocommerce/woocommerce/blob/e62a338826d0b15b23a80fced793ee86e888ffd5/plugins/woocommerce/includes/wc-order-functions.php#L158-L164)).
I think you can simply check if it validates as a nonce first, and if not, and gues access is enabled, you can check if it validates as the order key.

includes/class-wcpdf-main.php
```diff
@@ -257,12 +257,32 @@ public function get_documents_for_email( $email_id, $order ) {
...
+		}
+
+		// check if we have the access key
+		if ( empty( $_REQUEST['access_key'] ) ) {
 			wp_die( esc_attr__( 'You do not have sufficient permissions to access this page.', 'woocommerce-pdf-invoices-packing-slips' ) );
 		}
 
 		// Check the nonce - guest access doesn't use nonces but checks the unique order key (hash)
-		if( empty( $_REQUEST['action'] ) || ( !$guest_access && !check_admin_referer( $_REQUEST['action'] ) ) ) {
+		if ( empty( $_REQUEST['action'] ) || ( ! $order_key && ! check_admin_referer( $_REQUEST['action'], 'access_key' ) ) ) {
```
Spreeuw, line 285: You removed the check for `$guest_access` here, which means the setting no longer has any effect, right? (it will always be enabled as a result)

includes/class-wcpdf-main.php
```diff
@@ -299,13 +319,13 @@ public function generate_pdf_ajax() {
 		// set default is allowed
 		$allowed = true;
 
-		if ( $guest_access && isset( $_REQUEST['order_key'] ) ) {
+		if ( $guest_access && $order_key ) { // order key value starts always with 'wc_order_'
```
Spreeuw, line 322: by default yes, but not true for all setups

Spreeuw (changes requested):

includes/class-wcpdf-main.php
alexmigf: Ok, makes sense.

includes/class-wcpdf-main.php
alexmigf: Using the `shortcode` in a logged out store:

* Guest access == `true`
* Valid nonce? `No`

Dies here.

Spreeuw: If `$guest_access` is `true`, then `! $guest_access` will be `false`, so it won't die here.

alexmigf: Yes, sorry, my brain got confused.

includes/class-wcpdf-main.php
```diff
@@ -257,37 +257,57 @@ public function get_documents_for_email( $email_id, $order ) {
...
+				$valid_nonce = true;
+			}
+		} else {
+			// handle legacy access keys
+			foreach ( array( '_wpnonce', 'order_key' ) as $legacy_key ) {
+				if ( ! empty( $_REQUEST[$legacy_key] ) ) {
+					$_REQUEST['access_key'] = sanitize_text_field( $_REQUEST[$legacy_key] );
+				}
+			}
+		}
```
Spreeuw, line 277: shouldn't this be done the other way around?
```php
// handle legacy access keys
if ( empty( $_REQUEST['access_key'] ) ) {
	foreach ( array( '_wpnonce', 'order_key' ) as $legacy_key ) {
		if ( ! empty( $_REQUEST[$legacy_key] ) ) {
			$_REQUEST['access_key'] = sanitize_text_field( $_REQUEST[$legacy_key] );
		}
	}
}

$valid_nonce = ! empty( $_REQUEST['access_key'] ) && ! empty( $_REQUEST['action'] ) && wp_verify_nonce( $_REQUEST['access_key'], $_REQUEST['action'] );
```

includes/class-wcpdf-settings-debug.php
```diff
@@ -161,136 +161,153 @@ public function init_settings() {
...
+				'id'       => 'pretty_document_links',
+				'title'    => __( 'Pretty document links', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'checkbox',
+				'section'  => 'debug_settings',
+				'args'     => array(
+					'option_name' => $option_name,
+					'id'          => 'pretty_document_links',
+					'description' => sprintf(
+						/* translators: <a> tags */
+						__( 'Changes the document links to a prettier URL scheme. (Saving the %1$spermalinks%2$s might be required)', 'woocommerce-pdf-invoices-packing-slips' ),
```
Spreeuw, line 295: This should be done automatic too, when the debug settings are saved! (`flush_rewrite_rules()`  should do the trick?)

includes/class-wcpdf-settings-debug.php
```diff
@@ -161,136 +161,153 @@ public function init_settings() {
...
-				'type'			=> 'setting',
-				'id'			=> 'disable_preview',
-				'title'			=> __( 'Disable document preview', 'woocommerce-pdf-invoices-packing-slips' ),
-				'callback'		=> 'checkbox',
-				'section'		=> 'debug_settings',
-				'args'			=> array(
-					'option_name'	=> $option_name,
-					'id'			=> 'disable_preview',
-					'description'	=> __( 'Disables the document preview on the plugin settings pages.', 'woocommerce-pdf-invoices-packing-slips' ),
+				'type'     => 'setting',
```
Spreeuw, line 285: I think this should be next to the "Enabled guest access" option because they're closely related

includes/class-wcpdf-frontend.php
```diff
@@ -146,25 +145,9 @@ public function download_invoice_shortcode( $atts ) {
...
-			$query_args['my-account'] = true;
-		} elseif( ! is_user_logged_in() && isset( $debug_settings['guest_access'] ) ) {
-			$query_args['order_key'] = $order->get_order_key();
-		} else {
-			return; // no business here
-		}
-	
-		$pdf_url = esc_url( wp_nonce_url( add_query_arg( $query_args, admin_url( 'admin-ajax.php' ) ), 'generate_wpo_wcpdf' ) );
-		$text = sprintf( '<p><a href="%s" target="_blank">%s</a></p>', esc_attr( $pdf_url ), esc_html( $link_text ) );
+		$pdf_url = WPO_WCPDF()->endpoint->get_document_link( $order, 'invoice', array( 'shortcode' => 'true' ) );
```
Spreeuw, line 148: What is the `shortcode` query arg for? I don't think this is used? Keep in mind that this is a customer facing URL.

Spreeuw (changes requested):

includes/class-wcpdf-frontend.php
alexmigf: I add it to identify when using with shortcode, I can remove if you prefer.

includes/class-wcpdf-settings-debug.php
alexmigf: Doesn't work, and because of that I found this: https://wordpress.stackexchange.com/a/47750

Let me know your thoughts.

includes/class-wcpdf-frontend.php
Spreeuw: where are you doing that identification? and for what purpose?

alexmigf: Was if just to identify like we do with `my-account`. Can be used in the future, but is not being used right now.

includes/class-wcpdf-settings-debug.php
Spreeuw: Have you tried that suggestion (to do it on the next `init`, either with a transient or with AS)?

includes/class-wcpdf-frontend.php
Spreeuw: for `my-account` we actually do use the variable in the checks (although I think it's not absolutely necessary). If you're not using it now, I suggest removing it.

includes/class-wcpdf-settings-debug.php
Spreeuw: The fact that it's expensive doesn't matter if you only run it when saving the settings.

alexmigf: Agreed

Spreeuw (approved):
Code looks good to me! I haven't tested the new permalinks flushing, but I have requested an additional review from @YordanSoares so that he can take this into account as well

includes/class-wcpdf-settings-debug.php
```diff
@@ -161,136 +161,148 @@ public function init_settings() {
...
-					'description'	=> __( "Document numbers (such as invoice numbers) are generated using AUTO_INCREMENT by default. Use this setting if your database auto increments with more than 1.", 'woocommerce-pdf-invoices-packing-slips' ),
+				'type'     => 'setting',
+				'id'       => 'pretty_document_links',
+				'title'    => __( 'Pretty document links', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'checkbox',
+				'section'  => 'debug_settings',
+				'args'     => array(
+					'option_name' => $option_name,
+					'id'          => 'pretty_document_links',
+					'description' => __( 'Changes the document links to a prettier URL scheme.', 'woocommerce-pdf-invoices-packing-slips' ),
```
YordanSoares, line 214: _I'd add an example after the descriptions:_
"Changes the document links to a prettier URL scheme with the following pattern: example.com/wcpdf/{$document_type}/{$order_id}/{$nonce||$order_id}."

The home URL and the identifier would be generated to match with the actual ones. Later, we can update this description when we have written a Docs article for this setting.

YordanSoares (commented):
### Tests performed:
1. ✔️ Enabled the **Pretty document links** setting and generated all the default documents: invoice, packing slip, proforma invoice, and credit note.
 **_Results_**: All documents opened correctly, following this pattern:
   - [example.com/wcpdf/document-type/order-number/nonce](#!)
2. ✔️ Enabled the **Allow guest access** setting and try opening all the default documents using the order key instead of the nonce.
 **_Results_**: All documents opened correctly, following this pattern:
   - [example.com/wcpdf/document-type/order-number/order-key](#!)
3. ✔️ Added a code snippet to customize the pretty link identifier from `wcpdf` to `document`, using the `wpo_wcpdf_pretty_document_link_identifier` filter hook, try opening all the default documents using the order key and the nonce with the pretty link patterns (see extra comments about this below).
 **_Results_**: All documents opened correctly with both patterns:
   - [example.com/document/document-type/order-number/nonce](#!)
   - [example.com/document/document-type/order-number/order-key](#!)
4. ✔️ Disable the **Allow guest access** setting and try opening all the default documents using the order key.
 **_Results_**: None of the documents opened, and I received the "_You do not have sufficient permissions to access this page._" message instead.
5. ✔️ Disable the **Pretty document links** setting and try opening all the default documents using the order key and the nonce with the pretty link pattern.
 **_Results_**: None of the documents opened, and I was redirected to the front-end.

includes/class-wcpdf-endpoint.php
```diff
@@ -0,0 +1,113 @@
...
+
+		if ( isset( $debug_settings['pretty_document_links'] ) ) {
+			return true;
+		} else {
+			return false;
+		}
+	}
+
+	public function get_identifier() {
+		return apply_filters( 'wpo_wcpdf_pretty_document_link_identifier', 'wcpdf' );
```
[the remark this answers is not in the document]
alexmigf: If you applied the filter after saving the setting, yes, you need to refresh the permalinks. I don't think we should do it again if we are using the filter, but let @Spreeuw give his opinion on this.

includes/class-wcpdf-settings-debug.php
alexmigf: @Spreeuw what's your opinion?

alexmigf: @YordanSoares I think the feature is good for now, we could improve it later if necessary, let's ship it and improve over time.

YordanSoares: @alexmigf I agree! 👍

YordanSoares (approved):
✔️ LGTM!

PR #380, opened 2022-06-16
Adds a notice for Pro in the Packing Slip settings

### LINKED ISSUE #379: Add notice for Pro on Packing Slip "Attach to" setting
The "Attach to" setting in the Packing Slip is only available using the Pro extension, so it will be nice to have a notice that could inform that in the same place of the setting.

alexmigf:
closes #379

Captura de ecrã de 2022-06-16 09-10-45


includes/documents/class-wcpdf-packing-slip.php
```diff
@@ -146,7 +146,29 @@ public function init_settings() {
 				)
 			),
 		);
-
+		
+		if ( ! function_exists( 'WPO_WCPDF_Pro' ) ) {
```
Spreeuw, line 150: maybe a double check here whether `attach_to_email_ids` is already present?

Spreeuw (changes requested):

includes/documents/class-wcpdf-invoice.php
```diff
@@ -182,7 +182,7 @@ public function init_settings() {
 			array(
 				'type'			=> 'setting',
 				'id'			=> 'attach_to_email_ids',
-				'title'			=> __( 'Attach to:', 'woocommerce-pdf-invoices-packing-slips' ),
+				'title'			=> __( 'Attach to', 'woocommerce-pdf-invoices-packing-slips' ),
```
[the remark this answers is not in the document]
Spreeuw: I see, but that will get rid of all existing translations

includes/documents/class-wcpdf-packing-slip.php
alexmigf: By present you mean saved? Because Pro adds it using the filter below this new setting: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/379-packing-slip-attach-to-notice/includes/documents/class-wcpdf-packing-slip.php#L174

Spreeuw: I mean that it's already in the array. But you're right, since `wpo_wcpdf_settings_fields_documents_packing_slip` is applied after this code, there's no point in checking!

includes/documents/class-wcpdf-invoice.php
```diff
@@ -196,7 +196,7 @@ public function init_settings() {
 			array(
 				'type'			=> 'setting',
 				'id'			=> 'disable_for_statuses',
-				'title'			=> __( 'Disable for:', 'woocommerce-pdf-invoices-packing-slips' ),
+				'title'			=> __( 'Disable for', 'woocommerce-pdf-invoices-packing-slips' ),
```
Spreeuw, line 199: same as with the other conversation: this will break translations, so let's keep the colon for now!

Spreeuw (changes requested):
Just one small comment, other than that it looks good to me!

Spreeuw (approved):
LGTM!

PR #381, opened 2022-06-22
378 fixes bug filtered template path

### LINKED ISSUE #378: Plugin does not list the custom template in my theme with a custom template_path defined.
My theme apparently filters `woocommerce_template_path` value from `woocommerce` to `shop`. So it should search for the following `shop/pdf` instead of `woocommerce/pdf/`. The UI in the setting, is showing the correct path, but when I manually debug the cache value, it does not include `shop/pdf/`. I found out that the caching was processed too early, bypassing the filter that was defined in my theme. 

Admin UI, showing correct path. The theme is called `bug`
image

Folder structure to template in theme.
image

Correct me if I am wrong.

This is causes the template to be cached too early. Bypassing any theme filters.
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5f3858cc25002fc83facf3a4ce8cc7059a80c246/includes/class-wcpdf-main.php#L33

This will call `get_template_path.`
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5f3858cc25002fc83facf3a4ce8cc7059a80c246/includes/class-wcpdf-main.php#L402

This will make calls too early, caching way too early, bypassing theme filters like 'woocommerce_template_path`.
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5f3858cc25002fc83facf3a4ce8cc7059a80c246/includes/class-wcpdf-settings.php#L446

Version: 2.15.0 (updated from previous version)
WP Version: 6.0.

alexmigf:
closes #378

When "General settings" are called the `WC()->template_path()` is not callable, which ends to the hard coded `woocommerce/` default. Applying the WooCommerce filter here catches potential custom filtered paths.

Captura de ecrã de 2022-06-22 16-15-58

### DIFF
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -479,10 +479,7 @@ class Settings {
 			return $this->installed_templates;
 		}
 
-		$installed_templates = array();
-
-		// call the default WooCommerce filter here because by default it runs later and we could miss it
-		
+		$installed_templates = array();	
 
 		// get base paths
 		$template_base_path = ( function_exists( 'WC' ) && is_callable( array( 'WC', 'template_path' ) ) ) ? WC()->template_path() : apply_filters( 'woocommerce_template_path', 'woocommerce/' );
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -479,28 +479,27 @@ class Settings {
 			return $this->installed_templates;
 		}
 
-		$installed_templates = array();	
-
+		$installed_templates = array();
 		// get base paths
-		$template_base_path = ( function_exists( 'WC' ) && is_callable( array( WC(), 'template_path' ) ) ) ? WC()->template_path() : apply_filters( 'woocommerce_template_path', 'woocommerce/' );
-		$template_base_path = untrailingslashit( $template_base_path );
-		$template_paths = array (
+		$template_base_path  = ( function_exists( 'WC' ) && is_callable( array( WC(), 'template_path' ) ) ) ? WC()->template_path() : apply_filters( 'woocommerce_template_path', 'woocommerce/' );
+		$template_base_path  = untrailingslashit( $template_base_path );
+		$template_paths      = array (
 			// note the order: theme before child-theme, so that child theme is always preferred (overwritten)
-			'default'		=> WPO_WCPDF()->plugin_path() . '/templates/',
-			'theme'			=> get_template_directory() . "/{$template_base_path}/pdf/",
-			'child-theme'	=> get_stylesheet_directory() . "/{$template_base_path}/pdf/",
+			'default'     => WPO_WCPDF()->plugin_path() . '/templates/',
+			'theme'       => get_template_directory() . "/{$template_base_path}/pdf/",
+			'child-theme' => get_stylesheet_directory() . "/{$template_base_path}/pdf/",
 		);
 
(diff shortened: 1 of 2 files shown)


YordanSoares:
@alexmigf @Spreeuw maybe this is too much trouble, but I wonder if there is possible to add support for both paths to extend compatibility to old setups? That is, to set the default path to `WC()->template_path()`, but also search in `child-theme/woocommerce` as well, just in case.

YordanSoares:
> @YordanSoares what do you mean by 'old setups'? 

I mean people that currently have their PDF custom template under `child-theme/woocommerce`, but its current theme has a custom path set under the `woocommerce_template_path` filter.

Spreeuw:
@alexmigf what I don't fully understand is this:
If I'm not mistaken, the bug was introduced when this line:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/fadf76b26ca08ec3bb9dc1cb64f49c1da62e1295/includes/class-wcpdf-settings.php#L485

was fixed, resulting in this:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/cbc315809d0559cc18120abd92622c48dda236f1/includes/class-wcpdf-settings.php#L485

But that would result in the opposite behavior of what you wrote, I think? 

> When "General settings" are called the `WC()->template_path()` is not callable, which ends to the hard coded `woocommerce/` default.

Because if I'm not mistaken, the previous code would always fall back to `/woocommerce` and this new code may or may not...

Are we calling the class earlier than we did before? Would using the `options_callback` parameter (like we do [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/c99084e1764a01c14cf3fa9793edad878af6a42c/includes/documents/class-wcpdf-invoice.php#L205)) also resolve the issue?

Spreeuw:
Yeah I see it now too, good find Alex! 🦾 
```php
is_callable( 'WC', 'template_path' )
```
in the old code would return true, because it is the equivalent of
```php
is_callable( 'WC' )
```
the new code (attempting to fix the incorrect syntax) accidentally made it worse (return false).

So it turns out we're not calling it too early after all 😅

Spreeuw:
> I understood that we were always using `child-theme/woocommerce`, ignoring the custom path here

No, it always used the custom template. So if people were following the documentation and they had that filter (which I think is somewhat rare), it would never work. We could mention this in the documentation, but this PR is just to restore previous behavior like it has been since the very first version of this plugin :)

includes/class-wcpdf-settings.php
```diff
@@ -479,10 +479,10 @@ public function get_installed_templates( $force_reload = false ) {
 			return $this->installed_templates;
 		}
 
-		$installed_templates = array();
+		$installed_templates = array();
```
Spreeuw, line 482: accidental space?

YordanSoares:
@Spreeuw oh, I understand now. I thought that `WC()->template_path()` actually was never called until the commit mentioned above, but I now understand that it did. Thanks for the explanation!

includes/class-wcpdf-settings.php
alexmigf: Yup, cleaning...

alexmigf:
@Spreeuw done!

Spreeuw (approved):

Issue #384, opened 2022-06-28
Disable the `[wcpdf_download_invoice]` shortcode output if the invoice is not allowed


YordanSoares:
Currently, if you use the `[wcpdf_download_invoice]` shortcode, it will display the **Download invoice (PDF)** text by default with the link to the invoice, even if it's actually disabled.

See: https://wordpress.org/support/topic/download-wcpdf-invoice-on-order-details/

PR #385, opened 2022-06-28
Check if the invoice is allowed before the shortcode output

YordanSoares:
Closes: #384


Spreeuw (approved):
Great improvement, thanks!

PR #399, opened 2022-08-30
Replaces WP_Filesystem with PHP functions to delete temporary files

### LINKED ISSUE #398: Erroneous use of WP_Filesystem / silent failure
In https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/main/includes/class-wcpdf-main.php#L875= , `WP_Filesystem()` is called and then the global `$wp_filesystem` object used to operate on files under the "uploads" (media) folder (default `wp-content/uploads`), in a cron context (specifically, deleting old temporary files).
This is incorrect - a valid WP install requires the uploads folder to be directly writeable. The WP_Filesystem API is not used by WP core for writing to that directory. In a cron context, a user cannot provide credentials. `WP_Filesystem()` looks at the parent folder. Consequently, if `wp-content` is not directly writeable, and `wp-content/uploads` is (as it should be), then the temporary file deletion fails. On the first site I looked at which was doing this, 8GB of temporary files going back 18 months had collected. (The plugin does not use the WP_Filesystem API to write the temporary files, only to attempt to delete them).
(Furthermore, the failure is silent - though the method returns an error, nothing is done with it).
Instead of using the WP_Filesystem API, the files should be read and deleted using ordinary PHP filesystem functions.

alexmigf:
closes #398

Tested :heavy_check_mark:

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -889,18 +889,7 @@ class Main {
 	 * @return array  Output message
 	 */
 	public function temporary_files_cleanup( $delete_timestamp = 0 ) {
-		global $wp_filesystem;
-
-		$delete_before = ! empty( $delete_timestamp ) ? intval( $delete_timestamp ) : time();
-
-		if ( ! function_exists( 'WP_Filesystem' ) ) {
-			require_once( ABSPATH . 'wp-admin/includes/file.php' );
-		}
-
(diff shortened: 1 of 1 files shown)


DavidAnderson684 (approved):
All good (code reviewed, tested successfully on site that previously had problems).

PR #401, opened 2022-09-08
Adds a fallback function to list files if for some reason the WP function isn't available

### LINKED ISSUE #400: Fatal error: Call to undefined function WPO\WC\PDF_Invoices\list_files()
A [user reported](https://wordpress.org/support/topic/fatal-error-in-newest-version/?view=all#post-15991254) the fatal error below in the forums:

```
CRITICAL Uncaught Error: Call to undefined function WPO\WC\PDF_Invoices\list_files() in /[REDACTED]/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-main.php:909
Stack trace:
#0 /[REDACTED]/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-main.php(882): WPO\WC\PDF_Invoices\Main->temporary_files_cleanup(1662014081)
#1 /[REDACTED]/public_html/wp-includes/class-wp-hook.php(307): WPO\WC\PDF_Invoices\Main->schedule_temporary_files_cleanup()
#2 /[REDACTED]/public_html/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#3 /[REDACTED]/public_html/wp-includes/plugin.php(524): WP_Hook->do_action(Array)
#4 /[REDACTED]/public_html/wp-cron.php(138): do_action_ref_array('wp_scheduled_de...', Array)
#5 {main}
  thrown i /[REDACTED]/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-main.php in line 909
```

Checking my own local setup I could find the same error.

Something is calling the delete temporary files function before `list_files()` is ready, presumably an AS action, which triggers the error above.

We should add a fallback if the function `list_files()` is not available.

alexmigf:
closes #400

Tested :heavy_check_mark:

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -906,7 +906,11 @@ class Main {
 		$output           = array();
 
 		foreach ( $paths_to_cleanup as $path ) {
-			$files = array_merge( $files, list_files( $path, $folders_level ) );
+			if ( function_exists( 'list_files' ) ) {
+				$files = array_merge( $files, list_files( $path, $folders_level ) );
+			} else {
+				$files = array_merge( $files, $this->dir_file_list( $path, $folders_level ) );
+			}
 		}
 
 		if ( ! empty( $files ) ) {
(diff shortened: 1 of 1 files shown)


DavidAnderson684 (commented):
@alexmigf Why not just `include_once(ABSPATH.'wp-admin/includes/file.php')` ? Then you won't need to include a second version of it.

alexmigf:
@DavidAnderson684 sorry, didn't came to my mind but of course, that's way better.

DavidAnderson684 (approved):

PR #403, opened 2022-09-22
Adds a specific query string for the shortcode

### LINKED ISSUE #402: Permission to access Invoice denied using shortcode
Some customers are reporting issues to see the Invoices using the shortcode in the frontend. Reading the code we give an exception to the `my-account` query string [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/3f2d2ca3e91b1f04365739ee9247db9593b8a22d/includes/class-wcpdf-main.php#L337) but we don't have one for the shortcode.

alexmigf:
closes #402 

Waiting customer feedback! https://secure.helpscout.net/conversation/2016246216/40280

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -336,7 +336,7 @@ class Main {
 			if ( ! $full_permission ) {
 				if ( ! isset( $_GET['my-account'] ) && ! isset( $_GET['shortcode'] ) ) {
 					$allowed = false;
-				} else { // User call from my-account page
+				} else { // User call from my-account page or via shortcode
 					// Only for single orders!
 					if ( count( $order_ids ) > 1 ) {
 						$allowed = false;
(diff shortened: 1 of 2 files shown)


includes/class-wcpdf-main.php
```diff
@@ -334,7 +334,7 @@ public function generate_pdf_ajax() {
 			// Check the user privileges
 			$full_permission = WPO_WCPDF()->admin->user_can_manage_document( $document_type );
 			if ( ! $full_permission ) {
-				if ( ! isset( $_GET['my-account'] ) ) {
+				if ( ! isset( $_GET['my-account'] ) || ! isset( $_GET['shortcode'] ) ) {
```
Terminator-Barbapapa, line 337: Right now you are breaking both the my account action button and the shortcode download link. One of these conditional statements will always return `true` because either `my-account` isn't set or `shortcode` isn't set. You will have to use `&&` instead of `||` here.

Terminator-Barbapapa (changes requested):

includes/class-wcpdf-main.php
alexmigf: Yep, my bad.

Terminator-Barbapapa: No problem. I also updated the comment two lines down.

Terminator-Barbapapa (approved):
LGTM! 🚀

PR #406, opened 2022-10-04
Filter to allow customers to access your PDF with a unique key.

### LINKED ISSUE #365: Add a filter to enable the guest access programmatically
Currently, the only way to enable the guest access is checking the **Allow guest access** option in the **Status** tab from the plugin's settings:

image

We could add a new filter hook to do the same programmatically.

jhosagid7:
closes #365

### DIFF
--- a/includes/class-wcpdf-endpoint.php
+++ b/includes/class-wcpdf-endpoint.php
@@ -66,7 +66,7 @@ class Endpoint {
 		}
 	}
 
-	public function get_document_link( $order, $document_type, $additional_vars = array(), $allow_guest_access = false ) {
+	public function get_document_link( $order, $document_type, $additional_vars = array() ) {
 		if ( empty( $order ) || empty( $document_type ) ) {
 			return '';
 		}
@@ -76,12 +76,7 @@ class Endpoint {
 		if ( is_user_logged_in() ) {
 			$access_key = wp_create_nonce( $this->action );
 		} elseif ( ! is_user_logged_in() && isset( $debug_settings['guest_access'] ) ) {
-			$allow_guest_access = apply_filters( 'wpo_wcpdf_allow_guest_access', $allow_guest_access );
-			if ( ! empty( $allow_guest_access ) && $allow_guest_access === true ) {
-				$access_key = wp_create_nonce( $this->action );
-			}else{
-				$access_key = $order->get_order_key();
-			}
+			$access_key = $order->get_order_key();
 		} else {
 			return '';
 		}
--- a/includes/class-wcpdf-endpoint.php
+++ b/includes/class-wcpdf-endpoint.php
@@ -66,7 +66,7 @@ class Endpoint {
 		}
 	}
 
-	public function get_document_link( $order, $document_type, $additional_vars = array() ) {
+	public function get_document_link( $order, $document_type, $additional_vars = array(), $allow_guest_access = false ) {
 		if ( empty( $order ) || empty( $document_type ) ) {
 			return '';
 		}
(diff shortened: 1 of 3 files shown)


includes/class-wcpdf-endpoint.php
```diff
@@ -76,7 +76,12 @@ public function get_document_link( $order, $document_type, $additional_vars = ar
 		if ( is_user_logged_in() ) {
 			$access_key = wp_create_nonce( $this->action );
 		} elseif ( ! is_user_logged_in() && isset( $debug_settings['guest_access'] ) ) {
```
YordanSoares, line 78: @jhosagid7 I think you only need to filter `$debug_settings['guest_access']` here. The rest of the new code within this function, including the new parameter, has no sense to me.

YordanSoares (commented):
You should filter this too: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5383926b3bfca54aa926f59ad572332239a75cc1/includes/class-wcpdf-main.php#L259

alexmigf:
@jhosagid7 this is what I wrote in Slack: 

> Create a custom function to check if the option is enabled, and having a filter to bypass the option value
> Every time in the code that you need to check if "Guest access" is enabled, you need to replace that code with the new function

Example:
```php
/**
 * Checks if guest access is enabled
 * 
 * @return bool
 */
public function is_guest_access_enabled() {
	// return bool with filter applied
}
```

You could check the current value like this:

```php
$guest_access = isset( WPO_WCPDF()->settings->debug_settings['guest_access'] );
```

includes/class-wcpdf-settings.php
```diff
@@ -746,6 +747,21 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+	/**
+	 * Checks if guest access is enabled
+	 * 
+	 * @return bool
+	 */
+	public function is_guest_access_enabled() {
+		// return bool with filter applied
+		$guest_access = isset( $this->debug_settings['guest_access'] ) ? true : false;
+
+		$value_guest_access = apply_filters( 'wpo_wcpdf_guest_access_enabled', $guest_access );
```
YordanSoares, line 759: @jhosagid7, you aren't returning this variable, but `$guest_access`.

includes/class-wcpdf-settings.php
```diff
@@ -746,6 +747,21 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+
+		return $guest_access;
```
YordanSoares, line 761: I think you can move the filter to this place instead of wrapping it `$value_guest_access`

jhosagid7: Done

includes/class-wcpdf-settings.php
```diff
@@ -746,6 +747,17 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
 		return $new_settings;
 	}
 
+	/**
+	 * Checks if guest access is enabled
+	 * 
+	 * @return bool
+	 */
+	public function is_guest_access_enabled() {
+		// return bool with filter applied
```
YordanSoares, line 756: Jhonny, this is not very important, but I think this comment is not necessary.

YordanSoares (approved):
✔️ LGTM!

includes/class-wcpdf-settings.php
jhosagid7: Done

YordanSoares: You marked this as resolved, but the comment is still there.

YordanSoares: By the way, this change doesn't require a new review 😉

YordanSoares (approved):

includes/class-wcpdf-settings.php
jhosagid7: Done

includes/class-wcpdf-settings.php
```diff
@@ -746,6 +746,16 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
 
+	/**
+	 * Checks if guest access is enabled
+	 * 
+	 * @return bool
+	 */
+	public function is_guest_access_enabled() {
+		$guest_access = isset( $this->debug_settings['guest_access'] ) ? true : false;
+
+		return apply_filters( 'wpo_wcpdf_guest_access_enabled', $guest_access );
```
alexmigf, line 757: I would also add `$this` as secondary arg to the filter.

alexmigf (changes requested):

alexmigf (approved):
Good work!

Issue #411, opened 2022-10-07
Run action hooks within invoice columns under order list, even if the invoice is disabled


YordanSoares:
Currently, if the invoice is disabled, both `wcpdf_invoice_number_column_end` and `wcpdf_invoice_date_column_end` action hooks aren't reached, therefore, if you hook some code there, it won't run. We should always run these action hooks, whether the invoice is available or not.

See: https://secure.helpscout.net/conversation/2030334571/40610/#thread-6032542305

PR #412, opened 2022-10-07
Run action hooks within invoice columns under order list, even if the…

YordanSoares:
… invoice is disabled

Closes: #411

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -268,8 +268,8 @@ class Admin {
 				do_action( 'wcpdf_invoice_number_column_end', $order );
 				break;
 			case 'invoice_date_column':
-				$invoice_date = ! empty( $invoice ) && ! empty( $invoice->get_date() ) ? $invoice->get_date() : '';
-				echo $invoice_date->date_i18n( wcpdf_date_format( $invoice, 'invoice_date_column' ) );
+				$invoice_date = ! empty( $invoice ) && ! empty( $invoice->get_date() ) ? $invoice->get_date()->date_i18n( wcpdf_date_format( $invoice, 'invoice_date_column' ) ) : '';
+				echo $invoice_date;
 				do_action( 'wcpdf_invoice_date_column_end', $order );
 				break;
 			default:
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -249,8 +249,6 @@ class Admin {
 	 */
 	public function invoice_columns_data( $column ) {
 		global $post, $the_order;
-		$invoice_number = NULL;
-		$invoice_date = NULL;
 
 		$this->disable_storing_document_settings();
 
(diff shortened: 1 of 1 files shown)


alexmigf (changes requested):
I think you just need to do this before echoing:

```php
$invoice_number = ! empty( $invoice ) && ! empty( $invoice->get_number() ) ? $invoice->get_number() : '';
```

includes/class-wcpdf-admin.php
```diff
@@ -262,11 +260,8 @@ public function invoice_columns_data( $column ) {
 		}
 
 		$invoice = wcpdf_get_invoice( $order );
-		
-		if ( ! empty( $invoice ) ) {
-			$invoice_number = $invoice->get_number();
-			$invoice_date = $invoice->get_date();
-		}
+		$invoice_number = ! empty( $invoice ) && ! empty( $invoice->get_number() ) ? $invoice->get_number() : '';
+		$invoice_date = ! empty( $invoice ) && ! empty( $invoice->get_date() ) ? $invoice->get_date() : '';
```
alexmigf, line 264: I would do this inside the switch statement, no need to retrieve the date if you're want to have the number for example.

alexmigf (changes requested):

includes/class-wcpdf-admin.php
YordanSoares: Done here: [8cfb4b0](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/412/commits/8cfb4b02d493cf79b5248953f102254d2c1f4ab9)

includes/class-wcpdf-admin.php
```diff
@@ -261,18 +261,16 @@ public function invoice_columns_data( $column ) {
...
-				echo $invoice->get_number();
+				$invoice_number = ! empty( $invoice ) && ! empty( $invoice->get_number() ) ? $invoice->get_number() : '';
+				echo $invoice_number;
 				do_action( 'wcpdf_invoice_number_column_end', $order );
 				break;
 			case 'invoice_date_column':
-				if ( ! empty( $date = $invoice->get_date() ) ) {
-					echo $date->date_i18n( wcpdf_date_format( $invoice, 'invoice_date_column' ) );
+				$invoice_date = ! empty( $invoice ) && ! empty( $invoice->get_date() ) ? $invoice->get_date() : '';
+				if ( ! empty( $invoice_date ) ) {
```
alexmigf, line 272: This is not needed.

alexmigf (changes requested):

includes/class-wcpdf-admin.php
YordanSoares: Done here 68c2313

includes/class-wcpdf-admin.php
```diff
@@ -261,19 +261,15 @@ public function invoice_columns_data( $column ) {
...
+				$invoice_number = ! empty( $invoice ) && ! empty( $invoice->get_number() ) ? $invoice->get_number() : '';
+				echo $invoice_number;
 				do_action( 'wcpdf_invoice_number_column_end', $order );
 				break;
 			case 'invoice_date_column':
-				if ( ! empty( $date = $invoice->get_date() ) ) {
-					echo $date->date_i18n( wcpdf_date_format( $invoice, 'invoice_date_column' ) );
-				}
+				$invoice_date = ! empty( $invoice ) && ! empty( $invoice->get_date() ) ? $invoice->get_date() : '';
+				echo $invoice_date->date_i18n( wcpdf_date_format( $invoice, 'invoice_date_column' ) );
```
alexmigf, line 272: I think you need to make this call in the line above, because if `$invoice_date` is empty the `date_i18n` will throw an error.

alexmigf (changes requested):

includes/class-wcpdf-admin.php
alexmigf: Yes, perfect!

alexmigf (approved):

PR #419, opened 2022-10-13
Removes WC compatibility classes. Bumps WC minimum version to 3.0

### LINKED ISSUE #413: Check HPOS compatibility
We should check [HPOS](https://developer.woocommerce.com/2022/09/29/high-performance-order-storage-backward-compatibility-and-synchronization/) compatibility by using the default WC API instead of the WP postmeta functions.

Remove compatibility class and bump WC minimum version to 3.0.0

alexmigf:
closes #413


DavidAnderson684 (approved):

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -238,6 +243,29 @@ public function legacy_textdomain_enabled() {
...
+			return false;
+		}
+	}
+
+	/**
+	 * Woocommerce version < 3.0 notice
+	 */
+	public function need_woocommerce_3_0() {
+		/* translators: <a> tags */
+		$error = sprintf( esc_html__( 'PDF Invoices & Packing Slips for WooCommerce requires %1$sWooCommerce%2$s version 3.0 or superior!' , 'woocommerce-pdf-invoices-packing-slips' ), '<a href="http://wordpress.org/extend/plugins/woocommerce/">', '</a>' );
```
Terminator-Barbapapa, line 262: I would say 'higher' instead of 'superior'.

Terminator-Barbapapa (approved):

Issue #421, opened 2022-10-20
PDF documents return a fatal error if the shop logo image file is deleted from server when using mPDF extension


YordanSoares:
[This user](https://wordpress.org/support/topic/mpdf-extension-produces-pdf-of-0-bytes/) reported that the PDF document crashes when the shop logo was removed from the media content.

I tried to reproduce this behavior and confirmed that this happens, but not when the image is removed from the Media Library, but when the image is removed from the server directly, and the reference remains in WordPress database.

Below can be seen the behavior of this scenario with the mPDF extension activated and deactivated.

## Testing with the mPDF extension activated:

This is the error returned while trying to generate the PDF document with the mPDF extension activated:

![image](https://user-images.githubusercontent.com/38109855/197041704-122b0bf8-72ec-47d4-8ed8-4c9e3b3a67b3.png)

## Testing with the mPDF extension deactivated:

The PDF document is generated with our default PDF engine (dompdf), only returning the message "_Image not found or type unknown_":

![image](https://user-images.githubusercontent.com/38109855/197041937-176ed05e-69bb-4e30-9e09-ad1472214138.png)

PR #422, opened 2022-10-20
Check if the shop logo exists

YordanSoares:
Closes: #421

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -624,18 +624,20 @@ abstract class Order_Document {
 	 * Show logo html
 	 */
 	public function header_logo() {
-		if ($this->get_header_logo_id()) {
+		if ( $this->get_header_logo_id() ) {
 			$attachment_id = $this->get_header_logo_id();
-			$company = $this->get_shop_name();
-			if( $attachment_id ) {
-				$attachment = wp_get_attachment_image_src( $attachment_id, 'full', false );
+			$company       = $this->get_shop_name();
+
+			if ( $attachment_id ) {
+				$attachment      = wp_get_attachment_image_src( $attachment_id, 'full', false );
 				$attachment_path = get_attached_file( $attachment_id );
+
 				if ( empty( $attachment ) || empty( $attachment_path ) ) {
 					return;
 				}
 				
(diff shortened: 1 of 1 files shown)


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -640,8 +640,10 @@ public function header_logo() {
 
 				if ( apply_filters( 'wpo_wcpdf_use_path', true ) && file_exists( $attachment_path ) ) {
 					$src = $attachment_path;
-				} else {
+				} elseif ( file_exists( $attachment_src ) ) {
```
alexmigf, line 643: The `$attachment_src` is an URL not a local path, I think you must use this function: [wp_remote_head](https://developer.wordpress.org/reference/functions/wp_remote_head/)

The you should check if the response header code is `200`, which corresponds to success (exists).

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -624,27 +624,31 @@ public function get_header_logo_height() {
...
-				$attachment_src = $attachment[0];
-				$attachment_width = $attachment[1];
+				$attachment_src    = $attachment[0];
+				$attachment_width  = $attachment[1];
 				$attachment_height = $attachment[2];
 
 				if ( apply_filters( 'wpo_wcpdf_use_path', true ) && file_exists( $attachment_path ) ) {
 					$src = $attachment_path;
-				} else {
+				} elseif ( wp_remote_head( $attachment_src, [ 'sslverify' => false ] )['response']['code'] === 200 ) {
```
alexmigf, line 645: I think you probably should check the response first before asking for the keys, like this:

```php
elseif( ( is_array( $head = wp_remote_head( $attachment_src, [ 'sslverify' => false ] ) ) ) && $head['response']['code'] == 200 ) )
```

YordanSoares: Done here: 0601f05

alexmigf:
@YordanSoares I have made some adjustments, let me know what you think.

YordanSoares:
@alexmigf your improvements are great: Now it checks almost all what comes to my mind!

Also, I ran some tests and haven't any issues 👍

PR #424, opened 2022-10-21
Maybe reinstall fonts before PDF output

### LINKED ISSUE #423: Check if fonts are installed before trying to generate PDF
Currently if a user migrates the plugin to a new site without copying the fonts, it will break the PDF without notice. We should check if fonts exist before trying to generate a PDF.

alexmigf:
closes #423

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -523,6 +523,80 @@ class Main {
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


DavidAnderson684 (approved):

YordanSoares (approved):
✔️ LGTM!

Issue #428, opened 2022-10-26
Allow formatting HTML tags for product titles


YordanSoares:
Currently, after merging #408, we strip all the HTML tags for the product titles on the PDF documents. However, there are some cases when the common formatting HTML tags are desirable, like the case reported for [this user](https://wordpress.org/support/topic/html-tags-in-pdf/), in which they're applying styles directly in the product titles, like this:

![image](https://user-images.githubusercontent.com/38109855/198010111-fb3bb90d-45c3-46a8-af89-2dabe0e4fec0.png)

alexmigf:
Like we discuss on Slack, this issue appears to be related to some code introduced in this PR https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/408. We need to revert that code to the previous state.

PR #408, opened 2022-10-05
Check if HTML is escaped properly before echoing

alexmigf:
closes #407

PR #429, opened 2022-10-26
Replace `esc_html()` with `wp_kses_post()` to allow basic HTML format…

YordanSoares:
…ting

#428


alexmigf:
We need to revert some code from https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/408, closing this in favor of a new PR.

Issue #430, opened 2022-10-26
Stop PDF rendering if the order is unsaved or doesn't exist


YordanSoares:
Currently, if a manual order is unsaved, and you click on the PDF buttons, the HTML for the PDF rendering will crash:

![image](https://user-images.githubusercontent.com/38109855/198029293-e61aca4a-cd6b-4f4b-83ba-2a7fed89a3ab.png)

```
Fatal error: Call to a member function date_i18n() on null
C:\Users\Yordan Soares\Local Sites\pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document-methods.php (518)
#0 \pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document-methods.php(523): WPO\WC\PDF_Invoices\Documents\Order_Document_Methods->get_order_date()
#1 \pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\templates\Simple\invoice.php(83): WPO\WC\PDF_Invoices\Documents\Order_Document_Methods->order_date()
#2 \pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document.php(942): include('C:\\Users\\Yordan...')
#3 \pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document.php(843): WPO\WC\PDF_Invoices\Documents\Order_Document->render_template('C:/Users/Yordan...', Array)
#4 \pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document.php(798): WPO\WC\PDF_Invoices\Documents\Order_Document->get_html()
#5 \pdf-invoices\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document.php(864): WPO\WC\PDF_Invoices\Documents\Order_Document->get_pdf()
[...]

PR #431, opened 2022-10-26
Check if the order is unsaved or doesn't exist before rendering the PDF document

YordanSoares:
Closes: #430


alexmigf:
@YordanSoares I hope you don't mind but I did a few changes, could you please review?

YordanSoares:
@alexmigf no problem! Your code is better and works fine: I have already tested it!

PR #435, opened 2022-10-28
Reverts documents functions escape

### LINKED ISSUE #428: Allow formatting HTML tags for product titles

alexmigf:
closes #428

Reverts documents functions/templates escaping done in #408

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -255,12 +255,12 @@ class Admin {
 
 		switch ( $column ) {
 			case 'invoice_number_column':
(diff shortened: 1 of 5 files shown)


YordanSoares (approved):
✔️ LGTM!

testing-pdf-documents-output

dwalkerpriv (approved):
All good for me
Screenshot 2022-10-28 060008

Terminator-Barbapapa (approved):

PR #437, opened 2022-10-28
Fixes bug on getting the order ID from third party object

### LINKED ISSUE #436: Call to undefined method Class::get_id() on attach_pdf_to_email()
A customer reported an error when attaching PDF to custom WC emails here:

```
Uncaught Error: Call to undefined method CustomClass::get_id() in /home/customer/www/colibribag.com/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-main.php:80
```

This bug seems to be introduced with #419

alexmigf:
closes #436

Waiting customer feedback :hourglass:

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -77,7 +77,7 @@ class Main {
 			return $attachments;
 		}
 
-		$order_id = $order->get_id();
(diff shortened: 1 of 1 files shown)


Terminator-Barbapapa (approved):
LGTM!

PR #441, opened 2022-10-31
Fixes warning on deprecated argument from product get_dimentions() method

### LINKED ISSUE #440: Backend error after updating 3.2.3
After the 3.2.3 update you get an error in the webserver error log.

[lsapi:warn] [pid 27798:tid 140198411142912] [client 2a06:a463:54a1:1:f881:af13:982b:dcc3:63774] [host www.test.nl] Backend log: The WC_Product::get_dimensions argument is deprecated since version 3.0. By default, get_dimensions has an argument set to true so that HTML is returned. This is to support the legacy version of the method. To get HTML dimensions, instead use wc_format_dimensions() function. Pass false to this method to return an array of dimensions. This will be the new default behavior in future versions.\n, referer: https://www.test.nl/wp-admin/edit.php?post_type=shop_order

Revert back to 3.2.1 will solve it.

alexmigf:
closes #440


YordanSoares (approved):
✔️ LGTM!

PR #455, opened 2022-12-29
Fix: Add nonces to several admin unsecure requests

### LINKED ISSUE #454: Add nonces to several admin unsecure requests
We have several requests in the Admin panel that are unsecured with nonce. Is not critical as it happens in the backend, but we should fix them:

Captura de ecrã de 2022-12-28 08-39-10

jhosagid7:
Closes #454

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -56,10 +56,10 @@ class Admin {
 		if ( get_option( 'wpo_wcpdf_review_notice_dismissed' ) !== false ) {
 			return;
 		} else {
-			if ( isset( $_REQUEST['wpo_wcpdf_dismis_review'] ) && isset( $_REQUEST['_wpnonce'] ) ) {
+			if ( isset( $_REQUEST['wpo_wcpdf_dismiss_review'] ) && isset( $_REQUEST['_wpnonce'] ) ) {
 				// validate nonce
-				if ( ! wp_verify_nonce( $_REQUEST['_wpnonce'], 'dismis_review_nonce' ) ) {
-					wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_wcpdf_dismis_review' );
+				if ( ! wp_verify_nonce( $_REQUEST['_wpnonce'], 'dismiss_review_nonce' ) ) {
+					wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_wcpdf_dismiss_review' );
 					return;
 				} else {
 					update_option( 'wpo_wcpdf_review_notice_dismissed', true );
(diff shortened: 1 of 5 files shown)


includes/class-wcpdf-admin.php
```diff
@@ -56,9 +56,15 @@ public function review_plugin_notice() {
 		if ( get_option( 'wpo_wcpdf_review_notice_dismissed' ) !== false ) {
 			return;
 		} else {
-			if ( isset( $_GET['wpo_wcpdf_dismis_review'] ) ) {
-				update_option( 'wpo_wcpdf_review_notice_dismissed', true );
-				return;
+			if ( isset( $_REQUEST['wpo_wcpdf_dismis_review'] ) && isset( $_REQUEST['_wpnonce'] ) ) {
```
alexmigf, line 59: Is not your fault because it was like this before, but `dismiss` has two `s`

includes/class-wcpdf-main.php
```diff
@@ -715,16 +715,23 @@ public function no_dir_notice() {
...
-						wp_redirect( 'admin.php?page=wpo_wcpdf_options_page' );
-						exit;
+					if ( isset( $_REQUEST['wpo_wcpdf_hide_no_dir_notice'] ) && isset( $_REQUEST['_wpnonce'] ) ) {
+						// validate nonce
+						if ( ! wp_verify_nonce( $_REQUEST['_wpnonce'], 'hide_no_dir_notice_nonce' ) ) {
+							wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_wcpdf_hide_no_dir_notice' );
+							wp_redirect( 'admin.php?page=wpo_wcpdf_options_page' );
+							exit;
+						} else {
+							delete_option( 'wpo_wcpdf_no_dir_error', true );
```
alexmigf, line 731: Not your fault as well, but the `delete_option()` just has 1 argument, you could delete the `true`.

alexmigf (changes requested):
Great work!

includes/class-wcpdf-admin.php
jhosagid7: I'm not sure if the change I made is correct. Please correct me if it is not correct.

includes/class-wcpdf-main.php
jhosagid7: Done!

alexmigf (commented):
Still missing the `dismiss` fix on this function: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/85ab4a7511ada12327454ace4ecd22cafc79ba1e/includes/class-wcpdf-admin.php#L113-L151

alexmigf (approved):

Issue #456, opened 2023-01-06
Add new filters to allow user to disable the documents' data removal


YordanSoares:
Currently, if the users enable the "Remove personal data" feature on WooCommerce, our plugin removes the documents' data when the action is applied to an order ([source](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/99940b0469db57740cba0bd48e903a9ecbc2abd8)).

Adding a filter will allow users to disable this behavior, if they don't want to remove the documents' data.

PR #457, opened 2023-01-06
Add new filter to allow user to disable the documents' data removal

YordanSoares:
This PR introduces the `wpo_wcpdf_remove_order_personal_data_meta` filter hook, which allows you to override the default behavior of the document data removal, enabled by default.

You can use the filter hook to disable the documents' data removal, this way:

```php
add_filter( 'wpo_wcpdf_remove_order_personal_data', '__return_false' );
```

Closes: #456

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -1054,13 +1054,13 @@ class Main {
 	 * Remove all invoice data when requested
 	 */
 	public function remove_order_personal_data_meta( $meta_to_remove ) {
-			$wcpdf_private_meta = array(
-				'_wcpdf_invoice_number'         => 'numeric_id',
-				'_wcpdf_invoice_number_data'    => 'array',
-				'_wcpdf_invoice_date'           => 'timestamp',
-				'_wcpdf_invoice_date_formatted' => 'date',
-			);
-		return $meta_to_remove;
+		$wcpdf_private_meta = array(
+			'_wcpdf_invoice_number'         => 'numeric_id',
+			'_wcpdf_invoice_number_data'    => 'array',
+			'_wcpdf_invoice_date'           => 'timestamp',
+			'_wcpdf_invoice_date_formatted' => 'date',
+		);
+		return $meta_to_remove + $wcpdf_private_meta;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


alexmigf (commented):
Wouldn't be better to apply the filter [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/79fd56bbc0c9d751ac98ef480ffa2019f71b21c4/includes/class-wcpdf-main.php#L45-L46) for both functions?

YordanSoares:
> Wouldn't be better to apply the filter [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/79fd56bbc0c9d751ac98ef480ffa2019f71b21c4/includes/class-wcpdf-main.php#L45-L46) for both functions?

@alexmigf I agree. Done here: [e58e3a0](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/457/commits/e58e3a004574226719887d99e04d7761119dd096)

includes/class-wcpdf-main.php
```diff
@@ -42,8 +42,10 @@ public function __construct() {
 		add_action( 'wp_scheduled_delete', array( $this, 'schedule_temporary_files_cleanup' ) );
 
 		// remove private data
-		add_action( 'woocommerce_privacy_remove_order_personal_data_meta', array( $this, 'remove_order_personal_data_meta' ), 10, 1 );
-		add_action( 'woocommerce_privacy_remove_order_personal_data', array( $this, 'remove_order_personal_data' ), 10, 1 );
+		if ( true === apply_filters( 'wpo_wcpdf_remove_order_personal_data_meta', true ) ) {
```
alexmigf, line 45: Maybe like this is enough? I have renamed the filter, makes more sense to me like this.

```php
if ( apply_filters(  'wpo_wcpdf_remove_order_personal_data', true ) ) {
```

alexmigf (changes requested):

includes/class-wcpdf-main.php
YordanSoares: That makes sense. I followed the way we check this in other places, but I like more your suggestion.

alexmigf (approved):

PR #458, opened 2023-01-09
Replaces yearly debug enable property with filter

alexmigf:
I was thinking on this over the weekend, if for some reason we forgot the debug enabled and push changes to the branch and nobody notice could be problematic for customers, so I think we should use a filter instead.


Terminator-Barbapapa (approved):
Good thinking. I would call the filter `wpo_wcpdf_enable_yearly_reset_debug`.

Terminator-Barbapapa (approved):
💯

PR #459, opened 2023-01-10
Bail is document data is empty when saving order

alexmigf:
I notice in a customer store strange order comments, and after investigating I found that when the order is saved, for example when changing the order status to completed, our plugin function `save_invoice_number_date()` is triggered. Because the document data is empty, and we don't do a check on that, the order comments were registered anyway here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5048ca51bcfed8db1eba4e0f5ae72ecdbf01bfa6/includes/class-wcpdf-admin.php#L692

These are the wrongly registered comments:

Captura de ecrã de 2023-01-10 12-40-10

The PDF in this case was generated in the completed email trigger.

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -663,17 +663,21 @@ class Admin {
 	 */
 	public function save_invoice_number_date($post_id, $post) {
 		$post_type = get_post_type( $post_id );
-		if( $post_type == 'shop_order' ) {
+		if ( $post_type == 'shop_order' ) {
+			
 			// bail if this is not an actual 'Save order' action
-			if ( ! isset($_POST['action']) || $_POST['action'] != 'editpost' ) {
+			if ( ! isset( $_POST['action'] ) || $_POST['action'] != 'editpost' ) {
 				return;
 			}
+			
 			// Check if user is allowed to change invoice data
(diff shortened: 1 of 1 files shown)


includes/class-wcpdf-admin.php
```diff
@@ -675,9 +675,13 @@ public function save_invoice_number_date($post_id, $post) {
 
 			$order = wc_get_order( $post_id );
 			if ( $invoice = wcpdf_get_invoice( $order ) ) {
-				$is_new = false === $invoice->exists();
-				$_POST = stripslashes_deep( $_POST );
-				$document_data = $this->process_order_document_form_data( $_POST, $invoice->slug );
+				$is_new        = false === $invoice->exists();
+				$request_data  = stripslashes_deep( $_POST );
```
Terminator-Barbapapa, line 679: Perhaps rename this to `$form_data` as this is the parameter name we use in the `process_order_document_form_data` function.
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5048ca51bcfed8db1eba4e0f5ae72ecdbf01bfa6/includes/class-wcpdf-admin.php#L952

Terminator-Barbapapa (approved):
Left one comment on renaming a variable. Other than that looks good!

PR #461, opened 2023-01-11
Adds a document object arg to PDF maker class

alexmigf:
A customer required a dynamic paper height for Packing Slip using mPDF extension. Currently the PDF maker class doesn't have a way to identify the current document. This PR addresses this by passing the document object in the construct arguments and also to render filters.


Terminator-Barbapapa:
I think in this case passing just the document type, instead of the whole document object, is sufficient right?

Terminator-Barbapapa (approved):

PR #462, opened 2023-01-12
FIx: Replace the use of the deprecated `wcs_renewal_order_meta` hook

### LINKED ISSUE #460: Deprecated WooCommerce Subscription hooks used
Hi :wave:

Hope this is the right place to report issues with the plugin.

We're trying to use this plugin in combination with the "official" [WooCommerce Subscriptions](https://woocommerce.com/products/woocommerce-subscriptions/) plugin, version 4.7.0. The authors of that plugin have deprecated a hook used by this plugin in such a way that renewing subscriptions becomes impossible.

When renewing a subscription (interactively, but this also affects automatic renewals) the following notice is displayed:

> We couldn't create a renewal order for your subscription, please try again.

This is caused by a WP_Error thrown by the subscription plugins filter deprecator:

> Hook wcs_renewal_order_meta is deprecated since version wcs-core 2.5.0! Use wc_subscriptions_renewal_order_data instead

The action is added at: 

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/5048ca51bcfed8db1eba4e0f5ae72ecdbf01bfa6/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php#L20

I can confirm that temporarily removing the action (in addition to the one from https://wordpress.org/support/topic/deprecated-subscription-hooks-prevent-subscription-renewal/), fixes the renewal issues.

jhosagid7:
Closes #460

### DIFF
--- a/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
+++ b/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
@@ -13,14 +13,13 @@ if ( ! class_exists( '\\WPO\\WC\\PDF_Invoices\\Compatibility\\Third_Party_Plugin
 class Third_Party_Plugins {
 	function __construct()	{
 		// WooCommerce Subscriptions compatibility
-		if ( class_exists('WC_Subscriptions') ) {
+		if ( class_exists( 'WC_Subscriptions' ) ) {
 			if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
 				add_action( 'woocommerce_subscriptions_renewal_order_created', array( $this, 'woocommerce_subscriptions_renewal_order_created' ), 10, 4 );
 			} elseif ( version_compare( \WC_Subscriptions::$version, '2.5', '<' ) ) {
 				add_filter( 'wcs_renewal_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
 				add_filter( 'wcs_resubscribe_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
 			} else {
-				// Replace the use of the deprecated wcs_renewal_order_meta hook
 				add_filter( 'wc_subscription_renewal_order_data', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
 				add_filter( 'wc_subscriptions_resubscribe_order_data', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
 			}
--- a/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
+++ b/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
@@ -16,11 +16,9 @@ class Third_Party_Plugins {
 		if ( class_exists('WC_Subscriptions') ) {
 			if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
 				add_action( 'woocommerce_subscriptions_renewal_order_created', array( $this, 'woocommerce_subscriptions_renewal_order_created' ), 10, 4 );
-			} 
-
-			if ( version_compare( \WC_Subscriptions::$version, '2.0', '>=' ) && version_compare( \WC_Subscriptions::$version, '2.5', '<' ) ) {
-				add_action( 'wcs_renewal_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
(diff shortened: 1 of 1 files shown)


jhosagid7:
> Also check if there is also a replacement for the wcs_resubscribe_order_meta hook and do the same.

This hook changes in version 4.7.0 from `wcs_resubscribe_order_meta` to `wc_subscriptions_resubscribe_order_data`.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -16,9 +16,15 @@ function __construct()	{
...
-			} else {
+			} 
+
+			if ( version_compare( \WC_Subscriptions::$version, '2.0', '>=' ) && version_compare( \WC_Subscriptions::$version, '2.5', '<' ) ) {
 				add_action( 'wcs_renewal_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
 				add_action( 'wcs_resubscribe_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
+			} else {
+				// Replace the use of the deprecated wcs_renewal_order_meta hook
+				add_action( 'wc_subscription_renewal_order_data', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
+				add_action( 'wcs_resubscribe_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
```
alexmigf, line 27: This hook is wrong, should be `wc_subscriptions_resubscribe_order_data`

alexmigf (changes requested):

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -16,9 +16,15 @@ function __construct()	{
 		if ( class_exists('WC_Subscriptions') ) {
 			if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
 				add_action( 'woocommerce_subscriptions_renewal_order_created', array( $this, 'woocommerce_subscriptions_renewal_order_created' ), 10, 4 );
-			} else {
+			} 
+
+			if ( version_compare( \WC_Subscriptions::$version, '2.0', '>=' ) && version_compare( \WC_Subscriptions::$version, '2.5', '<' ) ) {
 				add_action( 'wcs_renewal_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
```
alexmigf, line 22: You should replace `add_action` by `add_filter` here too.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -16,9 +16,15 @@ function __construct()	{
 		if ( class_exists('WC_Subscriptions') ) {
 			if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
 				add_action( 'woocommerce_subscriptions_renewal_order_created', array( $this, 'woocommerce_subscriptions_renewal_order_created' ), 10, 4 );
-			} else {
+			} 
+
+			if ( version_compare( \WC_Subscriptions::$version, '2.0', '>=' ) && version_compare( \WC_Subscriptions::$version, '2.5', '<' ) ) {
```
alexmigf, line 21: This should be an `elseif` from the `if` above, like this:

```php
if ( version_compare( \WC_Subscriptions::$version, '2.0', '<' ) ) {
	add_action( 'woocommerce_subscriptions_renewal_order_created', array( $this, 'woocommerce_subscriptions_renewal_order_created' ), 10, 4 );
} elseif ( version_compare( \WC_Subscriptions::$version, '2.5', '<' ) ) {
	add_filter( 'wcs_renewal_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
	add_filter( 'wcs_resubscribe_order_meta', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
} else {
	add_filter( 'wc_subscription_renewal_order_data', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
	add_filter( 'wc_subscriptions_resubscribe_order_data', array( $this, 'wcs_renewal_order_meta' ), 10, 3 );
}
```

alexmigf (changes requested):

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
jhosagid7: Yes, I'm sorry, you're right. 😢

PR #465, opened 2023-01-18
Fix: displays a message if the typed next number is superior to MySQL INT max

### LINKED ISSUE #256: Show a notice to warn users when entered a number greater than the max number allowed to 'Next invoice number'
The maximum value signed by SQL for `int` is `2147483647` (See [Integer Types](https://dev.mysql.com/doc/refman/8.0/en/integer-types.html)), so we could to set the max value allowed for **Next invoice number** to `2147483647` and show a notice to warn the user about it, if they entered a number greater than this.

See: https://wordpress.org/support/topic/problem-with-invoice-starting-number/

alexmigf:
closes #256

### DIFF
--- a/includes/class-wcpdf-assets.php
+++ b/includes/class-wcpdf-assets.php
@@ -156,6 +156,7 @@ class Assets {
 						),
 					),
 					'dismissed_pointers'        => get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ),
+					'mysql_int_size_limit'      => __( "The typed number is superior to the MySQL allowed INT size, which is 2147483647. Perhaps your number includes the current year? If that's the case you should use the prefix/suffix fields for that.", 'woocommerce-pdf-invoices-packing-slips' ),
 				)
 			);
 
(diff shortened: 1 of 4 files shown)


includes/class-wcpdf-assets.php
```diff
@@ -156,6 +156,7 @@ public function backend_scripts_styles ( $hook ) {
 						),
 					),
 					'dismissed_pointers'        => get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ),
+					'mysql_int_size_limit'      => __( "The typed number is superior to the MySQL allowed INT size, which is 2147483647. Perhaps your number includes the current year? If that's the case you should use the prefix/suffix fields for that.", 'woocommerce-pdf-invoices-packing-slips' ),
```
Terminator-Barbapapa, line 159: I don't think we need to explain the precise technical limitations. I would go more for something like this:
```
The number should be smaller than 2147483647. Please note you should add your next document number without prefix, suffix or padding. 
```

Terminator-Barbapapa (approved):

YordanSoares (approved):
✔️ LGTM!

image

PR #467, opened 2023-01-18
Fix: allow preview gutters text to be translated

### LINKED ISSUE #414: Allow translate Preview gutter labels
Currently the Preview gutter labels are hard coded in the CSS file here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/main/assets/css/settings-styles.css#L801

Would be good to allow to translate them.

alexmigf:
closes #414


Terminator-Barbapapa (approved):
💯

PR #477, opened 2023-02-13
New: filter to change role capabilities to access plugin settings

### LINKED ISSUE #476: Allow filtering role capabilities to access plugin settings
Currently we only provide access to the plugin settings to roles with `manage_woocommerce` capabilities, but would be good to be possible to filter those capabilities.

alexmigf:
closes #476

### DIFF
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -83,7 +83,7 @@ class Settings {
 			$parent_slug,
 			esc_html__( 'PDF Invoices', 'woocommerce-pdf-invoices-packing-slips' ),
 			esc_html__( 'PDF Invoices', 'woocommerce-pdf-invoices-packing-slips' ),
-			'manage_woocommerce',
+			$this->settings_capabilities(),
 			'wpo_wcpdf_options_page',
 			array( $this, 'settings_page' )
 		);
@@ -117,21 +117,19 @@ class Settings {
 	}
 	
 	/**
+	 * Get current settings user role capabilities.
+	 * @return string
+	 */
+	public function settings_capabilities() {
+		return apply_filters( 'wpo_wcpdf_settings_user_role_capabilities', 'manage_woocommerce' );
+	}
+	
+	/**
 	 * Check if user role can manage settings.
 	 * @return bool
 	 */
-	public function user_can_manage_settings() {
-		$capabilities = apply_filters( 'wpo_wcpdf_settings_user_role_capabilities', [ 'manage_woocommerce' ] );
-		$permission   = false;
-		
-		foreach ( $capabilities as $capability ) {
-			$permission = current_user_can( $capability );
-			if ( $permission ) {
-				break;
-			}
-		}
-		
-		return $permission;
+	public function user_can_manage_settings() {		
+		return current_user_can( $this->settings_capabilities() );
 	}
 
 	function check_auto_increment_increment() {
@@ -439,13 +437,6 @@ class Settings {
 		register_setting( $option_group, $option_name, array( $this->callbacks, 'validate' ) );
 		add_filter( 'option_page_capability_'.$page, array( $this, 'settings_capabilities' ) );
 
-	}
-
-	/**
-	 * Set capability for settings page
-	 */
-	public function settings_capabilities() {
-		return 'manage_woocommerce';
 	}
 
 	public function get_common_document_settings() {
--- a/includes/class-wcpdf-settings.php
+++ b/includes/class-wcpdf-settings.php
@@ -48,7 +48,7 @@ class Settings {
 		add_filter( 'plugin_row_meta', array( $this, 'add_support_links' ), 10, 2 );
 
 		// settings capabilities
-		add_filter( 'option_page_capability_wpo_wcpdf_general_settings', array( $this, 'settings_capabilities' ) );
+		add_filter( 'option_page_capability_wpo_wcpdf_general_settings', array( $this, 'user_settings_capability' ) );
 
 		// admin notice for auto_increment_increment
 		// add_action( 'admin_notices', array( $this, 'check_auto_increment_increment') );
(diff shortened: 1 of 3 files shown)


Terminator-Barbapapa:
At the moment this only affects a small part of the functionality (for instance viewing the preview). But using this new filter does not allow users with other capabilities than `manage_woocommerce` to view and edit settings.

alexmigf:
@Terminator-Barbapapa fixed

Terminator-Barbapapa (changes requested):
This will let you change the required user capability to view and edit settings, but it won't allow you to add capabilities that are allowed to view and edit settings.

For instance if I want to allow a vendor that has the capability to `manage_product` to edit settings I can't (without an administrator losing his ability to edit settings).

alexmigf:
> For instance if I want to allow a vendor that has the capability to manage_product to edit settings I can't (without an administrator losing his ability to edit settings).

You can, just load the current user using `wp_get_current_user()` inside the filter and do the required check for it, the others can keep the default capability.

alexmigf:
Since `add_submenu_page` accepts only [one capability](https://developer.wordpress.org/reference/functions/add_submenu_page/), we are a bit limited here.

Terminator-Barbapapa:
Couldn't we rewrite `user_can_manage_settings()` like this:

```
public function user_can_manage_settings() {
	$user_can_manage_settings = false;	
	$capabilities = apply_filters( 'wpo_wcpdf_settings_user_role_capabilities', array(
		'manage_woocommerce',
	) );
	foreach ( $capabilities as $capability ) {
		if ( current_user_can( $capability ) ) {
			$user_can_manage_settings = $capability;
			break;
		}
	}
	return $user_can_manage_settings;
}
```

So it either returns a capability or `false`. We can then call it in the `menu()` function like this:

```
public function menu() {
	$parent_slug = 'woocommerce';
	
	$this->options_page_hook = add_submenu_page(
		$parent_slug,
		esc_html__( 'PDF Invoices', 'woocommerce-pdf-invoices-packing-slips' ),
		esc_html__( 'PDF Invoices', 'woocommerce-pdf-invoices-packing-slips' ),
		$this->user_can_manage_settings(),
		'wpo_wcpdf_options_page',
		array( $this, 'settings_page' )
	);
}
```

That way the filter gets simpler to user. You just set the capabilities you want to allow to edit settings. Like this:

```
add_filter( 'wpo_wcpdf_settings_user_role_capabilities', function( $capabilities ) {
	$capabilities = array( 'manage_woocommerce', 'manage_product' );
	return $capabilities;
}, 10, 1 );
```

alexmigf:
@Terminator-Barbapapa I believe you can't have a `bool` value on `add_submenu_page()`, but you gave me an idea.

Terminator-Barbapapa:
The above code worked for me without issues. But maybe I missed something.

alexmigf:
I know it works, but I'm not sure we gain much using an array:

* You can only pass one capability to `add_submenu_page()` and using your suggestion we can miss the correct one;
* If you're using `false` in `user_can_manage_settings()` if the user hasn't any of the passed capabilities you're passing `false` to `add_submenu_page()` when it expects a `string`.

I think we should keep using a single `string` value as before.

Terminator-Barbapapa:
I agree that returning `false` might not be the best thing to do (although I think it doesn't causes issues when calling `add_submenu_page()`)

But I think we can simplify it to the following:

```
public function user_settings_capability() {
	$user_capability       = 'manage_woocommerce';
	$capabilities_to_check = apply_filters( 'wpo_wcpdf_settings_user_role_capabilities', array( $user_capability ) );

	foreach ( $capabilities_to_check as $capability ) {
		if ( current_user_can( $capability ) ) {
			$user_capability = $capability;
			break;
		}
	}
	return $user_capability;
}
```

As I don't think we need a filter for the default capability. The `manage_woocommerce` capability should always allow viewing/editing of the settings.

alexmigf:
I agree, but I still think we don't need an array and the previous code was better.

Terminator-Barbapapa:
Which array do you mean? The `$user_capability` variable will always return a string. And the `$capabilities_to_check` variable also contained an array in your previous code, where it was called `$custom_capabilities`.

alexmigf:
I reverted my last commit from today, check the previous code.

Terminator-Barbapapa:
I see. But that would make the filter less intuitive to use use again.

alexmigf:
Can you explain?

Terminator-Barbapapa:
I think ideally you want to have a filter were you just add the capabilities that can view/edit settings. So just something like this:

```
add_filter( 'wpo_wcpdf_settings_user_role_capabilities', function( $caps ) {
	return array( 'manage_shop_order_terms', 'manage_product' );
}, 10, 1 ); 
```

alexmigf:
Ultimately you just need one capability to be valid under `current_user_can`, once that happens settings should be accessible. If it's not working with `manage_woocommerce` you just need to filter and return a different one, because in the end you just need one.

The array check [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/477#issuecomment-1436638436) just returns one value, you could have 20 capabilities but if the first is valid doesn't matter how many you have, that's why I think just validating one from the beginning should be enough and turns the code more clear and simple. Was my fault because initially I was thinking using an array would be good, but after your first review I saw that I was wrong, and maybe that's why Ewout was using just like that as well. I think we don't need to complicate this, we could also use the [filter](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/477#issuecomment-1431234848) to create specific user role scenarios.

Terminator-Barbapapa:
I understand that only one capability has to be valid to view/edit the settings. And I think your initial code works great. I just think it makes the filter more difficult to use. For instance let's say I want to provide shop managers and vendors access to the settings.

With the current setup the filter would look like this:

```
add_filter( 'wpo_wcpdf_settings_user_role_capabilities', function( $capabilities ) {
	if ( in_array( 'shop_manager', (array) $user->roles ) ) {
		$capabilities = 'manage_shop_order_terms';
	} elseif ( in_array( 'dc_vendor', (array) $user->roles ) ) {
		$capabilities = 'manage_product';
	}
	return $capabilities;
}, 10, 1 );
```

With the proposed setup it would look like this:

```
add_filter( 'wpo_wcpdf_settings_user_role_capabilities', function( $capabilities ) {
	return array( 'manage_shop_order_terms', 'manage_product' );
}, 10, 1 ); 
```

alexmigf:
Done, please leave your review.

Terminator-Barbapapa (approved):
🚀

PR #478, opened 2023-02-15
Fix: Save document settings in order meta only on doc init

### LINKED ISSUE #261: Document settings are being saved to order meta on document object load
Currently document settings, eg. `_wcpdf_invoice_settings`, are being saved to order meta on document object load and not exactly when the document is generated.

The affecting row: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9d1ba26ee7518c87113ea43e3673086bbcd29494/includes/documents/abstract-wcpdf-order-document.php#L154

Which is triggered here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9d1ba26ee7518c87113ea43e3673086bbcd29494/includes/documents/abstract-wcpdf-order-document.php#L118

Maybe we should save only when the document is generated, probably in `init()` somehow.

alexmigf:
closes #261

Currently when loading the document object the settings are instantly saved to the order meta, even when the document doesn't exist yet. This PR tries to fix this by saving the document settings only on document generation.


Terminator-Barbapapa (approved):
LGTM

YordanSoares (approved):
Tested with orders placed in the checkout and the admin, and LGTM!

PR #482, opened 2023-02-22
New: adds notice when RTL is detected

### LINKED ISSUE #480: Display a notice about the existence of the mPDF extension when the site language is RTL
The default PDF engine we use is not compatible with RTL, but we have an extension that switch the PDF engine with mPDF that it is compatible with RTL, but also includes a font compatible with Arabic characters.

[This user](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/479#issuecomment-1437124865) suggested displaying a notice about the existence of the mPDF extension when the site language is Arabic, but would add an `OR` extra check to also display the notice if the language is from the RTL script language family.

alexmigf:
closes #480

### DIFF
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -290,3 +290,16 @@ function wcpdf_catch_db_object_errors( $wpdb ) {
 
 	return $errors;
 }
+
+/**
+ * Determine the site locale
+ */
+function wcpdf_determine_locale() {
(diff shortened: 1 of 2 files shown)


YordanSoares (approved):
## Tests performed:

I tested this PR in both a single and a multisite installations and the message is correctly displayed, if the locale is an RTL language. See screenshots below.

### Single site:

image

### Multisite:

image

PR #487, opened 2023-03-17
Fix: invoice number/date screen options bug with HPOS enabled

### LINKED ISSUE #486: Wrong screen options on order edit page with HPOS enabled
The screen options on the order edit page now display the 'Invoice number' and 'Invoice date' checkboxes, which should only show on the order list page.

alexmigf:
closes #486

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -3,6 +3,7 @@
 namespace WPO\WC\PDF_Invoices;
 
 use \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController;
+use Automattic\WooCommerce\Utilities\OrderUtil;
 
 if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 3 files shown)


includes/class-wcpdf-admin.php
```diff
@@ -236,6 +237,10 @@ public function add_listing_actions( $order ) {
 	 * @param array $columns shop order columns
 	 */
 	public function add_invoice_columns( $columns ) {
+		if ( OrderUtil::custom_orders_table_usage_is_enabled() && isset( $_REQUEST['page'] ) && $_REQUEST['page'] == 'wc-orders' && isset( $_REQUEST['action'] ) && $_REQUEST['action'] == 'edit' ) {
```
Terminator-Barbapapa, line 240: Will also need the compatibility class for users that run an older WooCommerce version.

Terminator-Barbapapa (changes requested):

PR #489, opened 2023-03-24
Save invoice display date option

aparnascodex:
This PR is part of  [Add 'Display invoice date' option ](https://github.com/wpovernight/woocommerce-pdf-ips-number-tools/issues/16) issue.


includes/documents/class-wcpdf-invoice.php
```diff
@@ -58,8 +58,11 @@ public function init() {
 
 		if ( isset( $this->settings['display_date'] ) && $this->settings['display_date'] == 'order_date' && !empty( $this->order ) ) {
 			$this->set_date( $this->order->get_date_created() );
+			$this->set_display_date( 'Order Date' );
```
alexmigf, line 61: Maybe this string should be `order_date`? Then we could also have a getter method to retrieve the date label. We could also use that method to display the date options here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/87b6ee11bde97705770300c3de724faf5ff03105/includes/documents/class-wcpdf-invoice.php#L267-L268

alexmigf (changes requested):
I think we could display the date label in the Document data table:

Screenshot from 2023-03-28 14-30-48

includes/documents/class-wcpdf-invoice.php
aparnascodex: Implemented suggested changes.

Terminator-Barbapapa:
The value isn't updated when regenerating the document:

Schermafbeelding 2023-04-03 om 13 12 02

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1184,7 +1184,38 @@ public function document_notes() {
 		}
 	}
 
+	
+	public function document_display_date() {
```
alexmigf, line 1188: Shouldn't this function code be inside the `get_display_date` function? Seems redundant to me having two similar functions.

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document-methods.php
aparnascodex: I followed same convention as it is used for other fields like notes and invoice number. I can move the code inside 'get_display_date'

alexmigf: I see, in that case let's follow that then.

aparnascodex:
@Terminator-Barbapapa I have added code to update the value on document regeneration.

alexmigf (approved):

Terminator-Barbapapa (approved):

PR #496, opened 2023-04-06
Add upgrade tab

### LINKED ISSUE #488: Add extensions tab
We should include an extra tab that gives a good overview of what our paid PDF Invoices extensions offer. With a direct link to upgrade. We can also use that tab to give a brief overview of other interesting plugin we sell.

Terminator-Barbapapa:
This adds an 'Upgrade' tab to the settings. So users can get a quick overview of what our extensions offer them.

Schermafbeelding 2023-04-06 om 11 52 47

Closes #488


includes/class-wcpdf-settings-upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+		if ( $tab === 'upgrade' ) {
+			
+			$features = array(
+				array( 
+					'label' => __( 'Proforma invoice & Credit note', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description' => __( 'Update your workflow and handle refunds.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'extensions' => ['pro', 'bundle'],
+				),
+				array( 
+					'label' => __( 'Dropbox / FTP upload', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 26: I would call it `Cloud storage upload`, because we plan to also add a few more like GDrive in the future. Maybe you could add the Dropbox and FTP in the description below? I also would mention **SFTP**.

includes/views/upgrade-table.php
alexmigf, line 1: Wouldn't be better to have the styles in a standalone file?

includes/views/upgrade-table.php
```diff
@@ -0,0 +1,240 @@
...
+		);
+		printf(
+			'<td><a href="%s" target="_blank">%s</a></td>', 
+			'https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/', 
+			__( 'Upgrade now', 'woocommerce-pdf-invoices-packing-slips' ) 
+		);
+		?>
+		<td align="left" class="last">&nbsp;</td>
+	</tr>
+
```
alexmigf, line 238: Maybe you could remove this 2 empty lines?

alexmigf (changes requested):
Maybe you could also include this?

## Pro

* Separate numbering systems
* Customize the shipping & billing address format

## Premium Templates

* Add custom CSS styles easily

includes/class-wcpdf-settings-upgrade.php
Terminator-Barbapapa: Yes, great suggestions. I will add those and change the cloud storage label.

includes/views/upgrade-table.php
Terminator-Barbapapa: Yes, these should move to the settings css file. Only the top style needs to remain there as I think there is no good way of hiding the default settings form by checking for a specific class.

Terminator-Barbapapa:
@alexmigf I would also like to add a personalized upgrade link to the bundle when the user already has one of the extensions installed. Do you think that is possible?

alexmigf:
I believe it should, using an API request to our store and maybe caching that in a transient, to prevent multiple requests. Do you want me to handle that?

Btw, great work, looks great!

includes/class-wcpdf-settings-upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+					'extensions' => ['pro', 'bundle'],
+				),
+				array( 
+					'label' => __( 'Bulk export', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description' => __( 'Easily export documents for a specific date range.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'extensions' => ['pro', 'bundle'],
+				),
+				array( 
+					'label' => __( 'Multilingual support', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description' => __( 'Handle document translations via either WPML or Polylang.', 'woocommerce-pdf-invoices-packing-slips' ),
```
YordanSoares, line 37: Maybe we could add this: "Note: It has limited compatibility with other multilanguage plugins."

includes/views/upgrade-table.php
```diff
@@ -0,0 +1,240 @@
...
+		<span><?php printf( '%s: %s', esc_html_e( 'If you have any questions feel free to send us an email at', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="mailto:support@wpovernight.com">support@wpovernight.com</a>' ); ?><span>
+	</p>
+</div>
+
+<table id="upgrade-table">
+	<tr>
+		<th class="first" align="left">&nbsp;</th>
+		<th align="left"><?php esc_html_e( 'Professional', 'woocommerce-pdf-invoices-packing-slips' ); ?></th>
+		<th align="left"><?php esc_html_e( 'Premium Templates', 'woocommerce-pdf-invoices-packing-slips' ); ?></th>
+		<th align="left"><?php esc_html_e( 'PDF Invoices Bundle', 'woocommerce-pdf-invoices-packing-slips' ); ?></th>
```
YordanSoares, line 203: I would use only "Bundle" here.

includes/class-wcpdf-settings-upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+class Settings_Upgrade {
+
+	function __construct()	{
+		add_action( 'wpo_wcpdf_after_settings_page', array( $this, 'extension_overview' ), 10, 2 );
+	}
+
+	public function extension_overview( $tab, $section ) {
+		if ( $tab === 'upgrade' ) {
+			
+			$features = array(
```
YordanSoares, line 19: We could mention the extra settings for the packing slips, like the attachment option.

Terminator-Barbapapa:
> I believe it should, using an API request to our store and maybe caching that in a transient, to prevent multiple requests. Do you want me to handle that?
> 
> 
> 
> Btw, great work, looks great!

Thanks! Yes can you create a separate issue for that? It should also hide the Pro and Templates upgrade buttons if one of those extensions is already installed.

includes/class-wcpdf-settings-upgrade.php
Terminator-Barbapapa: I think if we do not provide full compatibility we should not advertise it.

includes/views/upgrade-table.php
Terminator-Barbapapa: That might indeed be cleaner.

includes/class-wcpdf-settings-upgrade.php
Terminator-Barbapapa: Good idea.

YordanSoares (approved):
✔️ LGTM!

alexmigf (changes requested):
What you think of changing the menu color to the Purple? To have more visibility.

Screenshot from 2023-04-12 12-40-18

assets/css/settings-styles.css
```diff
@@ -270,7 +270,7 @@ body.woocommerce_page_wpo_wcpdf_options_page {
 }
 
 .nav-tab-wrapper a.nav-tab.nav-tab-active {
-	border-bottom: 3px solid #51266b;
+	border-bottom: 3px solid #6e1edc;
```
alexmigf, line 273: Sorry, I meant to say the text color for the **Upgrade** tab, like this:

Screenshot from 2023-04-12 13-32-07

alexmigf (changes requested):

alexmigf (approved):

PR #497, opened 2023-04-06
Save document creation status

aparnascodex:
This is part of https://github.com/wpovernight/woocommerce-pdf-ips-number-tools/issues/17

Following are the status of document:
1. When invoice is generated using 'Pdf invoice' button located inside 'Create PDF' metabox. Document status will be Manual (invoice was not generated before clicking button)
2. When invoice is saved or regenerated, invoice status will be Manual, irrespective of whether the invoice was previously generated or not.
3. If invoice is generated when email is sent [eg. new order or processing order] status will be Automatic.
4. If invoice is generated manually using 1st or 2nd point and then sent in any email attachment, invoice status will be Manual (as it was already generated before)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -5,7 +5,7 @@ Tags: woocommerce, pdf, invoices, packing slips, print, delivery notes, invoice,
 Requires at least: 3.5
 Tested up to: 6.2
 Requires PHP: 7.1
-Stable tag: 3.5.1
+Stable tag: 3.5.2
 License: GPLv2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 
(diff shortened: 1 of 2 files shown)


alexmigf (changes requested):
I think you could use this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9758c9188aa0ac2fc1c59c2ee55f4760e545eee2/includes/class-wcpdf-main.php#L1206-L1214

And do similar to this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9758c9188aa0ac2fc1c59c2ee55f4760e545eee2/includes/class-wcpdf-main.php#L382

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/9758c9188aa0ac2fc1c59c2ee55f4760e545eee2/includes/class-wcpdf-main.php#L1221-L1237

aparnascodex:
@alexmigf I have modified the code. Please check code of includes/class-wcpdf-main.php file. Let me know if any further changes are needed.

PR #500, opened 2023-04-14
display details if document was automatically or manually created

aparnascodex:
closes https://github.com/wpovernight/woocommerce-pdf-ips-number-tools/issues/17

I had to reset the branch due to multiple conflicts. Please review the changes now.


alexmigf:
@aparnascodex did you saw my comments [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/497)?

aparnascodex:
@alexmigf I have modified the code. Please check code of includes/class-wcpdf-main.php file. Let me know if any further changes are needed.

alexmigf:
@aparnascodex hope you don't mind, but I have made a few changes. Can you please review and let me know your opinion?

aparnascodex:
@alexmigf I went through the code changes, it looks good.

includes/class-wcpdf-main.php
```diff
@@ -1208,6 +1222,35 @@ public function log_to_order_notes( $note, $document ) {
...
+	/**
+	 * Logs to the order meta
+	 *
+	 * @param object  $document
+	 * @param string  $trigger
+	 * @param boolean $force
+	 * @return void
+	 */
+	public function log_document_creation_status_to_order_meta( $document, $trigger, $force = false ) {
+		if ( $trigger == 'bulk' && property_exists( $document, 'order_ids' ) && ! empty( $document->order_ids ) ) { // bulk document
```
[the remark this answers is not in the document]
Terminator-Barbapapa: I think the trigger should always be saved in the order meta. But displaying it in the order notes should be optional. 

Also I think we might have to change the UI of the 'PDF document data' metabox(es) a bit. So the 'Invoice display date' and 'Invoice created' are hidden by default and can be viewed via something like a 'View details' button.

includes/class-wcpdf-admin.php
```diff
@@ -552,20 +555,29 @@ public function get_current_values_for_document( $document, $data ) {
...
 			$current['display_date'] = array(
 				'value' => $document->document_display_date(),
-				'name'  =>"_wcpdf_{$document->slug}_display_date",
+				'name'  => "_wcpdf_{$document->slug}_display_date",
 			);
 		}
 
+		if ( ! empty( $data['creation_status'] ) ) {
+			$document_triggers = WPO_WCPDF()->main->get_document_triggers();
+			$creation_status   = $document->get_creation_status();
```
Terminator-Barbapapa, line 574: I would rename this to: `$trigger = $document->get_trigger()`

This then also has to be changed in the rest of the code.

includes/class-wcpdf-admin.php
```diff
@@ -515,16 +515,19 @@ public function data_input_box_content( $post_or_order_object ) {
...
 				'date'   => array(
-					'label'  => __( 'Invoice Date:', 'woocommerce-pdf-invoices-packing-slips' ),
+					'label' => __( 'Invoice Date:', 'woocommerce-pdf-invoices-packing-slips' ),
 				),
 				'display_date' =>  array(
-					'label'  => __( 'Display Date:', 'woocommerce-pdf-invoices-packing-slips' ),
+					'label' => __( 'Invoice Display Date:', 'woocommerce-pdf-invoices-packing-slips' ),
+				),
+				'creation_status' =>  array(
+					'label' => __( 'Invoice Created:', 'woocommerce-pdf-invoices-packing-slips' ),
```
Terminator-Barbapapa, line 527: I would rename this to 'Invoice Created Via'. Also perhaps we should drop the title case for these labels? Especially since we now have labels with more than two words, I think:

```
Invoice number:
Invoice date:
Invoice display date:
Invoice created via:
``` 

looks better than:

```
Invoice Number:
Invoice Date:
Invoice Display Date:
Invoice Created Via:
```

What do you think @alexmigf, @YordanSoares, @aparnascodex?

Terminator-Barbapapa (changes requested):

includes/class-wcpdf-admin.php
alexmigf: I agree.

includes/class-wcpdf-main.php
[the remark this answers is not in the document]
alexmigf: The code does exactly that, so let's keep things as they are for now. @Terminator-Barbapapa can you do a mockup of the "View details" so we could address that in a new issue?

includes/class-wcpdf-admin.php
alexmigf: If you look at the document functions here https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/d6c5ae258ec0551845f0f6f104f96c26ef593566/includes/documents/abstract-wcpdf-order-document.php#L459-L477

They are more descriptive in their names, and since we are labeling this as `Invoice created via:`, I don't think we should name this `trigger`. But let me know.

includes/class-wcpdf-main.php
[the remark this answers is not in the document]
Terminator-Barbapapa: Yes, I was thinking something simple. Like this:

Schermafbeelding 2023-05-11 om 11 03 29

includes/class-wcpdf-admin.php
YordanSoares: I'm also agreed.

includes/class-wcpdf-admin.php
Terminator-Barbapapa: But we also call them triggers when getting a list of all possible triggers for creating the document:

```
$document_triggers = WPO_WCPDF()->main->get_document_triggers();
```

Would you prefer it if we called it something like `creation_trigger`?

alexmigf: `creation_trigger` seems more appropriate indeed.

includes/class-wcpdf-admin.php
alexmigf: What about just?

```
Number:
Date:
Display Date:
Created Via:
```

Terminator-Barbapapa: I like that visually, but I wonder if there is a downside from an accessibility point of view. For instance people that are vision impaired and use a screen reader. Perhaps we can leave it for now.

Terminator-Barbapapa (approved):

PR #503, opened 2023-04-21
New: detects the extension license status in the Upgrade tab

### LINKED ISSUE #502: Check if user has extension installed and reflect that in the Upgrade tab
We could once a while check if the user has extensions installed, and if yes check if the license is valid and reflect that in the Upgrade tab.

alexmigf:
closes #502 

We are passing the **Bundle Upgrade link** using the snippet below in the WPO website:

```php
add_filter( 'edd_remote_license_check_response', function( $response, $args, $license_id ) {
	if ( ! empty( $license_id ) && function_exists( 'edd_software_licensing' ) ) {
		$license = edd_software_licensing()->get_license( $license_id );
		if ( $license && function_exists( 'edd_sl_get_license_upgrades' ) ) {
			$upgrades = edd_sl_get_license_upgrades( $license_id );
			if ( $upgrades && ! $license->parent ) {
				$bundle_upgrades = [];
				foreach ( $upgrades as $upgrade_id => $upgrade ) {
					$download = new \EDD_SL_Download( $upgrade['download_id'] );
					if ( $download && ! empty( $download->post_title ) && $download->post_title == 'WooCommerce PDF Invoices and Packing Slips Bundle' ) {
						$bundle_upgrades[] = esc_url_raw( edd_sl_get_license_upgrade_url( $license_id, $upgrade_id ) );
					}
				}
				
				if ( count( $bundle_upgrades ) == 3 ) {
					$activation_limits = [ 1, 3, 25 ];
					foreach ( $activation_limits as $key => $limit ) {
						if ( $response['license_limit'] == $limit ) {
							$response['bundle_upgrade'] = $bundle_upgrades[$key];
						}
					}
				}
			} elseif ( $license->parent ) {
				$response['bundle_upgrade'] = 'is_bundled'; // license is from a bundled extension
			}
		}
	}
	return $response;
}, 10, 3 );
```


Terminator-Barbapapa:
Great! Couple of thoughts:

- The Upgrade tab should be about upgrading, and making that as easy as possible. So although showcasing some license information can be useful, it should be secondary. I would just check if which of our extensions is installed and show upgrade buttons accordingly. So when Pro or Premium Templates is active only the upgrade button for the Bundle has to be shown.
- As mentioned I think license info should be secondary. So I would prefer to just say 'Currently installed' when an extension is currently installed and active. We can also perform a license activation status check here, but show that as a subtitle when the license is not activated yet, with a link to our documentation page on how to activate your license. (Documentation page will have to be updated). So that will give you the following options:
Schermafbeelding 2023-04-21 om 13 49 35

- If a license key for one of the extensions is found, can we make a call to EDD and use the appropriate personal upgrade link for the Bundle's upgrade button? The same links that are available in HelpScout. So users can directly upgrade their current license to the Bundle.

alexmigf:
I'm not using the API, just using the Updater, which I think it's enough, but we have one problem, I can't detect if the user has a Bundle license or not because is not possible to activated in the customers store.

alexmigf:
@Terminator-Barbapapa can you review again please?

alexmigf:
We are passing the **Lowest Bundle Upgrade link** using the snippet below in the WPO website:

```php
add_filter( 'edd_remote_license_check_response', function( $response, $args, $license_id ) {
	if ( ! empty( $license_id ) && function_exists( 'edd_sl_get_license_upgrades' ) ) {
		$upgrades                   = edd_sl_get_license_upgrades( $license_id );
		$response['lowest_bundle_upgrade'] = '';
		foreach ( $upgrades as $upgrade_id => $upgrade ) {
			$download = new \EDD_SL_Download( $upgrade['download_id'] );
			if ( $download && ! empty( $download->post_title ) && $download->post_title == 'WooCommerce PDF Invoices and Packing Slips Bundle' ) {
				$response['lowest_bundle_upgrade'] = esc_url_raw( edd_sl_get_license_upgrade_url( $license_id, $upgrade_id ) ); // lowest is always the first
				break;
			}
		}
	}
	return $response;
}, 10, 3 );
```

includes/views/upgrade-table.php
```diff
@@ -38,21 +38,68 @@
...
+					$title = __( 'Currently installed', 'woocommerce-pdf-invoices-packing-slips' );
+					if ( empty( $info['status'] ) || $info['status'] != 'valid' ) {
+						$subtitle = sprintf(
+							/* translators: learn more link */
+							__( 'License not yet activated: %s', 'woocommerce-pdf-invoices-packing-slips' ),
+							'<a href="https://docs.wpovernight.com/general/installing-wp-overnight-plugins/" target="_blank">'.__( 'Learn more', 'woocommerce-pdf-invoices-packing-slips' ).'</a>'
+						);
+					} else {
+						$subtitle = __( 'License activated!', 'woocommerce-pdf-invoices-packing-slips' );
+					}
```
Terminator-Barbapapa, line 60: As license info is secondary on this tab I don't think we need to display this. Only when the license is not activated we can give a little nudge for the user to activate it.

includes/class-wcpdf-settings-upgrade.php
```diff
@@ -98,11 +98,90 @@ public function extension_overview( $tab, $section ) {
...
+	 * @param  string  $extension  can be 'pro' or 'templates'
+	 * @return array
+	 */
+	public function get_extension_license_info( $extension ) {
+		$license_info = [
+			'enabled' => false
+		];
+		
+		if ( $this->extension_is_enabled( $extension ) ) {
+			$license_info['enabled'] = true;
```
Terminator-Barbapapa, line 148: I don't think we should save this in the transient as it should represent the value at the time of loading the page. 

Also later in the `upgrade-table.php` view we only want to check if the extension is installed. It doesn't even has to be active. If we find either Pro or Templates is installed we can assume they have bought the extension and therefore have a license. So we can show 'Currently installed' for that extension, and display the 'Upgrade now' button for the Bundle.

assets/css/settings-styles.css
```diff
@@ -1011,7 +1011,16 @@ div.upgrade-table-description a,
...
 }
 
-#upgrade-table .upgrade-links a {
+#upgrade-table .upgrade-links h4 {
+	margin: 1em 0 0.5em 0;
+}
+
+#upgrade-table .upgrade-links p {
+	margin: 0;
+	font-style: oblique;
```
Terminator-Barbapapa, line 1020: We can set the `font-size` inline with the other subtitles to `0.8em`;

includes/class-wcpdf-settings-upgrade.php
```diff
@@ -98,11 +98,90 @@ public function extension_overview( $tab, $section ) {
...
+						'license_key' => trim( $license_key ),
+					);
+				} else {
+					$args = array();
+				}
+				
+				$request = $extension_main_function()->updater->remote_license_actions( $args );
+				if ( is_object( $request ) && isset( $request->license ) ) {
+					$license_info['status']         = $request->license;
+					$license_info['bundle_upgrade'] = ! empty( $request->lowest_bundle_upgrade ) && is_string( $request->lowest_bundle_upgrade ) ? $request->lowest_bundle_upgrade : ''; // https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/503#issuecomment-1523066696
```
Terminator-Barbapapa, line 168: This might cause an issue with users that have a 3-site or 25-site license for one of the extensions. Because I think it will return a 1-site Bundle upgrade link.

Terminator-Barbapapa (changes requested):

includes/class-wcpdf-settings-upgrade.php
alexmigf: If it's only active I don't have a way to get the upgrade link because I don't know the license key.

Terminator-Barbapapa (changes requested):

PR #505, opened 2023-04-28
New: adds filter to stick Document data metabox

### LINKED ISSUE #504: PDF Document data metabox is fixed permanent after the order details
Currently the PDF Document data metabox is permanently fixed in the same position after the Order details, would be good to have a filter to allow to be re-positioned.

alexmigf:
closes #504

Waiting customer feedback :hourglass:


includes/class-wcpdf-admin.php
```diff
@@ -347,12 +347,16 @@ public function request_query_sort_by_column( $query_vars ) {
 	/**
 	 * Add the meta boxes on the single order page
 	 */
-	public function add_meta_boxes() {
+	public function add_meta_boxes( $wc_screen, $wc_order ) {
```
aparnascodex, line 350: I guess we can skip $wc_order variable as it is not used in function.

alexmigf: I think we should keep for future reference, if we need to use it later, but I will add instructions to the comment above.

aparnascodex (approved):

PR #507, opened 2023-05-02
Fix: adds semaphore to Invoice number init to prevent concurrent number issues

### LINKED ISSUE #506: Prevent that two PHP processes generates a document number (race condition)
Some users have been having issues with race conditions related with the payment method. The most common issue is with PayPal, when you have both "PDT" and "IPN" methods activated, two different methods for confirming payment.

Although this is not an issue related with our plugin itself, we could try to avoid this behavior from our end, maybe locking the document number if has been already created.

alexmigf:
closes #506

### DIFF
--- a/includes/documents/class-wcpdf-invoice.php
+++ b/includes/documents/class-wcpdf-invoice.php
@@ -1,6 +1,8 @@
 <?php
 namespace WPO\WC\PDF_Invoices\Documents;
 
+use WPO\WC\PDF_Invoices\Updraft_Semaphore_3_0 as Semaphore;
+
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly
 }
@@ -17,6 +19,15 @@ if ( !class_exists( '\\WPO\\WC\\PDF_Invoices\\Documents\\Invoice' ) ) :
  */
 
 class Invoice extends Order_Document_Methods {
+	
+	public $type;
+	public $title;
+	public $icon;
+	public $lock_name;
+	public $lock_context;
+	public $lock_time;
+	public $lock_retries;
+	
 	/**
 	 * Init/load the order object.
 	 *
@@ -24,9 +35,15 @@ class Invoice extends Order_Document_Methods {
 	 */
 	public function __construct( $order = 0 ) {
 		// set properties
-		$this->type		= 'invoice';
-		$this->title	= __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
-		$this->icon		= WPO_WCPDF()->plugin_url() . "/assets/images/invoice.svg";
+		$this->type		    = 'invoice';
+		$this->title	    = __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
+		$this->icon		    = WPO_WCPDF()->plugin_url() . "/assets/images/invoice.svg";
+		
+		// semaphore
+		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
+		$this->lock_context = array( 'source' => "wpo-wcpdf-{$this->type}-semaphore" );
+		$this->lock_time    = apply_filters( 'wpo_wcpdf_semaphore_lock_time', 300 );
+		$this->lock_retries = apply_filters( 'wpo_wcpdf_semaphore_lock_retries', 0 );
 		
 		// Call parent constructor
 		parent::__construct( $order );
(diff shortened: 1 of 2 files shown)


includes/documents/class-wcpdf-invoice.php
```diff
@@ -17,16 +19,31 @@
...
 	 * @param  int|object|WC_Order $order Order to init.
 	 */
 	public function __construct( $order = 0 ) {
 		// set properties
-		$this->type		= 'invoice';
-		$this->title	= __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
-		$this->icon		= WPO_WCPDF()->plugin_url() . "/assets/images/invoice.svg";
+		$this->type		    = 'invoice';
+		$this->title	    = __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
+		$this->icon		    = WPO_WCPDF()->plugin_url() . "/assets/images/invoice.svg";
```
Terminator-Barbapapa, line 40: Spacing is off (I think only spaces need to be used here).

includes/documents/class-wcpdf-invoice.php
```diff
@@ -17,16 +19,31 @@
...
+		
+		// semaphore
+		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
+		$this->lock_context = array( 'source' => "wpo-wcpdf-{$this->type}-semaphore" );
+		$this->lock_time    = apply_filters( 'wpo_wcpdf_semaphore_lock_time', 300 );
```
Terminator-Barbapapa, line 45: Isn't `300` seconds long for a lock? If anything goes wrong in the invoice number creation no invoice number can be created for 5 minutes.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -75,34 +92,49 @@ public function exists() {
...
-				$this->set_number( $invoice_number );
-			} else {
-				// invoice number is not numeric, treat as formatted
-				// try to extract meaningful number data
-				$formatted_number = $invoice_number;
-				$number = (int) preg_replace('/\D/', '', $invoice_number);
-				$invoice_number = compact( 'number', 'formatted_number' );
-				$this->set_number( $invoice_number );
+				if ( ! empty( $invoice_number ) ) { // overriden by plugin or set to order number
+					if ( ! is_numeric( $invoice_number ) || ! $invoice_number instanceof Document_Number ) {
```
Terminator-Barbapapa, line 110: I think the way this is structured now might cause issues when an a `Document_Number` object is passed here.

Terminator-Barbapapa (changes requested):

includes/documents/class-wcpdf-invoice.php
alexmigf: Which number do you suggest instead?

includes/documents/class-wcpdf-invoice.php
alexmigf: I don't think that's a problem because if you look at the old code structure it just required this: `$this->set_number( $invoice_number );` which is done only line `123`.

includes/documents/class-wcpdf-invoice.php
Terminator-Barbapapa: I think because this lock is only meant to better handle a race condition maybe one second would be enough? What are your thoughts on this @DavidAnderson684? I'm not too familiar with the intricacies of the Semaphore class.

DavidAnderson684: Yes, the maximum length of the lock should be related to how long you reasonably expect something to take in the code that is going to run whilst the lock is being held.
i.e. You're saying "let's wait up to (this) amount of time, and after that, we don't expect the code in the other thread that locked this to still be running, and so after that we will try again". So, you don't want to have a time interval that is implausibly long because then you're just wasting time and as Michael says, if it's a lock that other processes may want to have, you're going to hold them all up.
The default is 300 because its initial use was for backups in UpdraftPlus - those may well run for 300 seconds on a large site, and the only thing that's going to call for the lock is a new backup process, and holding those up is the right thing to do. But that's different here.

includes/documents/class-wcpdf-invoice.php
Terminator-Barbapapa: The old code checked the following before moving to `set_number()`:

```
if ( !empty( $invoice_number ) ) { // overriden by plugin or set to order number
	if ( is_numeric($invoice_number) || $invoice_number instanceof Document_Number ) {
		$this->set_number( $invoice_number );
	} else {
		// invoice number is not numeric, treat as formatted
		// try to extract meaningful number data
		$formatted_number = $invoice_number;
		$number = (int) preg_replace('/\D/', '', $invoice_number);
		$invoice_number = compact( 'number', 'formatted_number' );
		$this->set_number( $invoice_number );
	}
	return $invoice_number;
}

$number_store   = $this->get_sequential_number_store();
$invoice_number = $number_store->increment( intval( $this->order_id ), $this->get_date()->date_i18n( 'Y-m-d H:i:s' ) );

$this->set_number( $invoice_number );
```

Which is now replaced by:

```
if ( ! empty( $invoice_number ) ) { // overriden by plugin or set to order number
	if ( ! is_numeric( $invoice_number ) || ! $invoice_number instanceof Document_Number ) {
		// invoice number is not numeric, treat as formatted
		// try to extract meaningful number data
		$formatted_number = $invoice_number;
		$number           = (int) preg_replace( '/\D/', '', $invoice_number );
		$invoice_number   = compact( 'number', 'formatted_number' );
	}
} else {
	$number_store   = $this->get_sequential_number_store();
	$invoice_number = $number_store->increment( intval( $this->order_id ), $this->get_date()->date_i18n( 'Y-m-d H:i:s' ) );
}

if ( ! is_null( $invoice_number ) ) {
	$this->set_number( $invoice_number );
}
```

Note that
```
if ( ! is_numeric( $invoice_number ) || ! $invoice_number instanceof Document_Number )
``` 
does not give you the same result as the `else` statement in the old code:

```
if ( is_numeric($invoice_number) || $invoice_number instanceof Document_Number ) {
        //...
} else {
        //...
}
```

For instance if I use the `wpo_wcpdf_external_invoice_number` filter to generate an invoice number with custom prefix/suffix/padding and return an instance of `Document_Number`,  the check in the new code will result in `true` as `$invoice_number` is not numeric but an object. This then stores the whole `Document_Number` object under `formatted_number` in the order meta. And leaving `prefix`, `suffix` and `padding` blank. 

Schermafbeelding 2023-05-05 om 13 39 47

alexmigf: Could you here your code snippet for the `wpo_wcpdf_external_invoice_number` please?

alexmigf: I'm still not getting your point of view, maybe I'm missing something here, but let me explain mine:

- If `$invoice_number` is `Document_Number` object:

## Old code

- It goes through this line because it's not empty: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/7dfe79784913dd1a624dab5c4dc15741903dd4db/includes/documents/class-wcpdf-invoice.php#L87
- It goes through this line because it's an object instance of Document_number: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/7dfe79784913dd1a624dab5c4dc15741903dd4db/includes/documents/class-wcpdf-invoice.php#L88
- Number is `set` here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/7dfe79784913dd1a624dab5c4dc15741903dd4db/includes/documents/class-wcpdf-invoice.php#L89
- Variable is returned after the number is set here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/7dfe79784913dd1a624dab5c4dc15741903dd4db/includes/documents/class-wcpdf-invoice.php#L98
  
## New code

- It goes through this line because it's not empty: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f88ffebef964d4dfd374c980f23267fd8c163308/includes/documents/class-wcpdf-invoice.php#L109
- Doesn't go through he next check because it's `Document_number` instance: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f88ffebef964d4dfd374c980f23267fd8c163308/includes/documents/class-wcpdf-invoice.php#L110
- It goes through this line because it's not null: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f88ffebef964d4dfd374c980f23267fd8c163308/includes/documents/class-wcpdf-invoice.php#L122
- Sets the number: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f88ffebef964d4dfd374c980f23267fd8c163308/includes/documents/class-wcpdf-invoice.php#L123
- Variable is returned after the number is set: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f88ffebef964d4dfd374c980f23267fd8c163308/includes/documents/class-wcpdf-invoice.php#L138

Terminator-Barbapapa: > - Doesn't go through the next check because it's `Document_number` instance: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f88ffebef964d4dfd374c980f23267fd8c163308/includes/documents/class-wcpdf-invoice.php#L110

It does not go through that check because it isn't numeric. That is why I said:

> Note that
```
if ( ! is_numeric( $invoice_number ) || ! $invoice_number instanceof Document_Number ) {
        //...
}
``` 
does not give you the same result as the `else` statement in the old code:

```
if ( is_numeric($invoice_number) || $invoice_number instanceof Document_Number ) {
        //...
} else {
        //...
}
```

alexmigf: You're right, for some reason my head was misleading by that numeric check.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -17,16 +19,31 @@
...
-		$this->icon		= WPO_WCPDF()->plugin_url() . "/assets/images/invoice.svg";
+		$this->type         = 'invoice';
+		$this->title        = __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
+		$this->icon         = WPO_WCPDF()->plugin_url() . "/assets/images/invoice.svg";
+		
+		// semaphore
+		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
+		$this->lock_context = array( 'source' => "wpo-wcpdf-{$this->type}-semaphore" );
+		$this->lock_time    = apply_filters( 'wpo_wcpdf_semaphore_lock_time', 2 );
+		$this->lock_retries = apply_filters( 'wpo_wcpdf_semaphore_lock_retries', 0 );
```
Terminator-Barbapapa, line 46: Perhaps we need to add the document type to these filter names? So if necessary we can change lock time and lock retries independently from the lock time and retries of the yearly invoice number reset.

So something like: 
```
wpo_wcpdf_{$this->slug}_number_lock_time
wpo_wcpdf_{$this->slug}_number_lock_retries
```

Terminator-Barbapapa (changes requested):

Terminator-Barbapapa (approved):
Awesome!

PR #509, opened 2023-05-02
Fix: attached invoice column hooks only on setting condition

### LINKED ISSUE #508: Slow Order List page loading speed when our plugin is installed
At least 2 users reported that the Order List page is slow to load when the orders per page number is big and our plugin is installed. Doing some profiling in a customer site I found that this hook is fired for every order even if the column settings is disabled: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/7dfe79784913dd1a624dab5c4dc15741903dd4db/includes/class-wcpdf-admin.php#L284-L308

alexmigf:
closes #508

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -15,12 +15,14 @@ class Admin {
 	function __construct()	{
 		add_action( 'woocommerce_admin_order_actions_end', array( $this, 'add_listing_actions' ) );
 
-		add_filter( 'manage_woocommerce_page_wc-orders_columns', array( $this, 'add_invoice_columns' ), 999 ); // WC 7.1+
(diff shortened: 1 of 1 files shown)


alexmigf:
The user feedback is positive, good improvement in page speed with this changes.

YordanSoares (approved):
Firstly, the invoice number search seems to be working correctly:

image

Regarding the page generation time, on loading 20 orders (the default number of items) this is about 0.95s, when the columns for the invoice number and invoice date are activated:

image

This time decrease about 0.17s when those settings are disabled:

image

On the other hand, the page generation time takes about 0.65s, when you run a number invoice search and both columns for the invoice number and invoice date are activated:

image

But this time decrease about 0.12s after deactivating the columns:

image

PR #517, opened 2023-05-19
Fix: log invoice number generation on setting

### LINKED ISSUE #516: Disable Semaphore logs by default
Currently the Semaphore lock logs are being created when a document number is generated, maybe we should disable by default and add a new setting to enable this.

> 2023-05-19T10:29:17+00:00 DEBUG Lock option (updraft_lock_wpo_wcpdf__number_lock, wp_options) was created in the database
2023-05-19T10:29:17+00:00 INFO Lock (updraft_lock_wpo_wcpdf__number_lock, wp_options) acquired after initialising the database
2023-05-19T10:29:17+00:00 INFO Lock option (updraft_lock_wpo_wcpdf__number_lock, wp_options) released

alexmigf:
closes #516

### DIFF
--- a/includes/documents/class-wcpdf-invoice.php
+++ b/includes/documents/class-wcpdf-invoice.php
@@ -94,7 +94,8 @@ class Invoice extends Order_Document_Methods {
 	}
 
 	public function init_number() {
-		$lock           = new Semaphore( $this->lock_name, $this->lock_time, array( wc_get_logger() ), $this->lock_context );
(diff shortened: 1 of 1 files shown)


YordanSoares (approved):
Perfect! The link you have added makes easier to check the logs:

image

dwalkerpriv (approved):
I didn't notice this log before. I did 2 tests with the option enabled and with it disabled and it seems to work as expected.

PR #518, opened 2023-05-19
Fix: adds support for legacy Sidekick activated licenses in Upgrade tab

alexmigf:
Currently the Upgrade tab only check license status from the built-in updater, this PR adds support for the legacy Sidekick as well.


Terminator-Barbapapa (approved):
LGTM! 🚀

PR #519, opened 2023-05-30
pdf data ui changes

### LINKED ISSUE #510: Little change to the PDF Document Data UI
@Terminator-Barbapapa suggested this [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/500#discussion_r1190832060):

> Also I think we might have to change the UI of the 'PDF document data' metabox(es) a bit. So the 'Invoice display date' and 'Invoice created' are hidden by default and can be viewed via something like a 'View details' button.

237639837-9e643db5-6b5f-4f13-b9f0-cf639ada8848

aparnascodex:
closes #510


assets/js/order-script.js
```diff
@@ -154,4 +154,17 @@ jQuery( function( $ ) {
 		}
 	}
 
+	jQuery('#wpo_wcpdf-data-input-box').on('click', '.view-more', function(e) {
+		e.preventDefault();
+		jQuery(this).addClass('pdf-hide');
```
alexmigf, line 159: You can use the [`show`](http://api.jquery.com/show/) & [`hide`](http://api.jquery.com/hide/) functions instead of a class.

assets/js/order-script.js
```diff
@@ -154,4 +154,17 @@ jQuery( function( $ ) {
...
+		jQuery('.pdf-more-details').slideDown('slow');
+		jQuery('.hide-details').removeClass('pdf-hide');
+	});
+	jQuery('#wpo_wcpdf-data-input-box').on('click', '.hide-details', function(e) {
```
alexmigf, line 163: I think you can combine both functions, they are very similar.

includes/class-wcpdf-admin.php
```diff
@@ -676,27 +676,38 @@ public function output_number_date_edit_fields( $document, $data ) {
...
 								</p>
-							</p>
+							</div>
+							<?php endif; ?>
 						</div>
-						<?php endif; ?>	
-											
+						<?php if ( isset( $data['display_date'] ) || isset( $data['creation_trigger'] ) ) : ?>
+							<div>
+								<a href='#' class='view-more'>
```
alexmigf, line 703: I think the link element can be in a single line.

includes/class-wcpdf-admin.php
```diff
@@ -676,27 +676,38 @@ public function output_number_date_edit_fields( $document, $data ) {
...
 						</div>
 						<?php endif; ?>
-						<?php if( isset( $data['display_date'] ) ) : ?>
-						<div class="<?= esc_attr( $document->get_type() ); ?>-display-date">
-							<p class="form-field form-field-wide">
-								<p>
-									<span><strong><?= wp_kses_post( $data['display_date']['label'] ); ?></strong></span>
-									<span><?= esc_attr( $data['display_date']['value'] ); ?></span>
+						<div class='pdf-hide pdf-more-details'>
+							<?php if( isset( $data['display_date'] ) ) : ?>
```
alexmigf, line 680: Space missing after `if`.

alexmigf (changes requested):
Good work so far!

assets/js/order-script.js
```diff
@@ -154,4 +154,17 @@ jQuery( function( $ ) {
 		}
 	}
 
+	jQuery('#wpo_wcpdf-data-input-box').on('click', '.view-more', function(e) {
```
alexmigf, line 157: Don't forget the spacing here as you do in `php`.

alexmigf (changes requested):

assets/css/order-styles.css
```diff
@@ -175,3 +175,6 @@
 	0% { transform: rotate(0deg); }
 	100% { transform: rotate(-360deg); }
 }
+.pdf-hide {
```
alexmigf, line 178: I think you could avoid using this class by hiding the elements with JS.

alexmigf (changes requested):

assets/css/order-styles.css
aparnascodex: I have used this class to hide the details div and 'Hide details' link on page load. If there is delay in js loading due slow network or some other reason, then it will first display all the details and then it will hide elements after js is loaded. There will be a display glitch.

alexmigf: I understand, good point. Just include the instruction directly in the element then, without using the class. I don't think it worth having a class just for that.

alexmigf (approved):

PR #530, opened 2023-06-13
Fix: font deletion bug after plugin upgrade

MohamadNateqi:
I tried several times to reproduce this issue by installing version 3.5.4 and upgrading to the latest version 3.5.5 multiple times, but I did not encounter the reported problem.

Upon reviewing the codes, I found out that the upgrade method always copies fonts in any situation. So, this issue is strange because everything seems fine!

I read support tickets related to this problem in both HelpScout and wordpress.org after the latest release, and I saw a few tickets related to the fonts issue, but none of them specifically referred to any problems associated with the "upgrade" process.

Also, about another individual's comment in the main topic (https://wordpress.org/support/topic/had-to-manually-reinstall-plugin-fonts/), I'm not sure whether he had this issue after the upgrade or not!

Anyway, as Yorden suggested, I implemented the deletion of "wpo_wcpdf_subfolder_fonts_has_files" transient to ensure that the problem does not occur. But I would like to revisit this issue later to make sure whether it originates from our side or the user's side.

### DIFF
--- a/includes/class-wcpdf-install.php
+++ b/includes/class-wcpdf-install.php
@@ -216,9 +216,12 @@ class Install {
 		} else {
 			// don't try merging fonts with local when updating pre 2.0
 			$pre_2 = ( $installed_version == 'versionless' || version_compare( $installed_version, '2.0-dev', '<' ) );
-			$merge_with_local = $pre_2 ? false : true;
+			$merge_with_local = !$pre_2;
 			WPO_WCPDF()->main->copy_fonts( $font_path, $merge_with_local );
 		}
+
+        // to ensure fonts will be copied to the upload directory
+        delete_transient( 'wpo_wcpdf_subfolder_fonts_has_files' );
 		
 		// 1.5.28 update: copy next invoice number to separate setting
 		if ( $installed_version == 'versionless' || version_compare( $installed_version, '1.5.28', '<' ) ) {
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -522,7 +522,7 @@ class Main {
 		// * UNLESS the ‘UPLOADS’ constant is defined in wp-config (http://codex.wordpress.org/Editing_wp-config.php#Moving_uploads_folder)
 		//
 		// May also be overridden by the wpo_wcpdf_tmp_path filter
-		
+
 		$wp_upload_base = $this->get_wp_upload_base();
 		if( $wp_upload_base ) {
 			if( $append_random_string && $code = $this->get_random_string() ) {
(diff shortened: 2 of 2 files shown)


includes/class-wcpdf-install.php
```diff
@@ -216,9 +216,12 @@ protected function upgrade( $installed_version ) {
 		} else {
 			// don't try merging fonts with local when updating pre 2.0
 			$pre_2 = ( $installed_version == 'versionless' || version_compare( $installed_version, '2.0-dev', '<' ) );
-			$merge_with_local = $pre_2 ? false : true;
+			$merge_with_local = !$pre_2;
```
alexmigf, line 219: Can you explain this? I'm assuming that you're running versions superior to 2.0, so will always be `true` right?

alexmigf (changes requested):

includes/class-wcpdf-install.php
MohamadNateqi: Yes, I'm working on versions superior to 2.0, which means that `$merge_with_local` will always be `true`.
Could you please provide a bit more detail about what you'd like me to explain or consider?

alexmigf: You made this change for the code to be cleaner?

MohamadNateqi: Yes, my intention was to make it cleaner and simpler. Is it ok?

alexmigf: Sure, just to clarify! Good work.

alexmigf (approved):

includes/class-wcpdf-main.php
```diff
@@ -634,9 +634,6 @@ public function maybe_reinstall_fonts( $force = false ) {
 			$this->copy_fonts( $fonts_path );
 
 			// save to cache
-			if ( get_transient( 'wpo_wcpdf_subfolder_fonts_has_files' ) !== false ) {
-				delete_transient( 'wpo_wcpdf_subfolder_fonts_has_files' );
-			}
 			set_transient( 'wpo_wcpdf_subfolder_fonts_has_files', 'yes' , DAY_IN_SECONDS );
```
Terminator-Barbapapa, line 637: Where are we using this transient?

alexmigf: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/520-delete-font-transient-on-upgrade/includes/class-wcpdf-main.php#L575

Terminator-Barbapapa: Thanks. It indeed looks like we are already doing everything we can to make sure the fonts are available. I think running a cron to reinstall the fonts daily, like one of the users suggested, is a bit overkill. And also doesn't address the issue in a correct way.

Terminator-Barbapapa (approved):

PR #534, opened 2023-06-16
New: Add a frontend shortcode for packing slip

### LINKED ISSUE #528: Create shortcode to download packing slip from link
We support the shortcode for the [download Invoice link](https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/download-invoice-link-with-shortcode/), maybe we should also have one for the Packing Slip.

MohamadNateqi:
closes #528

### DIFF
--- a/includes/class-wcpdf-frontend.php
+++ b/includes/class-wcpdf-frontend.php
@@ -118,7 +118,7 @@ class Frontend {
 		), $atts );
 
 		$is_document_type_valid = false;
-		$documents = WPO_WCPDF()->documents->get_documents();
+		$documents              = WPO_WCPDF()->documents->get_documents();
 		foreach ( $documents as $document ) {
 			if ( $document->get_type() === $values['document_type'] ) {
 				$is_document_type_valid = true;
@@ -126,7 +126,7 @@ class Frontend {
 			}
 		}
 
-		if (!$is_document_type_valid) {
+		if ( ! $is_document_type_valid ) {
 			return;
 		}
 
--- a/includes/class-wcpdf-frontend.php
+++ b/includes/class-wcpdf-frontend.php
@@ -14,6 +14,7 @@ class Frontend {
 		add_filter( 'woocommerce_api_order_response', array( $this, 'woocommerce_api_invoice_number' ), 10, 2 );
 		add_action( 'wp_enqueue_scripts', array( $this, 'open_my_account_pdf_link_on_new_tab' ), 999 );
 		add_shortcode( 'wcpdf_download_invoice', array( $this, 'download_invoice_shortcode' ) );
+		add_shortcode( 'wcpdf_download_packing_slip', array( $this, 'download_packing_slip_shortcode' ) );
 	}
 
 	/**
@@ -103,6 +104,17 @@ class Frontend {
 	 * Download invoice frontend shortcode
 	 */
 	public function download_invoice_shortcode( $atts ) {
+		return $this->generate_document_link_shortcode($atts,'invoice');
+	}
+
+    /**
+	 * Download packing slip frontend shortcode
+	 */
+	public function download_packing_slip_shortcode( $atts ) {
+		return $this->generate_document_link_shortcode($atts, 'packing-slip');
+	}
+
+	private function generate_document_link_shortcode( $atts, $document_type ) {
 		global $wp;
 
 		if ( is_admin() ) {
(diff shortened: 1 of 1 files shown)


alexmigf (commented):
Thinking about this, maybe we could have a generic shortcode for all documents (just keeping the invoice one), like this:

`[wcpdf_download_pdf link_text="Download your Packing Slip PDF" order_id="23" document_type="packing-slip"]`

What you think?

Btw, here's the documentation for the Invoice one: https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/download-invoice-link-with-shortcode/

MohamadNateqi:
@alexmigf This is a good idea! I will do that too.

alexmigf:
@MohamadNateqi I think you can replace the Packing Slip shortcode with the generic, so only two will exist, the invoice (because of compatibility) and the generic.

includes/class-wcpdf-frontend.php
```diff
@@ -103,27 +104,42 @@ public function woocommerce_api_invoice_number ( $data, $order ) {
...
+			'order_id'      => '',
+			'link_text'     => __(
+				'Download ' . str_replace( '-', ' ', $atts['document_type'] ?? 'invoice' ) . ' (PDF)',
+				'woocommerce-pdf-invoices-packing-slips'
+			),
+			'document_type' => 'invoice'
 		), $atts );
 
-		if ( empty( $values ) ) {
+		if ( ! in_array( $values['document_type'], [ 'invoice', 'packing-slip', 'credit-note', 'proforma' ] ) ) {
```
alexmigf, line 134: I recommend replacing the explicit array with the function `get_documents()` and a `foreach` loop, because that way you are not limited to those 4 types and custom documents can also take advantage of the shortcode.

```php
$documents = WPO_WCPDF()->documents->get_documents();
	// do you stuff
}
```

includes/class-wcpdf-frontend.php
```diff
@@ -103,27 +104,42 @@ public function woocommerce_api_invoice_number ( $data, $order ) {
 	 * Download invoice frontend shortcode
 	 */
 	public function download_invoice_shortcode( $atts ) {
+		return $this->generate_document_link_shortcode( $atts );
+	}
+
+	/**
+	 * Download pdf frontend general shortcode
+	 */
+	public function download_pdf_shortcode( $atts ) {
```
alexmigf, line 113: Maybe we don't need this 2 functions and we could use the `generate_document_link_shortcode()` directly?

```php
add_shortcode( 'wcpdf_download_pdf', array( $this, 'generate_document_link_shortcode' ) );
```

alexmigf (changes requested):

includes/class-wcpdf-frontend.php
MohamadNateqi: Thanks for the hint about `get_documents()`, I was thinking about such thing.

includes/class-wcpdf-frontend.php
```diff
@@ -99,31 +100,41 @@ public function woocommerce_api_invoice_number ( $data, $order ) {
...
+		$is_document_type_valid = false;
+		$documents = WPO_WCPDF()->documents->get_documents();
+		foreach ( $documents as $document ) {
+			if ( $document->get_type() === $values['document_type'] ) {
+				$is_document_type_valid = true;
+				break;
+			}
+		}
+
+		if (!$is_document_type_valid) {
```
alexmigf, line 129: WPCS

alexmigf (changes requested):

includes/class-wcpdf-frontend.php
```diff
@@ -99,31 +100,41 @@ public function woocommerce_api_invoice_number ( $data, $order ) {
...
 		}
 
 		// Default values
-		$values = shortcode_atts (array(
-			'order_id'		=> '',
-			'link_text'		=> ''
+		$values = shortcode_atts( array(
+			'order_id'      => '',
+			'link_text'     => __(
+				'Download ' . str_replace( '-', ' ', $atts['document_type'] ?? 'invoice' ) . ' (PDF)',
```
alexmigf, line 114: Sorry, I think this translation string must be improved, is not easy to understand for translators. Maybe you should use a `sprintf` with a placeholder and also add the translators help comment, like this:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/bd60be023109387b4eb34c01ca2a7b9ee7115613/includes/class-wcpdf-admin.php#L715-L718

alexmigf (changes requested):

includes/class-wcpdf-frontend.php
alexmigf: Sure, no problem.

PR #540, opened 2023-07-03
Fix: Revert breaking long characters

### LINKED ISSUE #539: Revert overflow-wrap for order data table
A while ago we added soms CSS that made it possible to have very long text strings (like URLs) on our PDFs without having them run of the page. This was fixed in the following PR: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/445

However this also affects users that have long labels in the order data table. 

Scherm­afbeelding 2023-07-03 om 13 17 10

We've gotten multiple support tickets about this. I think it is better to revert this CSS style on the order data table. And to only provide the `overflow-wrap: anywhere;` rule as a code snippet fix for customers that require this on their PDFs, for instance because they display a URL there.

MohamadNateqi:
closes #539

### DIFF
--- a/templates/Simple/style.css
+++ b/templates/Simple/style.css
@@ -125,7 +125,7 @@ table.order-data-addresses {
 
 table.order-data-addresses td.address,
 table.order-data-addresses td.order-data {
-	overflow-wrap: anywhere;
+	overflow-wrap: break-word;
 }
 
 td.order-data {
--- a/templates/Simple/style.css
+++ b/templates/Simple/style.css
@@ -125,7 +125,7 @@ table.order-data-addresses {
 
 table.order-data-addresses td.address,
 table.order-data-addresses td.order-data {
-	overflow-wrap: break-word;
+	overflow-wrap: anywhere;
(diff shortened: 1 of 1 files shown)


MohamadNateqi:
@Terminator-Barbapapa From what I understood, you only wished to change things back in the `order-data` section. So I only updated that section.

alexmigf:
I think this should only be fixed for the labels and not for the value.

MohamadNateqi:
@alexmigf Do you mean what I said here? 
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/539#issuecomment-1618213629

alexmigf:
Correct.

MohamadNateqi:
@Terminator-Barbapapa Michael, do you think we should change it so that only the label is updated and not the value?

MohamadNateqi:
@Terminator-Barbapapa We would like your final opinion about this, Michael. Thank you :)

alexmigf (approved):

PR #546, opened 2023-07-12
Feature: Add an invoice number column to the WooCommerce order analytics

### LINKED ISSUE #452: Display the PDF document numbers in the WC Analytics
A user in the forums asked us how to display the Invoice number in the WC Analytics Orders report [here](https://wordpress.org/support/topic/invoice-number-in-woocommerce-analytics/?view=all#post-16306048).

Because this Analytics use JavaScript instead of PHP we need to investigate how to do this properly.

Reference: https://developer.woocommerce.com/2020/02/20/extending-wc-admin-reports/

MohamadNateqi:
closes #452


assets/js/analytics-order.build.js
```diff
@@ -0,0 +1 @@
...>{"use strict";function r(r){return function(r){if(Array.isArray(r))return e(r)}(r)||function(r){if("undefined"!=typeof Symbol&&null!=r[Symbol.iterator]||null!=r["@@iterator"])return Array.from(r)}(r)||function(r,t){if(r){if("string"==typeof r)return e(r,t);var n=Object.prototype.toString.call(r).slice(8,-1);return"Object"===n&&r.constructor&&(n=r.constructor.name),"Map"===n||"Set"===n?Array.from(r):"Arguments"===n||/^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n)?e(r,t):void 0}}(r)||function(){throw new TypeError("Invalid attempt to spread non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method.")}()}function e(r,e){(null==e||e>r.length)&&(e=r.length);for(var t=0,n=new Array(e);t<e;t++)n[t]=r[t];return n}(0,wp.hooks.addFilter)("woocommerce_admin_report_table","wpo-",(function(e){return"orders"!==e.endpoint?e:(e.headers=[].concat(r(e.headers),[{label:"Invoice Number",key:"order_invoice_number",screenReaderLabel:"Invoice Number",isSortable:!1}]),e.items&&e.items.data&&e.items.data.length?(e.rows=e.rows.map((function(t,n){var o=e.items.data[n];return[].concat(r(t),[{display:o.invoice_number,value:o.invoice_number}])})),e):e)}))})();
```
alexmigf, line 1: This shouldn't be minified by default since we have an action to create the minified version.

assets/js/analytics-order.mjs
```diff
@@ -0,0 +1,42 @@
...
+    'wpo-',
+    (reportTableData) => {
+        if (reportTableData.endpoint !== 'orders') {
+            return reportTableData;
+        }
+
+        reportTableData.headers = [
+            ...reportTableData.headers,
+            {
+                label: 'Invoice Number',
```
alexmigf, line 14: I think this should be ready for translation.

assets/js/analytics-order.mjs
```diff
@@ -0,0 +1,42 @@
...
+                key: 'order_invoice_number',
+                screenReaderLabel: 'Invoice Number',
```
alexmigf, line 16: This too.

includes/class-wcpdf-assets.php
```diff
@@ -200,7 +200,21 @@ public function backend_scripts_styles ( $hook ) {
...
-		
+
+		if (
+			$hook === 'woocommerce_page_wc-admin' &&
+			class_exists( 'Automattic\WooCommerce\Admin\PageController' ) &&
+			\Automattic\WooCommerce\Admin\PageController::is_admin_page()
+		) {
+			wp_enqueue_script(
+				'wpo-wcpdf-analytics-order',
+				WPO_WCPDF()->plugin_url() . '/assets/js/analytics-order.build.js',
```
alexmigf, line 211: Suffix is missing in the filename.

alexmigf (changes requested):

assets/js/analytics-order.build.js
MohamadNateqi: Yes, I'm building this file locally using the `pnpm run build` command, which produces a built file that is automatically minified, hence the `build` in the filename.
I can commit the unminified version and remove `build` from the filename.
Would that be acceptable?

alexmigf: Yes, the action will take care of the minification.

includes/class-wcpdf-assets.php
```diff
@@ -203,16 +203,21 @@ public function backend_scripts_styles ( $hook ) {
...
 				'wpo-wcpdf-analytics-order',
-				WPO_WCPDF()->plugin_url() . '/assets/js/analytics-order.build.js',
-				array( 'wp-hooks' ),
+				WPO_WCPDF()->plugin_url() . '/assets/js/analytics-order' . $suffix . '.js',
+				array( 'wp-hooks', 'wp-i18n' ),
 				WPO_WCPDF_VERSION,
 				true
 			);
+
+			wp_set_script_translations(
```
alexmigf, line 216: Wouldn't be easy to pass the strings using `wp_localize_script`?

Eg.

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e4d69034f97da4ab9bbd2aa9fa058a4fa9bb4586/includes/class-wcpdf-assets.php#L224

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e4d69034f97da4ab9bbd2aa9fa058a4fa9bb4586/assets/js/debug-script.js#L53

alexmigf (changes requested):

includes/class-wcpdf-assets.php
MohamadNateqi: Indeed that is simpler! I'm not sure why I didn't think of that before. Thank you!

assets/js/analytics-order.js
```diff
@@ -1,42 +1,38 @@
...
+            (function (r) {
+                if (Array.isArray(r)) return e(r);
+            })(r) ||
+            (function (r) {
+                if (("undefined" != typeof Symbol && null != r[Symbol.iterator]) || null != r["@@iterator"]) return Array.from(r);
+            })(r) ||
+            (function (r, t) {
+                if (r) {
+                    if ("string" == typeof r) return e(r, t);
+                    var n = Object.prototype.toString.call(r).slice(8, -1);
```
alexmigf, line 14: Maybe you should use `let` instead of `var`?

assets/js/analytics-order.js
```diff
@@ -1,42 +1,38 @@
 (() => {
     "use strict";
-    const e = wp.hooks,
-        r = wp.i18n;
-    function t(e) {
+    function r(r) {
```
alexmigf, line 3: Shouldn't the function names be more explicit? I haven't checked WC code, is this something they do as well?

alexmigf (changes requested):
Great work!

assets/js/analytics-order.js
MohamadNateqi: I created this version using `pnpm` and unminified it for the GitHub action.
The file was built from the `analytics-order.mjs` script, which should be edited directly (I renamed it to .mjs to avoid github action error because of the `import` usage in the script).
So I think keeping the file as it is maybe appropriate, but editing would also be ok.
The WC uses the build version, such as `assets/client/admin/chunks/analytics-report-orders.js` for orders in analytics.
Also, it is better not to include the `.mjs` file from production since it is unnecessary and only for developers' future reference.
What do you think?

alexmigf: Ok, let's keep like that.

includes/class-wcpdf-admin.php
```diff
@@ -1194,6 +1196,14 @@ public function process_order_document_form_data( $form_data, $document_slug )
 
 		return $data;
 	}
+
+	public function add_invoice_number_to_order_report( $response ) {
+		if ( ! empty( $order = wc_get_order( $response->data['order_id'] ) ) ) {
```
alexmigf, line 1201: I think the variable assignment should be in one separated line to comply with WPCS.

alexmigf (changes requested):

includes/class-wcpdf-admin.php
MohamadNateqi: Yeah, thanks for noticing.

YordanSoares (approved):
The Invoice Number is displayed under Analytics > Orders:

image

Also, it is included in the report files you can download in this screen:

image

You can toggle the column if you want, so this is a flexible column:

image

Great job @MohamadNateqi! 🎉

PR #550, opened 2023-07-25
New: adds a new Status tool to reset settings

### LINKED ISSUE #549: Add option to reset plugin settings
Currently there is no way of resetting the plugin settings. Maybe we could add a selector in the Status tools to do that.

alexmigf:
closes #549

Screenshot from 2023-07-25 09-30-06


includes/class-wcpdf-settings-debug.php
```diff
@@ -348,6 +373,66 @@ public function import_settings( $data ) {
...
+		
+		switch ( $type ) {
+			case 'general':
+				$settings_option = 'wpo_wcpdf_settings_general';
+				break;
+			case 'debug':
+				$settings_option = 'wpo_wcpdf_settings_debug';
+				break;
+			default:
+				$settings_option = apply_filters( 'wpo_wcpdf_reset_settings_option', $settings_option, $type );
```
MohamadNateqi, line 395: I think it will be better to make this hook, and the hook used in the `export_settings` method line 272 have the same pattern in their name.

includes/class-wcpdf-settings-debug.php
```diff
@@ -348,6 +373,66 @@ public function import_settings( $data ) {
...
+				wp_send_json_error( compact( 'message' ) );
+			}
+		}
+		
+		$updated = update_option( $settings_option, [] );
+		if ( $updated ) {
+			$message = sprintf(
+				/* translators: settings type */
+				__( '%s settings reset successfully!', 'woocommerce-pdf-invoices-packing-slips' ),
+				$setting_types[$type]
```
MohamadNateqi, line 421: `$setting_types[$type]` needs to be `$type`.

includes/class-wcpdf-settings-debug.php
```diff
@@ -348,6 +373,66 @@ public function import_settings( $data ) {
...
+			);
+			wcpdf_log_error( $message, 'info' );
+			wp_send_json_success( compact( 'type', 'message' ) );
+		} else {
+			$message = sprintf(
+				/* translators: settings type */
+				__( 'An error occurred when trying to reset the %s settings.', 'woocommerce-pdf-invoices-packing-slips' ),
+				$setting_types[$type]
```
MohamadNateqi, line 429: Same.

assets/js/debug-script.js
```diff
@@ -17,20 +17,29 @@ jQuery( function( $ ) {
...
-			contentType: false,
-			success ( response ) {
-				process_form_response( tool, response, $form );
-			},
-			error ( xhr, error, status ) {
-				//console.log( error, status );
-			}
-		} );
+		let reset = false;
+		if ( 'reset-settings' == tool ) {
```
MohamadNateqi, line 21: WPCS - it would be better not to use loose comparisons

MohamadNateqi (changes requested):

includes/class-wcpdf-settings-debug.php
alexmigf: Can you explain better?

MohamadNateqi: Sure, I meant it might be better to write it as `wpo_wcpdf_reset_settings`, the same as `wpo_wcpdf_export_settings` without the `option` at the end or adding an `_option` at the end of the `wpo_wcpdf_export_settings`.

alexmigf: Are different things, in the `wpo_wcpdf_export_settings` the hook return value is the settings itself, and `array`, and in the `wpo_wcpdf_reset_settings_option` is the settings option key, a `string`.

MohamadNateqi (approved):
LGTM!

dwalkerpriv (approved):
Works pretty well on my end

includes/class-wcpdf-settings-debug.php
```diff
@@ -348,6 +373,66 @@ public function import_settings( $data ) {
...
+				/* translators: settings type */
+				__( '%s settings reset successfully!', 'woocommerce-pdf-invoices-packing-slips' ),
+				$type
+			);
+			wcpdf_log_error( $message, 'info' );
+			wp_send_json_success( compact( 'type', 'message' ) );
+		} else {
+			$message = sprintf(
+				/* translators: settings type */
+				__( 'An error occurred when trying to reset the %s settings.', 'woocommerce-pdf-invoices-packing-slips' ),
```
YordanSoares, line 428: This error message is triggered when you try to delete twice a settings option. It's a similar case to which I commented when you added the export/export feature ([here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/473#discussion_r1095123603)), but the other way around:

image

But more than an error, this should be a notice, letting know the users that the settings option was already reset.

includes/class-wcpdf-settings-debug.php
YordanSoares, line 395: As I installed the Professional extension, I realized that the Pro settings are selectable too:

image

But the reset does not work for this settings option yet:

image


Will the Pro settings resetting available in the Professional extension instead, or maybe should we to do something like the following?:

```php
$settings_option = apply_filters( 'wpo_wcpdf_reset_settings_option', "wpo_wcpdf_settings_{$type}", $type );
```

YordanSoares (changes requested):
I tested resetting all these options and the reset was successful:

1. General
2. Debug
3. Invoice
4. Packing Slip
5. Proforma Invoice
6. Credit Note

However, I'm leaving some remarks below.

includes/class-wcpdf-settings-debug.php
alexmigf: Pro and Templates are not ready yet, we need to open new issues there to hook here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/690defa47ac1713b382554ce56e5bc994140a069/includes/class-wcpdf-settings-debug.php#L395

I have changed the notice message, in case they had a legacy version of the extension without the reset support.

alexmigf: https://github.com/wpovernight/woocommerce-pdf-ips-pro/issues/257
https://github.com/wpovernight/woocommerce-pdf-ips-templates/issues/229

YordanSoares (approved):
LGTM! 🚀

PR #562, opened 2023-08-14
New: document link access type implementation

### LINKED ISSUE #561: Force the use of the order key if guest access is enabled
Currently even if "**Guest access**" is enabled, if the link is generated from backend it has the `nonce` attached, but maybe we should have another option to force the use of the **order key** instead, because this way we could copy and paste the document link in an external source and still be able to see it.

alexmigf:
closes #561

This PR introduces a new Status tab setting, which replaces the old "Guest Access":

Screenshot from 2023-08-15 12-36-33

The 3 options are:

* **Logged in (recommended)**: it's the default and behaves exactly as before without the 'Guest access' enabled. A user must be logged in to access the document in the admin panel, and in the frontend must be the order owner. Uses nonce.
* **Guest**: behaves exactly as before with the order key, but this time it can also behave as "Logged in", if the user is logged in using a nonce.
* **Full**: is like 'Guest' but without checking the logged in status, it always use the order key instead of nonce.

A migration is also in place for previous versions of the plugin, passing the previous setting value if exists, to the new key.


dwalkerpriv (approved):
Works on my end.

A small correction in the description:

'By default, documents can only be accessed when logged in, from **the** admin panel or customer account. You could also give access permissions to guest customers, or full permission to everyone with a document link.'

I think a 'the' should be added before the 'admin panel'

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -520,14 +520,20 @@ public function init_settings() {
...
 				'section'  => 'debug_settings',
-				'args'     => array(
+				'args'	   => array(
 					'option_name' => $option_name,
-					'id'          => 'guest_access',
-					'description' => __( 'Enable this to allow customers that purchase without an account to access their PDF with a unique key', 'woocommerce-pdf-invoices-packing-slips' ),
+					'id'          => 'document_link_access_type',
+					'default'     => 'logged_in',
+					'options'     => array(
+						'logged_in' => __( 'Logged in (recommended)' , 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi, line 532: WPCS - space before the comma

includes/class-wcpdf-main.php
```diff
@@ -341,39 +342,67 @@ public function generate_pdf_ajax() {
...
-			if ( ! is_user_logged_in() ) {
-				$allowed = false;
-			}
-
-			// Check the user privileges
-			$full_permission = WPO_WCPDF()->admin->user_can_manage_document( $document_type );
-			if ( ! $full_permission ) {
-				if ( ! isset( $_GET['my-account'] ) && ! isset( $_GET['shortcode'] ) ) {
+				
+				if ( ! $valid_nonce ) {
```
MohamadNateqi, line 358: Maybe good to merge with the above if statement.

includes/class-wcpdf-main.php
```diff
@@ -341,39 +342,67 @@ public function generate_pdf_ajax() {
...
+				}
+				
+				// single order only
+				if ( count( $order_ids ) > 1 ) {
+					$allowed = false;
+					break;
+				}
+				
+				// no order
+				if ( ! $order ) {
```
MohamadNateqi, line 401: Shouldn't this condition be at first?

MohamadNateqi (changes requested):

MohamadNateqi (approved):
LGTM!

YordanSoares (commented):
These are the tests I had run:

1. A new installation: It worked well for all the options ✅ 
2. Updating an existing installation: The migration was not made for the existing public release (v3.5.6) ⚠️ 
3. Tested the code snippet to add the download link on the thank-you page ([here](https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/download-link-on-the-thank-you-page/#enabling-guest-access)) ✅

includes/class-wcpdf-install.php
```diff
@@ -433,6 +433,16 @@ protected function upgrade( $installed_version ) {
 				WPO_WCPDF()->settings->schedule_yearly_reset_numbers();
 			}
 		}
+		
+		// 3.5.7-dev-1: migrate 'guest_access' setting to 'document_link_access_type'
+		if ( version_compare( $installed_version, '3.5.7-dev-1', '<' ) ) {
```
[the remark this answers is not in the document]
alexmigf: No, you need to change here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a30ed16ffc7b24d001b78d09848083d41feb6ada/woocommerce-pdf-invoices-packingslips.php#L24

YordanSoares (approved):
Alex, I updated the plugin version to test the migration and it worked well:

image

BTW, the improvement you just made to the description makes clearer what each option do: Great job!

PR #566, opened 2023-08-15
Fix: UTF-8 decode functions bug with `mb_convert_encoding`

### LINKED ISSUE #565: iconv(): Wrong encoding
by @YordanSoares 

I tested the above on Local (by Flywheel). However, I first try testing on tastewp.com, but it returned this issue while trying to generate a PDF document with this PR (PHP 8.x.x):

```
Warning:  iconv(): Wrong encoding, conversion from "UTF-8" to "ISO-8859-1//IGNORE" is not allowed in /wp-content/plugins/woocommerce-pdf-invoices-packing-slips-561-new-document-link-access-type-implementation/vendor/symfony/polyfill-mbstring/Mbstring.php on line 116
```

...and this is the error for PHP 7.x.x:

```
Notice:  iconv(): Wrong charset, conversion from `UTF-8' to `ISO-8859-1//IGNORE' is not allowed in /wp-content/plugins/woocommerce-pdf-invoices-packing-slips-561-new-document-link-access-type-implementation/vendor/symfony/polyfill-mbstring/Mbstring.php on line 116
````

I have tested on tastewp.com with PHP 7.3, 7.4, 8.0, 8.1, and 8.2, with the same results.

It seems that this is happening after merging the PR https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/557.

I'm aware that these testing providers have a minimal server setup, but the PDF invoice generation have been working so far. Maybe should we try to add more compatibility with these kinds of setups?

alexmigf:
closes #565

### DIFF
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -302,11 +302,6 @@ function wcpdf_utf8_decode( $string ) {
 		return $string;
 	}
 	
-	// provided by composer 'symfony/polyfill-mbstring' library.
-	if ( class_exists( '\\Symfony\\Polyfill\\Mbstring\\Mbstring' ) ) {
-		return \Symfony\Polyfill\Mbstring\Mbstring::mb_convert_encoding( $string, 'ISO-8859-1', 'UTF-8' );
-	}
-	
 	// 'utf8_decode()' is deprecated in PHP 8.2 and set to be removed on PHP 9.0.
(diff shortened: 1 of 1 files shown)


alexmigf:
@YordanSoares did you tried on PHP 8.2?

YordanSoares:
@alexmigf yes, I did: It's working on every PHP version, since 7.2.x until 8.2.x 👍

PR #571, opened 2023-08-16
New: filter hook to impeding Shipping Notes strip HTML tags

### LINKED ISSUE #570: Shipping notes not respecting line breaks
Because now we allow Shipping notes to be translated, we are removing all HTML tags here:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/559/files#diff-3c27d95148592338471e6af2989845953abad29560a97a4e73b989aaa9d1253fR1098

Maybe we could add a filter to ignore that for customers that require the tags.

alexmigf:
closes #570

New filter to impede strip HTML tags from the Shipping Notes: `wpo_wcpdf_shipping_notes_strip_all_tags`

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -1091,9 +1091,12 @@ abstract class Order_Document_Methods extends Order_Document {
 			$shipping_notes = '';
 		}
 		
-		$shipping_notes = wp_strip_all_tags( $shipping_notes );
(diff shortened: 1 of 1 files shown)


YordanSoares (approved):
I entered the following multiline text in the shipping notes:

image

This is how the notes behave by default:

image

I added this code snippet, to disable the tag removal using the filter hook added in this PR:

```php
add_filter( 'wpo_wcpdf_shipping_notes_strip_all_tags', '__return_false' );
```
And this is how the notes behave after activating the code snippet above:

image

PR #575, opened 2023-08-18
Tweak: removes legacy mode & textdomain from debug settings

### LINKED ISSUE #553: Remove legacy settings
In the status tab we have 2 settings for compatibility with legacy versions that was introduced [6 years ago](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/c1114eb03a1e0b2e3e379f5518b56d357aafdcc9) when the plugin version version changed to `2.0.0`.

Screenshot from 2023-07-28 14-35-49

Maybe it's time to move on and remove this? I see some customers having this options enabled using the latest version and no custom templates.

For hooks and functions we could deprecated them, this way customers with older custom templates will get notices to update them.

alexmigf:
closes #553

I'm keeping the `legacy_mode` setting commented and the classes in case something goes wrong and we need to go back. I think we could safely remove them after some time.

### DIFF
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -24,7 +24,6 @@ class WPO_WCPDF {
 	public $version = '3.6.1';
 	public $plugin_basename;
 	public $legacy_mode;
-	public $legacy_textdomain;
 	public $third_party_plugins;
 	public $order_util;
 	public $settings;
@@ -70,11 +69,6 @@ class WPO_WCPDF {
 		add_action( 'admin_notices', array( $this, 'nginx_detected' ) );
 		add_action( 'admin_notices', array( $this, 'mailpoet_mta_detected' ) );
 		add_action( 'admin_notices', array( $this, 'rtl_detected' ) );
-
-		// legacy textdomain fallback
-		if ( $this->legacy_textdomain_enabled() === true ) {
-			add_filter( 'load_textdomain_mofile', array( $this, 'textdomain_fallback' ), 10, 2 );
-		}
 	}
 	
 	private function autoloaders() {
@@ -104,9 +98,6 @@ class WPO_WCPDF {
 		$dir    = trailingslashit( WP_LANG_DIR );
 
 		$textdomains = array( 'woocommerce-pdf-invoices-packing-slips' );
-		if ( $this->legacy_mode_enabled() === true ) {
-			$textdomains[] = 'wpo_wcpdf';
-		}
 
 		/**
 		 * Frontend/global Locale. Looks in:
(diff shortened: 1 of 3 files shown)


vendor/composer/autoload_classmap.php
```diff
@@ -217,12 +217,9 @@
     'WPO\\WC\\PDF_Invoices\\Font_Synchronizer' => $baseDir . '/includes/class-wcpdf-font-synchronizer.php',
     'WPO\\WC\\PDF_Invoices\\Frontend' => $baseDir . '/includes/class-wcpdf-frontend.php',
     'WPO\\WC\\PDF_Invoices\\Install' => $baseDir . '/includes/class-wcpdf-install.php',
-    'WPO\\WC\\PDF_Invoices\\Legacy\\Deprecated_Hooks' => $baseDir . '/includes/legacy/class-wcpdf-legacy-deprecated-hooks.php',
-    'WPO\\WC\\PDF_Invoices\\Legacy\\Legacy_Document' => $baseDir . '/includes/legacy/class-wcpdf-legacy-document.php',
-    'WPO\\WC\\PDF_Invoices\\Legacy\\Legacy_Export' => $baseDir . '/includes/legacy/class-wcpdf-legacy-export.php',
+    'WPO\\WC\\PDF_Invoices\\Legacy' => $baseDir . '/includes/class-wcpdf-legacy.php',
```
MohamadNateqi, line 220: Shouldn't these be deleted?

woocommerce-pdf-invoices-packingslips.php
MohamadNateqi, line 1: It seems these two need to be removed:
```
	public $legacy;
	public $deprecated_hooks;
```

MohamadNateqi (changes requested):
There are some codes/comments that have a legacy keyword in the project, but I'm not sure if they're related to the legacy mode, such as:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a68faa7d249763bf26ea34e5ecacf06f549eb9c7/includes/class-wcpdf-settings.php#L545

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a68faa7d249763bf26ea34e5ecacf06f549eb9c7/includes/class-wcpdf-admin.php#L1049

Would it be possible to take a look again at these?

alexmigf:
> $form_data = $this->legacy_date_fields_replacements( $form_data, $document->slug );

This seems more recent, 2 years ago, and not related to the other legacy stuff.

MohamadNateqi (approved):
LGTM!

YordanSoares (approved):
LGTM!

PR #578, opened 2023-08-18
Fix: bug when the temporary directory is not secured but the option exists in the database

### LINKED ISSUE #577: If the temporary directory random string exists, and is not secured, the tool keeps hidden
If the temporary directory exists with the name `wpo_wcpdf`, and is not secured but the option exists in the database, the tool to generate the random string keeps hidden which isn't ideal if we want to secure it.

alexmigf:
closes #577

This will make the option to generate a new random string for the temporary directory always visible:

Screenshot from 2023-08-18 16-41-47

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -686,9 +686,9 @@ class Main {
 	 * Get random string
 	 */
 	public function get_random_string () {
-		$code = sanitize_text_field( get_option( 'wpo_wcpdf_random_string' ) );
-		if( $code ) {
-			return $code;
+		$code = get_option( 'wpo_wcpdf_random_string', '' );
+		if ( ! empty( $code ) ) {
+			return esc_attr( $code );
 		} else {
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


MohamadNateqi (approved):
LGTM!

MohamadNateqi:
@alexmigf The reason the customer had a directory without the code is still unknown, right?

YordanSoares:
@MohamadNateqi it seems that this happened because they are using NGINX with certain security configuration.

alexmigf:
Maybe something is blocking the directory with the random string to be created in this case, because it seems the string is saved in the database. I asked for a temporary login to see if I can find something else.

YordanSoares (approved):
Alex, although I was not able to replicate the user scenario, now the button is appearing always.

One extra remarks, not related with this PR: I realized that when you click on it, it generates a random string with its respective folder, without deleted the previous one. I tried clicking several times, and this is the result:

image

Should we maybe search folders with the `wpo_wcpdf_` prefix to delete those different to the new folder created?

alexmigf:
@YordanSoares  Yes, we should improve that. Please open a new issue.

YordanSoares:
@alexmigf done: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/580

PR #581, opened 2023-08-21
Fix: remove old temp dir when generating new one from status tool

### LINKED ISSUE #580: Delete old plugin's folders after generating a new temporary folder
I realized that when you click on the **Generate temporary directory** button, under the **Status** tab, it generates a new temporary folder, without deleting the old one. I tried clicking several times, and this is the result:

Therefore, we should search folders with the `wpo_wcpdf_` prefix to delete those plugin's folders different to the new generated temporary folder.

alexmigf:
closes #580

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -791,11 +791,17 @@ class Main {
 	/**
 	 * Copy contents from one directory to another
 	 */
-	public function copy_directory ( $old_path, $new_path ) {
-		if( empty($old_path) || empty($new_path) ) return;
-		if( ! is_dir($old_path) ) return;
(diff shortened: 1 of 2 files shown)


MohamadNateqi (approved):
LGTM!

YordanSoares (commented):
@alexmigf, now the old folders are not remaining as residual folders.

However, I realized that the fonts break:

image

Maybe should we run the fonts reinstalling after clicking on the **Generate temporary directory** button, what do you think?

alexmigf:
@YordanSoares good catch, because the font from my PDF from testing was really close to the expected one.

YordanSoares (approved):
✅ Now it works perfectly!

PR #583, opened 2023-08-22
Fix: bug of passing the wrong variable for the filtered order emails

alexmigf:
This PR fixes the issue with filtered order emails not being added to the metabox, for example the Order Notification email from the Professional extension.


MohamadNateqi (approved):
LGTM!

PR #597, opened 2023-08-31
Fix: add missing "Invoice Number" column to orders export of WooCommerce analytics

### LINKED ISSUE #595: Invoice column is missing in the WooCommerce analytic orders export via email
WooCommerce handles the exporting process of analytics in the background when dealing with a large dataset. Once the preparation is complete, the dataset is sent to the user via email.

A customer [reported](https://secure.helpscout.net/conversation/2344951595/53687/) that the invoice column exists when a direct download happens, but this field is missing when it is sent via email.

So this needs to be investigated, and if it is indeed missing, it should be solved.

MohamadNateqi:
closes #595 

If you require multiple dummy orders to be created, you can use this plugin:
https://wordpress.org/plugins/generate-random-orders-for-woocommerce/.

If used, an update on their status using bulk action is recommended.

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -70,7 +70,10 @@ class Admin {
 		// document actions
 		add_action( 'wpo_wcpdf_document_actions', array( $this, 'add_regenerate_document_button' ) );
 
+        // add "invoice number" column to WooCommerce Analytic - Orders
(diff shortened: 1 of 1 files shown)


YordanSoares (approved):
The invoice number is included in the report download sent by email:

image

alexmigf (approved):

PR #600, opened 2023-09-08
New: bump min required PHP version to 7.2

### LINKED ISSUE #585: Bump PHP requirement version to 7.2
We added a notice [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/582) to inform users running older PHP versions that we will bump the requirement version to 7.2, we should give them some time before making this change, but we should do it in a month or 2.

alexmigf:
closes #585


MohamadNateqi (approved):
LGTM!

PR #604, opened 2023-09-14
Fix: PHP error on calling `get_title()` document method on bool

### LINKED ISSUE #603: Call to a member function `get_title()` on bool
When using the Number Tools extension a fatal error is triggered:

> PHP Fatal error:  Uncaught Error: Call to a member function get_title() on bool in /woocommerce-pdf-invoices-packing-slips/includes/views/wcpdf-settings-page.php:110

This happens because `numbers` is being passed here:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/456cafcc4a5f5c00493583b1de968571622208cf/includes/views/wcpdf-settings-page.php#L85

Which is not a valid document type, so `$document` is `false`.

https://github.com/wpovernight/woocommerce-pdf-ips-number-tools/issues/19#issuecomment-1716938229

alexmigf:
closes #603

### DIFF
--- a/includes/class-wcpdf-documents.php
+++ b/includes/class-wcpdf-documents.php
@@ -78,7 +78,7 @@ class Documents {
 	}
 
 	public function get_document( $document_type, $order ) {
-		foreach ( $this->get_documents( 'all' ) as $class_name => $document) {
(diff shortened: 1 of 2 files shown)


MohamadNateqi (approved):
LGTM!

PR #608, opened 2023-09-15
Fix: output html URL query string not working

### LINKED ISSUE #607: The URL query string to output HTML seems broken
After the UBL code being merge, I discovered that the output HTML URL query string seems broken.

alexmigf:
closes #607


MohamadNateqi (approved):
LGTM!

PR #609, opened 2023-09-15
Tweak: hide next Invoice number if displaying order number

### LINKED ISSUE #605: Hide document next number option when using order number
When using the document number as order number, we should hide the next number option.

alexmigf:
closes #605


MohamadNateqi (approved):
LGTM!

PR #612, opened 2023-09-21
Fix: my account PDF link conflict with Jetpack analytics module

### LINKED ISSUE #611: My account PDF links not working because of a conflict with Jetpack
A user reported that he couldn't access the PDF document links from the My Account order page. After investigation in the user store, I could find that the PDF URL was changed:

`admin-ajax.php?action=generate_wpo_wcpdf&_wca_initiator=action#038;document_type=invoice&order_ids=42905&access_key=99999999&my-account=true`

This part shouldn't be there:

`_wca_initiator=action#038;`

And it seems related to a Jetpack module: https://github.com/Automattic/jetpack/blob/1a062c5388083c7f15b9a3e82e61fde838e83047/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-my-account.php#L235

alexmigf:
closes #611

One user tested and it's working fine :heavy_check_mark:

### DIFF
--- a/includes/class-wcpdf-frontend.php
+++ b/includes/class-wcpdf-frontend.php
@@ -19,7 +19,7 @@ class Frontend {
 	}
 	
 	public function __construct()	{
-		add_filter( 'woocommerce_my_account_my_orders_actions', array( $this, 'my_account_pdf_link' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


MohamadNateqi (approved):
LGTM!

PR #615, opened 2023-09-25
Fix: bug when displaying the Invoice number/date columns. Significant drop of performance.

### LINKED ISSUE #614: Avoid unnecessary work if column not matching the custom plugin ones
A [user](https://wordpress.org/support/topic/performance-issue-with-your-plugin/) reported a significant drop in performance when using the Invoice number/date columns. This happens because we are not checking if the column is from our plugin, so it will run the same code for every column.

alexmigf:
closes #614


MohamadNateqi (approved):
LGTM!

PR #618, opened 2023-09-28
Fix: bug on UBL notice not hiding

### LINKED ISSUE #617: UBL notice not hiding when all settings are filled
A customer reported that the new UBL notice is not hiding even when all the settings are filled:

Scherm-afbeelding-2023-09-27-om-19-29-20

alexmigf:
closes #617


MohamadNateqi (approved):
LGTM!

PR #621, opened 2023-10-04
Fix: bug with HTML output using pretty links. Improved endpoint. 

### LINKED ISSUE #620: The `&output=html` parameter does not work with pretty links
Although the **Output to HTML** setting does work with pretty links, the `&output=html` parameter does not.

alexmigf:
closes #620 

This PR fixes an issue when trying to output the document as HTML using pretty links.

This PR also improves the pretty link to include the output at the end:

- PDF render: `.../wcpdf/invoice/575/e28f82db34/pdf`
- HTML render: `.../wcpdf/invoice/575/e28f82db34/html`
- UBL download: `.../wcpdf/invoice/575/e28f82db34/ubl`
- The default link still uses: `&output=html`

### DIFF
--- a/includes/class-wcpdf-endpoint.php
+++ b/includes/class-wcpdf-endpoint.php
@@ -133,9 +133,6 @@ class Endpoint {
 		// handle additional query vars
 		$additional_vars = apply_filters( 'wpo_wcpdf_document_link_additional_vars', $additional_vars, $order, $document_type );
 		if ( ! empty( $additional_vars ) && is_array( $additional_vars ) ) {
-			if ( isset( $additional_vars['output'] ) && $this->pretty_links_enabled() ) {
-				unset( $additional_vars['output'] );
-			}
 			$document_link = add_query_arg( $additional_vars, $document_link );
 		}
 
--- a/includes/class-wcpdf-endpoint.php
+++ b/includes/class-wcpdf-endpoint.php
@@ -133,6 +133,9 @@ class Endpoint {
 		// handle additional query vars
 		$additional_vars = apply_filters( 'wpo_wcpdf_document_link_additional_vars', $additional_vars, $order, $document_type );
 		if ( ! empty( $additional_vars ) && is_array( $additional_vars ) ) {
+			if ( isset( $additional_vars['output'] ) && $this->pretty_links_enabled() ) {
+				unset( $additional_vars['output'] );
+			}
 			$document_link = add_query_arg( $additional_vars, $document_link );
 		}
 
--- a/includes/class-wcpdf-endpoint.php
+++ b/includes/class-wcpdf-endpoint.php
@@ -112,12 +112,13 @@ class Endpoint {
 		}
 
 		if ( $this->pretty_links_enabled() ) {
+			$output     = isset( $additional_vars['output'] ) ? esc_attr( $additional_vars['output'] ) : 'pdf';
 			$parameters = array(
 				$this->get_identifier(),
 				$document_type,
 				$order->get_id(),
 				$access_key,
-				'pdf'
+				$output
 			);
 			$document_link = trailingslashit( get_home_url() ) . implode( '/', $parameters );
 		} else {
@@ -132,6 +133,9 @@ class Endpoint {
 		// handle additional query vars
 		$additional_vars = apply_filters( 'wpo_wcpdf_document_link_additional_vars', $additional_vars, $order, $document_type );
 		if ( ! empty( $additional_vars ) && is_array( $additional_vars ) ) {
+			if ( isset( $additional_vars['output'] ) && $this->pretty_links_enabled() ) {
+				unset( $additional_vars['output'] );
+			}
 			$document_link = add_query_arg( $additional_vars, $document_link );
 		}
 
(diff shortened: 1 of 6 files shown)


MohamadNateqi (changes requested):
The function name in this line needs to be updated:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/febe8ca5b1949be2a18301c35fb7950dbedeff69/includes/class-wcpdf-frontend.php#L85

Also, the `! empty( $wp->query_vars['output'] )` condition in this line:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1a12a6dccfda033a109297601008d17d03adf989/includes/class-wcpdf-endpoint.php#L78
prevents the document from displaying when the 'pretty document link' is checked for me when I click any type of document on the order page.

alexmigf:
> prevents the document from displaying when the 'pretty document link' is checked for me when I click any type of document on the order page.

@MohamadNateqi  Can you paste the document link here to see it?

MohamadNateqi:
@alexmigf 
`http://localhost/mohamad/wcpdf/invoice/330/wc_order_exOOE3LBZ6lmH/pdf/`
It will display my home page.

alexmigf:
What happens if you save the permalinks?

MohamadNateqi:
It'll save it as an HTML document which is my home page.

alexmigf:
No, I mean go under **Settings > Permalinks > Click save changes**, and try to open the document again from the order page.

MohamadNateqi:
It works now.

alexmigf:
Great! This commit should fix that without the need to do the steps manually: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/621/commits/260c4b0044a9d6892c2989552bd10b0e2c70d1d4

includes/class-wcpdf-endpoint.php
```diff
@@ -129,6 +133,9 @@ public function get_document_link( $order, $document_type, $additional_vars = ar
 		// handle additional query vars
 		$additional_vars = apply_filters( 'wpo_wcpdf_document_link_additional_vars', $additional_vars, $order, $document_type );
 		if ( ! empty( $additional_vars ) && is_array( $additional_vars ) ) {
+			if ( isset( $additional_vars['output'] ) && $this->pretty_links_enabled() ) {
```
MohamadNateqi, line 136: Since the `add_query_arg()` will update existing queries, I think there is no harm in removing this if-statement.

MohamadNateqi (approved):
LGTM!

YordanSoares (commented):
1. The UBL slug seems not to be working, it returns this error: 

image

2. I think we should not pass `/pdf` as default, but leave it clean, like this:

> https://example.com/wcpdf/invoice/99/39a98abd4b/

alexmigf:
> The UBL slug seems not to be working, it returns this error:

@YordanSoares can you follow [this steps](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/621#issuecomment-1748625564) and let me know?

> I think we should not pass /pdf as default, but leave it clean, like this:

I had a customer this year that wanted to have `.pdf` to the pretty link. At the time I told him that was not possible, but now that I did this way it seems logic to me, because from the link we can't identify the real output of it. In my opinion I like it, doesn't shock me, but I understand that without it seems more clean. But I'm afraid that removing it will also affect the other methods, specially the `html` which I think it's simpler now for the pretty links. @MohamadNateqi what's your opinion?

MohamadNateqi:
This option will make the link pretty, and I think it is still pretty with the output type at the end. Since this is not shortening the URL, I think it's OK to have it. Additionally, we have some benefits that Alex mentioned. So, I think keeping it is better.

YordanSoares:
@alexmigf I ran two separated tests from scratch (on a live environment and the other on Local), following [your advice here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/621#issuecomment-1748625564), but I still have issues with the UBL document.

This is the output I get in the live site:

image

...and a similar screen returned on Local:

image

Both site generated this URL pattern: `/wcpdf/invoice/63/1a2b3c4d5e/ubl?output=ubl`

I tried with `/wcpdf/invoice/63/1a2b3c4d5e/ubl`, but it did not work either.

alexmigf:
@YordanSoares can I have access to your live staging?

YordanSoares:
@alexmigf sure! I have sent you the link via DM.

alexmigf:
@YordanSoares could you try now?

YordanSoares (approved):
LGTM! 

Just a remark: Even if you deactivate the UBL invoice, it can be retrieved if you change the last part of the URL from `/pdf` to `/ubl`, is this the expected behavior?

alexmigf:
@YordanSoares should be fixed now, please confirm.

YordanSoares:
@alexmigf now the UBL is not download if it is disabled, although it still opens/generates the PDF output, instead of displaying a notice, although this is an edge case, so I don't think it's too much trouble. What do you think?

alexmigf:
I know, `pdf` is the default output, and since you opened a pdf link I don't see a trouble with that.

YordanSoares (approved):
Then 🚀 !

MohamadNateqi (approved):
LGTM!

includes/class-wcpdf-endpoint.php
MohamadNateqi: It seems there was some harm, yeah?

alexmigf: No, but was adding the `?output=ubl` in front of the pretty URL, which already includes `/ubl`.

PR #626, opened 2023-10-09
Fix: error when calling order ID on null under `output_number_date_edit_fields` method

### LINKED ISSUE #625: Uncaught Error: Call to a member function `get_id()` on `null`
A user reported the error below using the last version `3.6.3`:

> Une erreur de type E_ERROR a été causée dans la ligne 661 du fichier .../wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-admin.php. Message d'erreur : Uncaught Error: Call to a member function get_id() on null in .../wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-admin.php:661
Stack trace:
#0 .../wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-admin.php(600): WPO\WC\PDF_Invoices\Admin->output_number_date_edit_fields()
#1 .../wp-admin/includes/template.php(1443): WPO\WC\PDF_Invoices\Admin->data_input_box_content()
#2 .../wp-admin/edit-form-advanced.php(688): do_meta_boxes()
#3 .../wp-admin/post.php(206): require('/home/u81637615...')
#4 {main}

This seems a particular situation, because the order object should be available.

alexmigf:
closes #625

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -790,7 +790,7 @@ class Admin {
 	}
 
 	public function output_number_date_edit_fields( $document, $data ) {
-		if ( empty( $document ) || empty( $data ) ) {
+		if ( empty( $document ) || empty( $data ) || empty( $document->order ) || ! is_callable( array( $document->order, 'get_id' ) ) ) {
 			return;
 		}
 		$data = $this->get_current_values_for_document( $document, $data );


MohamadNateqi (commented):
Isn't it better if we check the emptiness of `$document->order`?
Since we are using it in the next lines of this function, such as https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a875f7ad5943d67c6cbe565f884b3e12f9cf13b2/includes/class-wcpdf-admin.php#L861

alexmigf:
I'm doing it, see my commit: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/626/commits/df9b1d6e47e4e7f1fd2da86d6c85992a4817a3da

MohamadNateqi (approved):
Sorry, I missed the first condition you added. I just saw the is_callable.

PR #628, opened 2023-10-16
Fix: potential fatal error if document is `false` in order list action buttons

### LINKED ISSUE #627: FatalError for is_allowed()  after Woocommerce Update
Hi,

we love your plugin but today after a woocommerce update to = 8.2.0 2023-10-13 = we've got a fatal error on all plugin related views:

`Fatal error: Uncaught Error: Call to a member function is_allowed() on bool
in /home/customer/www/yogastudiofox.com/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/wcpdf-functions.php on line 48`

seems like the function expects an object but gets a boolean.

any ideas for this?

thanks and cheers

alexmigf:
closes #627


MohamadNateqi (approved):
LGTM!

PR #629, opened 2023-10-17
Some improvements on David remarks

alexmigf:


assets/js/order-script.js
```diff
@@ -1,6 +1,6 @@
 jQuery( function( $ ) {
 
-	$( document.body ).on( 'click', '#doaction, #doaction2', function( e ) {
+	$( '.post-type-shop_order .tablenav' ).on( 'click', '#doaction, #doaction2', function( e ) {
```
DavidAnderson684, line 3: @alexmigf Since that mentions post types, I'm guessing that's not compatible with HPOS?

DavidAnderson684 (commented):
@alexmigf Just one question.

assets/js/order-script.js
alexmigf: Actually it is compatible with HPOS, I have checked that before and tested, but since you mentioned it, I agree that we shouldn't rely on something that could be potentially deleted in the future. Fixed.

DavidAnderson684 (approved):

PR #632, opened 2023-10-20
Fix: bug on UBL preview that forces the invoice creation

### LINKED ISSUE #631: switching between PDF & UBL tabs creates an invoice
A user has reported a bug where switching between to the UBL tab creates an invoice, regardless of the "Attach to" for the PDF invoice or the UBL invoice setting:

Screenshot 2023-10-19 162526

I have tested on the latest version and have found the same issue. You don't need to interact with the UBL settings for the invoice to be created. It also creates invoices when all of the settings except "Enable" is off.

Screenshot 2023-10-19 163044

alexmigf:
closes #631

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -958,6 +958,9 @@ abstract class Order_Document {
 	}
 	
 	public function preview_ubl() {
+		// get last settings
+		$this->settings = ! empty( $this->latest_settings ) ? $this->latest_settings : $this->get_settings( true );
+		
 		return $this->output_ubl( true );
 	}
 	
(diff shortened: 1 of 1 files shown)


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -958,6 +958,9 @@ public function output_html() {
 	}
 	
 	public function preview_ubl() {
+		// get last settings
+		$this->settings = ! empty( $this->latest_settings ) ? $this->latest_settings : $this->get_settings( true );
```
MohamadNateqi, line 962: May I ask why we need to have this line?

MohamadNateqi (approved):
LGTM!
I could reproduce the bug, and this PR has fixed that.

includes/documents/abstract-wcpdf-order-document.php
alexmigf: To display the document with the last settings.

Terminator-Barbapapa (approved):

PR #637, opened 2023-10-26
New: add due date to invoice

### LINKED ISSUE #636: Add due date to invoice

MohamadNateqi:
closes #636


includes/documents/class-wcpdf-invoice.php
```diff
@@ -650,6 +682,27 @@ public function get_date_title() {
...
+
+		$due_days = ( 'custom' === $this->settings['due_date'] ) ? intval( $this->settings['due_date_custom_days'] ?? 0 ) : absint( strtok( $this->settings['due_date'], '_' ) );
+
+		if ( 0 >= $due_days ) {
+			return false;
+		}
+
+		$due_date_datetime = $order->get_date_created()->modify( "+$due_days days" );
+
+		return apply_filters( "wpo_wcpdf_{$this->slug}_invoice_due_date", date( 'Y-m-d', $due_date_datetime->getTimestamp() ), $due_date_datetime->getTimestamp(), $order );
```
alexmigf, line 703: For the display format you should use the `wcpdf_date_format` function.

templates/Simple/invoice.php
```diff
@@ -88,6 +88,12 @@
 					<td><?php echo $payment_method; ?></td>
 				</tr>
 				<?php endif; ?>
+                <?php if ( $due_date = $this->get_due_date( $this->order ) ) : ?>
```
alexmigf, line 91: Maybe we should use the filter below `wpo_wcpdf_after_order_data`? This way a user that has a custom Simple template can take advantage of this new feature without the need to update the template.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -506,6 +506,38 @@ public function get_pdf_settings_fields( $option_name ) {
...
+				'id'			=> 'due_date',
+				'title'			=> __( 'Due date', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback'		=> 'select',
+				'section'		=> $this->type,
+				'args'			=> array(
+					'option_name'	=> $option_name,
+					'id'			=> 'due_date',
+					'options'       => apply_filters( 'wpo_wcpdf_due_date_options', array(
+						// For predefined dates, use this format: "{number_of_days}_day(s)".
+						'0_day'   => __( 'None', 'wpo_wcpdf_pro' ),
```
alexmigf, line 520: Maybe this slugs could be:

- 1
- 7
- 30
- custom

instead? I think we could avoid having a `none` option there doing the same we do for the document date for example: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/88942c4a2667a355ec31e9f1956e72b24ab19227/includes/documents/class-wcpdf-invoice.php#L278

includes/documents/class-wcpdf-invoice.php
```diff
@@ -506,6 +506,38 @@ public function get_pdf_settings_fields( $option_name ) {
...
+						'1_day'   => __( '1 day', 'wpo_wcpdf_pro' ),
+						'7_days'  => __( '7 days', 'wpo_wcpdf_pro' ),
+						'30_days' => __( '30 days', 'wpo_wcpdf_pro' ),
+						'custom'  => __( 'Custom date', 'wpo_wcpdf_pro' ),
```
alexmigf, line 524: Shouldn't be **Custom days**?

includes/documents/class-wcpdf-invoice.php
```diff
@@ -506,6 +506,38 @@ public function get_pdf_settings_fields( $option_name ) {
 					'description'	=> sprintf(__( "Disable document when the order total is %s", 'woocommerce-pdf-invoices-packing-slips' ), function_exists('wc_price') ? wc_price( 0 ) : 0 ),
 				)
 			),
+			array(
```
alexmigf, line 509: Maybe this setting would be better placed below the document date?

MohamadNateqi: Do you mean below the "Display invoice date"?

alexmigf: Yes. What you think?

MohamadNateqi: Sure, it seems better.

templates/Simple/invoice.php
MohamadNateqi: I added this hook and its method in the main class since I think it's proper to have it there. Do you agree?

includes/class-wcpdf-main.php
```diff
@@ -1572,7 +1574,40 @@ public function wc_webhook_topics( $topics ) {
...
+	 * @param \WC_Order|\WC_Order_Refund $order
+	 *
+	 * @return void
+	 */
+	public function display_due_date( string $document_type, $order ): void {
+		if ( 'invoice' !== $document_type ) {
+			return;
+		}
+
+		$invoice_settings = get_option( 'wpo_wcpdf_documents_settings_invoice', array() );
```
alexmigf, line 1589: You have a method to get the document settings: `get_document_settings()`

includes/class-wcpdf-main.php
```diff
@@ -1572,7 +1574,40 @@ public function wc_webhook_topics( $topics ) {
...
+			return;
+		}
+
+		$due_days = ( 'custom' === $invoice_settings['due_date'] ) ? intval( $invoice_settings['due_date_custom_days'] ?? 0 ) : absint( $invoice_settings['due_date'] );
+
+		if ( 0 >= $due_days ) {
+			return;
+		}
+
+		$due_date_datetime = $order->get_date_created()->modify( "+$due_days days" );
```
alexmigf, line 1601: You should check if the order exists.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -333,6 +333,37 @@ public function get_pdf_settings_fields( $option_name ) {
...
+					'option_name'	=> $option_name,
+					'id'			=> 'due_date',
+					'options'       => apply_filters( 'wpo_wcpdf_due_date_options', array(
+						''       => __( 'No', 'wpo_wcpdf_pro' ),
+						'1'      => __( '1 day', 'wpo_wcpdf_pro' ),
+						'7'      => __( '7 days', 'wpo_wcpdf_pro' ),
+						'30'     => __( '30 days', 'wpo_wcpdf_pro' ),
+						'custom' => __( 'Custom days', 'wpo_wcpdf_pro' ),
+					) ),
+					'description'	=> __( 'Select a due date to display in the invoice.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 352: I think we need to be more descriptive about the days and where they are added to.

includes/class-wcpdf-main.php
alexmigf, line 1601: I think we probably need a new option to select which date to use? Because maybe we could also use the Invoice date?

alexmigf (changes requested):

includes/documents/class-wcpdf-invoice.php
MohamadNateqi: What do you think about this sentence:
_Displays a due date below the order data (invoice number, order date, etc.)._

alexmigf: That would help yes, but I was thinking about also indication that the days are added to the selected date type.

MohamadNateqi: Do you think it might be confusing?
Since the options of the due date are in days, I think it implies that the due date will be calculated based on days. And now that we have a base date (recent commits), it might be clearer.

alexmigf: Ok, I agree with that.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -349,13 +349,13 @@ public function get_pdf_settings_fields( $option_name ) {
 						'30'     => __( '30 days', 'wpo_wcpdf_pro' ),
 						'custom' => __( 'Custom days', 'wpo_wcpdf_pro' ),
 					) ),
-					'description'	=> __( 'Select a due date to display in the invoice.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description'	=> __( 'Displays a due date below the order data (invoice number, order date, etc.).', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 352: I think you could avoid the parenthesis content.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -364,6 +364,22 @@ public function get_pdf_settings_fields( $option_name ) {
 					'description'	=> __( 'Enter the number of days for the due date.', 'woocommerce-pdf-invoices-packing-slips' ),
 				)
 			),
+			array(
+				'type'          => 'setting',
+				'id'            => 'due_date_base_date',
+                'title'         => __( 'Due date base date', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 370: I would avoid repeating the word **date**. **Due date base** seems good to me. What you think?

includes/documents/class-wcpdf-invoice.php
```diff
@@ -349,13 +349,13 @@ public function get_pdf_settings_fields( $option_name ) {
...
-					'description'	=> __( 'Select a due date to display in the invoice.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description'	=> __( 'Displays a due date below the order data (invoice number, order date, etc.).', 'woocommerce-pdf-invoices-packing-slips' ),
 				)
 			),
 			array(
 				'type'			=> 'setting',
 				'id'			=> 'due_date_custom_days',
-				'title'			=> __( 'Custom Due Date (Days)', 'woocommerce-pdf-invoices-packing-slips' ),
+				'title'			=> __( 'Custom due date (days)', 'woocommerce-pdf-invoices-packing-slips' ),
 				'callback'		=> 'text_input',
```
alexmigf, line 359: Shouldn't this be a `number` input type?

alexmigf (changes requested):
The due date selector triggers the Preview but nothing happens there. The due date is not being displayed in the Preview.

includes/documents/class-wcpdf-invoice.php
MohamadNateqi: I was just worried if it was not clear enough.

includes/documents/class-wcpdf-invoice.php
MohamadNateqi: Yeah, it seems good.

MohamadNateqi:
> The due date selector triggers the Preview but nothing happens there. The due date is not being displayed in the Preview. Should we avoid it?

The preview works correctly for me and adds the due date. Maybe something else is wrong!
Can you try it again?

alexmigf (commented):
I have tested and seems to be working now, but If you set custom days the preview doesn't work again.

MohamadNateqi:
It was a bug I fixed. It should work correctly now.

alexmigf (approved):
Perfect!

PR #641, opened 2023-11-04
Add label to due date

### LINKED ISSUE #640: Improvements for the due date feature
- Moving the `get_due_date()` method to the abstract document class
- Removing the "base due date" to introduce in the Pro extension
- Adding new filters

MohamadNateqi:
close #640


alexmigf (commented):
Just one remark. You have placed the method `get_due_date()` in the Invoice document class, but the `display_due_date()` on the main class. Since this is restricted to the Invoice, don't you think they should both be in the Invoice class?

MohamadNateqi:
It seems that the `invoice` class is being called multiple times. If we were to move the `display_due_date()` function to this class, it would add a due date to the invoice every time it's called.
Initially, I tried this approach, but I noticed that multiple due dates were being added to the invoice, which made me realize that the class was being called multiple times. So, I moved the function back to the main class.

alexmigf (commented):
In that case I think we should make the function more flexible:

* adding a filter in the document type check
* using `wcpdf_get_document` instead of `wcpdf_get_invoice`
* add a callable check under `get_due_date`, just in case
* Renaming this hook: `wpo_wcpdf_{document_slug}_due_date_display`

MohamadNateqi:
> In that case I think we should make the function display_due_date a bit more flexible:

To make it more general purpose instead of just working for the `invoice` document. Got it.

includes/class-wcpdf-main.php
```diff
@@ -1582,25 +1582,31 @@ public function wc_webhook_trigger( $document, $order ) {
 	 * @return void
 	 */
 	public function display_due_date( string $document_type, $order ): void {
-		if ( 'invoice' !== $document_type || empty( $order ) ) {
+		if ( empty( $order ) ) {
 			return;
 		}
 
-		$invoice = wcpdf_get_invoice( $order );
+		$allowed_document_types = apply_filters( 'wpo_wcpdf_due_date_allowed_document_types', array( 'invoice' ), $order, $document_type );
```
alexmigf, line 1589: Looking into this, maybe we could avoid this filter only by checking if the `get_due_date` is callable?

MohamadNateqi: Or we can make it possible to get their custom function name, and we call that instead. And then it's ok to keep the filter.
What do you think?

alexmigf: I'm not sure if I understand.

MohamadNateqi: I meant something like this:
```php
		$due_date_timestamp = 0;
		$due_date_function  = apply_filters( 'wpo_wcpdf_getting_due_date_function_name', '', $order, $document_type );

		if ( ! empty( $due_date_function ) ) {
			if ( is_callable( $due_date_function ) ) {
				$due_date_timestamp = call_user_func( $due_date_function );
			}
		} else {
			if ( is_callable( array( $document, 'get_due_date' ) ) ) {
				$due_date_timestamp = $document->get_due_date();
			}
		}
```

With this, we can create a callable function for other documents with a snippet if users request.

alexmigf: Why we would use custom functions? Do you have a scenario when that could happen? I think if we add to the abstract class, we could use in every document if the setting exist there right?

MohamadNateqi: The only scenario that came to my mind is when a user wants to add the due date for other types of documents for any reason. In that case, allowing to call a custom function can easily make it possible.

> Maybe we could add to the abstract document class and use in the documents we want when the setting exists?

Do you think we will add the due date for other documents? I think it'll be only for the invoice, so it may not be necessary to add it there. Not sure!

alexmigf: > Do you think we will add the due date for other documents? I think it'll be only for the invoice, so it may not be necessary to add it there. Not sure!

I can see this being used in the Proposal or similar documents like Quotes.

MohamadNateqi: Ok then, I'll add the `get_due_date()` to the abstract document class and remove `$allowed_document_types`.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -1364,6 +1364,38 @@ public function get_number_store_year( $table_name ) {
...
+		}
+
+		$due_date_days = ( 'custom' === $this->settings['due_date'] ) ? intval( $this->settings['due_date_custom_days'] ?? 0 ) : absint( $this->settings['due_date'] );
+		$due_date_days = apply_filters( "wpo_wcpdf_{$this->slug}_due_date_days", $due_date_days, $this );
+
+		if ( 0 >= $due_date_days ) {
+			return 0;
+		}
+
+		if ( isset( $this->settings['due_date_base_date'] ) && 'invoice_date' === $this->settings['due_date_base_date'] && $this->exists() ) {
```
alexmigf, line 1384: Maybe we could use the document type here? `"{$this->get-type()}_date"`

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -1364,6 +1364,38 @@ public function get_number_store_year( $table_name ) {
...
+			// The `get_date()` method has been used to get the date as a WC_DateTime object.
+			$base_date = $this->get_date( 'invoice' );
```
alexmigf, line 1386: And also pass the document type here.

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document.php
MohamadNateqi: Ah, missed that.
I think changing the key to "document_date" in settings would be better.
What do you think?

alexmigf: Yes, I agree.

includes/class-wcpdf-main.php
```diff
@@ -1582,29 +1582,31 @@ public function wc_webhook_trigger( $document, $order ) {
...
+		$due_date_timestamp = is_callable( array( $document, 'get_due_date' ) ) ? $document->get_due_date() : 0;
 
 		if ( 0 >= $due_date_timestamp ) {
 			return;
 		}
 
-		$due_date = apply_filters( 'wpo_wcpdf_invoice_due_date_display', date( wcpdf_date_format( $this, 'due_date' ), $due_date_timestamp ), $due_date_timestamp, $order );
+		$due_date          = apply_filters( "wpo_wcpdf_{$document->slug}_due_date_display", date( wcpdf_date_format( $this, 'due_date' ), $due_date_timestamp ), $due_date_timestamp, $order, $document_type );
+		$document_settings = $document->get_settings();
+		$due_date_label    = ! empty( $document_settings['due_date_label'] ) ? $document_settings['due_date_label'] : __( 'Due Date:', 'woocommerce-pdf-invoices-packing-slips' );
```
alexmigf, line 1603: I think that the `$document_settings['due_date_label']` must be able to be translated

includes/documents/class-wcpdf-invoice.php
```diff
@@ -365,6 +363,18 @@ public function get_pdf_settings_fields( $option_name ) {
...
+			array(
+				'type'          => 'setting',
+				'id'            => 'due_date_label',
+				'title'         => 'Due date label',
+				'callback'      => 'text_input',
+				'section'       => $this->type,
+				'args'          => array(
+					'option_name' => $option_name,
+					'id'          => 'due_date_label',
+					'placeholder' => 'Due date:'
```
alexmigf, line 375: Placeholder must be able to be translated

includes/documents/class-wcpdf-invoice.php
```diff
@@ -376,7 +386,7 @@ public function get_pdf_settings_fields( $option_name ) {
 	                'id'            => 'due_date_base_date',
                     'options'       => apply_filters( 'wpo_wcpdf_due_date_base_date_options', array(
```
alexmigf, line 387: Maybe you should include the document slug in this filter?

alexmigf (changes requested):

includes/class-wcpdf-main.php
MohamadNateqi: I'm not sure I understood. Do you mean the 'Due Date:' string, which is already translatable?

alexmigf: No, I mean the one that you fill in the field.

MohamadNateqi: @alexmigf If I understood the implementation correctly, the other part of the translation should be in the pro extension, in the `filter_document_titles`(). Am I right?

includes/class-wcpdf-main.php
```diff
@@ -1598,13 +1599,11 @@ public function display_due_date( string $document_type, $order ): void {
...
 
-		$due_date          = apply_filters( "wpo_wcpdf_{$document->slug}_due_date_display", date( wcpdf_date_format( $this, 'due_date' ), $due_date_timestamp ), $due_date_timestamp, $order, $document_type );
-		$document_settings = $document->get_settings();
-		$due_date_label    = ! empty( $document_settings['due_date_label'] ) ? $document_settings['due_date_label'] : __( 'Due Date:', 'woocommerce-pdf-invoices-packing-slips' );
+		$due_date = apply_filters( "wpo_wcpdf_{$document->slug}_due_date_display", date( wcpdf_date_format( $this, 'due_date' ), $due_date_timestamp ), $due_date_timestamp, $order, $document_type );
 
 		if ( ! empty( $due_date ) ) {
 			echo '<tr class="due-date">
-				<th>', $due_date_label, '</th>
+				<th>', $document->get_due_date_title(), '</th>
```
alexmigf, line 1606: Maybe you should check if this is callable?

alexmigf (changes requested):

includes/class-wcpdf-main.php
alexmigf: I think you just need to wrap the saved value under `__()`? But I haven't tested.

MohamadNateqi: Do you mean I should have to wrap `$document_settings['due_date_label']` under `__()`?

By the way, I added the below code in the `filter_document_titles()` method in the pro extension and tested it with Polylang. It was working correctly:
```php
add_filter( "wpo_wcpdf_{$_document->slug}_due_date_title", function ( $title, $document = null ) use ( $_document ) {
	return $this->get_custom_title_for_filter( 'due_date_title', $title, $document, $_document );
}, 10, 2 );
```

alexmigf: Hum, maybe the label field should be added only in Pro like the others, and in the free we use the filter. Now that you mentioned the `filter_document_titles()` I saw that we have all title fields in Pro only. Sorry for the trouble :(

MohamadNateqi:
@alexmigf In the latest commit, I also updated the hook name for filters. Instead of adding the document slug in the hook name, I send the document type as an argument to make them uniform. It seemed there was no need to have different hook names for them.
Please let me know if it's not a good idea, and I'll revert that.

MohamadNateqi:
@alexmigf Ready to review.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -1379,10 +1379,9 @@ public function get_due_date(): int {
 			return 0;
 		}
 
-		$due_date_days = ( 'custom' === $this->settings['due_date'] ) ? intval( $this->settings['due_date_custom_days'] ?? 0 ) : absint( $this->settings['due_date'] );
-		$due_date_days = apply_filters( "wpo_wcpdf_due_date_days", $due_date_days, $this->type, $this );
+		$due_date_days = apply_filters( "wpo_wcpdf_due_date_days", $this->settings['due_date'], $this->type, $this );
```
alexmigf, line 1382: Maybe single quotes on the filter name?

includes/documents/class-wcpdf-invoice.php
```diff
@@ -345,24 +345,10 @@ public function get_pdf_settings_fields( $option_name ) {
 						'1'      => __( '1 day', 'wpo_wcpdf_pro' ),
```
alexmigf, line 345: Maybe you should also provide the document type on this options filter.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -345,24 +345,10 @@ public function get_pdf_settings_fields( $option_name ) {
...
 						'7'      => __( '7 days', 'wpo_wcpdf_pro' ),
 						'30'     => __( '30 days', 'wpo_wcpdf_pro' ),
-						'custom' => __( 'Custom days', 'wpo_wcpdf_pro' ),
 					) ),
 					'description'	=> __( 'Displays a due date below the order data.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 349: WPCS `=>` alignment

includes/documents/class-wcpdf-invoice.php
```diff
@@ -345,24 +345,10 @@ public function get_pdf_settings_fields( $option_name ) {
 						'1'      => __( '1 day', 'wpo_wcpdf_pro' ),
 						'7'      => __( '7 days', 'wpo_wcpdf_pro' ),
```
alexmigf, line 346: Wrong textdomain.

alexmigf (changes requested):

alexmigf (approved):
Perfect!

PR #645, opened 2023-11-06
Fix: bug on historical settings not being deleted when using most current settings

### LINKED ISSUE #643: Historical settings not being deleted when using current settings
Historical settings not being deleted when using most current settings. This happens when using **Test Mode** or **Most Current Settings** (document setting) and then revert back to historical, which are still there to be used.

alexmigf:
closes #643

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -140,6 +140,11 @@ abstract class Order_Document {
 		$this->settings        = $this->get_settings();
 		$this->latest_settings = $this->get_settings( true );
 		
+		// don't save settings on Preview requests
+		if ( isset( $_REQUEST['action'] ) && 'wpo_wcpdf_preview' === $_REQUEST['action'] ) {
+			return;
+		}
+		
 		// save settings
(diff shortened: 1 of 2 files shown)


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -134,11 +134,20 @@ public function init_settings() {
...
+		$this->order_settings  = $this->get_order_settings();
 		// pdf
-		$this->settings            = $this->get_settings();
-		$this->latest_settings     = $this->get_settings( true );
+		$this->settings        = $this->get_settings();
+		$this->latest_settings = $this->get_settings( true );
+		
+		// save settings
+		$latest = $this->use_historical_settings() ? false : true;
+		$this->save_settings( $latest );
```
MohamadNateqi, line 150: I think it would be good to directly use `$this->use_historical_settings()` instead of `$latest`.

MohamadNateqi (approved):
LGTM!

MohamadNateqi (approved):

Terminator-Barbapapa (approved):
LGTM

PR #648, opened 2023-11-09
Allowing multiple document generation with "Full" access type

### LINKED ISSUE #647: Allow multiple document generation with "Full" access type
When the "Document link access type" is FULL, creating multiple documents raises permission error — reported [here](https://secure.helpscout.net/conversation/2415089075/55653?folderId=221726).

MohamadNateqi:
closes #647

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -444,8 +444,15 @@ class Main {
 				}
 				break;
 			case 'full':
-				// no order
-				if ( ! $order ) {
+				// check the user privileges
+				$full_permission = WPO_WCPDF()->admin->user_can_manage_document( $document_type );
+
+				if ( $full_permission ) {
+                    break;
+                }
+
+				// no order, single order only
+				if ( ! $order || count( $order_ids ) > 1 ) {
 					$allowed = false;
 					break;
 				}
@@ -455,12 +462,7 @@ class Main {
 					$allowed = false;
 					break;
 				}
-				
-				// single order only
-				if ( count( $order_ids ) > 1 ) {
-					$allowed = false;
-					break;
-				}
+
 				break;
 		}
 


includes/class-wcpdf-main.php
```diff
@@ -444,8 +444,15 @@ public function generate_document_ajax() {
 				}
 				break;
 			case 'full':
-				// no order
-				if ( ! $order ) {
+				// check the user privileges
+				$full_permission = WPO_WCPDF()->admin->user_can_manage_document( $document_type );
```
alexmigf, line 448: I'm not sure if this is correct, because the `user_can_manage_document` checks for this:

`current_user_can( 'manage_woocommerce_orders' ) || current_user_can( 'edit_shop_orders' )`

which means that `guest` access will not work here.

The full intent is to allow anyone with the link to have access, I think this way is more restrictive.

I think the issue is related to the Bulk Action not passing the access key in the URL.

MohamadNateqi: > which means that guest access will not work here.

May I ask why it won't work? I'm a little confused.
Since the condition I added will break the `case` if there is full permission. If the access type is "guest" and the user is not logged in, the `$full_permission` is also `false`, so it won't break the `case` and should work as before.

> I think the issue is related to the Bulk Action not passing the access key in the URL.

Yes, the access key doesn't exist in the URL, but `_wpnonce` exists, and this block of code fills the `access_key`:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/21e59a7a35c23d700870da9bd17c7820370e12e9/includes/class-wcpdf-main.php#L330

After that, it gets to this block of code. It won't execute for bulk action, and the `$order` will always be `false`:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/21e59a7a35c23d700870da9bd17c7820370e12e9/includes/class-wcpdf-main.php#L383

So I think the issue was the condition in the full case that checked if the `$order` is not true (`if ( ! $order )`).
Am I right?

alexmigf: I think you're right, I haven't check the `access_key` part in the code, didn't recall that existed, and ignored all the rest because without the access key the below code doesn't execute, sorry.

includes/class-wcpdf-main.php
```diff
@@ -444,8 +444,15 @@ public function generate_document_ajax() {
...
+
+				if ( $full_permission ) {
```
alexmigf, line 450: I think you should follow the same procedure as the other checks, by setting the `$allowed = false;`.

includes/class-wcpdf-main.php
```diff
@@ -444,8 +444,15 @@ public function generate_document_ajax() {
...
+                    break;
```
alexmigf, line 451: I think you have a problem with indentation in your changes, must be tab with 4 spaces

alexmigf (commented):
I think we need to change this comment text: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/25da3535d5c408614fabc0ff9158ce10734109e4/includes/class-wcpdf-main.php#L329

Because we could potential remove it thinking it's legacy, when 1 key still being used.

alexmigf (commented):
I still think we need to do this differently, because if you check the permissions like you're doing, this will break in every request, and we just want to break when it's a multi-order request (Bulk Actions for example).

includes/class-wcpdf-main.php
MohamadNateqi: This is because of my IDE! I don't know what is wrong with it!
I hit `enter` to go to the next line, and sometimes it creates it with spaces instead of tabs! Or, I press the tab, and it replaces it with spaces! Weird!

MohamadNateqi:
I have made some modifications that I think address the concerns you raised.

MohamadNateqi:
Should it be allowed to see multiple order documents with a "Full" access type?

alexmigf:
> Should it be allowed to see multiple order documents with a "Full" access type?

Maybe we should restrict that with the permissions check.

alexmigf:
I don't think that will work for guest users?

MohamadNateqi:
In [this commit](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/648/commits/05beacc57f75beda91d77384ecf84629f3b523ab), it has to work for guest type too since I pulled that condition out of the switch.

alexmigf:
What happens if you are a guest user, no account created, and you arrive here? https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/648/commits/05beacc57f75beda91d77384ecf84629f3b523ab#diff-e3b7726be07c18694e3393ba2ae22255e14266cd09f4cf13ff28d9064173a133R415

If I have no account this will return `false` right? `( current_user_can( 'manage_woocommerce_orders' ) || current_user_can( 'edit_shop_orders' )`

MohamadNateqi:
> If I have no account this will return false right?

Yes. It returns false. So you mean we let guest users to access multi-order documents, yes?

alexmigf:
No, I mean we should let guest users access only single documents. If the user pass the permissions check it can also access bulk documents.

MohamadNateqi:
Doesn't this code stop guest users from accessing multi-order docs?
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/05beacc57f75beda91d77384ecf84629f3b523ab/includes/class-wcpdf-main.php#L418

As you mentioned `$full_permission` is `false` so the `! $full_permission` of a guest user would be `true` here, and this will stop access to bulk documents.

includes/class-wcpdf-main.php
```diff
@@ -406,58 +405,46 @@ public function generate_document_ajax() {
 
 		// set default is allowed
 		$allowed = true;
-		
+
+		// no order when it is a single order
+		if ( ! $order && count( $order_ids ) === 1 ) {
```
alexmigf, line 410: WPCS: `1 === ...`

MohamadNateqi:
> I have missed the exclamation mark somehow. I will review again.

I missed one last night, and for half an hour, I was asking why the condition was not working correctly! Happily I finally saw that otherwise I could have had a nightmare! 😂

alexmigf:
It happens to me sometimes, specially when I'm doing other code work which I pause to review a different code, I think my mindset it's still in the previous code 😂

MohamadNateqi:
Same here! 😂

alexmigf (changes requested):
Maybe you could also improve this? https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/648#pullrequestreview-1722301941

MohamadNateqi:
> Maybe you could also improve this? https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/648#pullrequestreview-1722301941

What do you think about it: `// handle the bulk order access key`

alexmigf:
Seems good to me.

alexmigf:
Maybe we don't need to keep the `order_key`, I don't think it's used anymore.

MohamadNateqi:
> Maybe we don't need to keep the order_key, I don't think it's used anymore.

There may be usage we are unaware of, but we can remove it and see if any customer reports any issues.

alexmigf:
I think we used before we change to the `access_key`, which before was 2 different keys, one for nonce and other for the order key.

includes/class-wcpdf-main.php
```diff
@@ -326,13 +326,9 @@ public function generate_document_ajax() {
...
 		$redirect_url = WPO_WCPDF()->endpoint->get_document_denied_frontend_redirect_url();
 
-		// handle legacy access keys
-		if ( empty( $_REQUEST['access_key'] ) ) {
-			foreach ( array( '_wpnonce', 'order_key' ) as $legacy_key ) {
-				if ( ! empty( $_REQUEST[ $legacy_key ] ) ) {
-					$_REQUEST['access_key'] = sanitize_text_field( $_REQUEST[ $legacy_key ] );
-				}
-			}
+		// handle the bulk order access key
```
alexmigf, line 329: In this case you can change to: `// handle bulk actions access key`

alexmigf (changes requested):

alexmigf (approved):
Perfect!

PR #652, opened 2023-11-10
Fix: revert handling access keys

MohamadNateqi:
@alexmigf I was not sure about the comment. Let me know please If you have a better suggestion.


alexmigf (approved):
Perfect!

Issue #659, opened 2023-11-16
Fatal error: Call to undefined method WPO_WCPDF_Receipt_Document::number()


alexmigf:
A [customer](https://secure.helpscout.net/conversation/2421912617/55822?folderId=221726) reported the error below because the template is calling a function that is only available in the Professional extension:

> Fatal error: Call to undefined method WPO_WCPDF_Receipt_Document::number()

This error should also happen for the document date function as well.

Of course, the Professional extension is required for the Receipt document, but maybe we should uniform this number/date callings in the templates.

Maybe we should migrate the functions to the free plugin, and maybe update in the templates.

PR #660, opened 2023-11-16
Tweak: improves the templates number/date functions

alexmigf:
#659

This PR improves the templates display functions:

New functions:
```php
$this->number( $document_type );
$this->date( $document_type );
$this->number_title();
$this->date_title();
$this->due_date_title();
``` 

Deprecated functions:
`get_invoice_number`
`invoice_number`
`get_invoice_date`
`invoice_date`

Changed functions:
`get_number`: now includes `$formatted` `bool` arg
`get_date`: now includes `$formatted` `bool` arg

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -1167,7 +1167,7 @@ abstract class Order_Document_Methods extends Order_Document {
 	public function get_invoice_number() {
 		wcpdf_log_error( 'The method get_invoice_number() is deprecated since version 3.7.3, please use the method get_number() instead.' );
 		
-		if ( is_callable( array( $this, 'get_formatted_number' ) ) ) {
+		if ( is_callable( array( $this, 'get_number' ) ) ) {
 			return $this->get_number( 'invoice', null, 'view', true );
 		} else {
 			return '';
@@ -1195,7 +1195,7 @@ abstract class Order_Document_Methods extends Order_Document {
 	public function get_invoice_date() {
 		wcpdf_log_error( 'The method get_invoice_date() is deprecated since version 3.7.3, please use the method get_date() instead.' );
 		
-		if ( is_callable( array( $this, 'get_formatted_date' ) ) ) {
+		if ( is_callable( array( $this, 'get_date' ) ) ) {
 			return $this->get_date( 'invoice', null, 'view', true );
 		} else {
 			return '';
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -1160,6 +1160,10 @@ abstract class Order_Document_Methods extends Order_Document {
 		return $item_price;
 	}
 
+	/**
+	 * Legacy function (v3.7.2 or inferior)
+	 * Use $this->get_number() or $this->get_formatted_number() instead.
+	 */
 	public function get_invoice_number() {
 		// Call the woocommerce_invoice_number filter and let third-party plugins set a number.
 		// Default is null, so we can detect whether a plugin has set the invoice number
@@ -1175,10 +1179,18 @@ abstract class Order_Document_Methods extends Order_Document {
 		}
 	}
 
+	/**
+	 * Legacy function (v3.7.2 or inferior)
+	 * Use $this->number( 'invoice' ) instead.
+	 */
 	public function invoice_number() {
 		echo $this->get_invoice_number();
 	}
 
+	/**
+	 * Legacy function (v3.7.2 or inferior)
+	 * Use $this->get_date() or $this->get_formatted_date() instead.
+	 */
 	public function get_invoice_date() {
 		if ( $invoice_date = $this->get_date('invoice') ) {
 			return $invoice_date->date_i18n( wcpdf_date_format( $this, 'invoice_date' ) );
@@ -1187,6 +1199,10 @@ abstract class Order_Document_Methods extends Order_Document {
 		}
 	}
 
+	/**
+	 * Legacy function (v3.7.2 or inferior)
+	 * Use $this->date( 'invoice' ) instead.
+	 */
 	public function invoice_date() {
 		echo $this->get_invoice_date();
 	}
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -1162,18 +1162,13 @@ abstract class Order_Document_Methods extends Order_Document {
 
 	/**
 	 * Legacy function (v3.7.2 or inferior)
-	 * Use $this->get_number() or $this->get_formatted_number() instead.
+	 * Use $this->get_formatted_number() instead.
 	 */
 	public function get_invoice_number() {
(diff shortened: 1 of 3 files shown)


includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1160,6 +1160,10 @@ public function get_formatted_item_price ( $item, $type, $tax_display = '' ) {
 		return $item_price;
 	}
 
+	/**
+	 * Legacy function (v3.7.2 or inferior)
```
MohamadNateqi, line 1164: Can we have `_deprecated_function()` for these legacy functions? And maybe calling the new functions like the `$this->number( $document_type );` instead of their current codes?

includes/documents/abstract-wcpdf-order-document.php
MohamadNateqi, line 1: The quotation document (`WPO_WCPDF_Quotation_Document` class) has the `get_formatted_number()` and the `get_formatted_date()` methods, but without the `$document_type` argument. And since adding those new functions makes them an override, it will raise this warning:
`PHP Warning:  Declaration of WPO_WCPDF_Quotation_Document::get_formatted_number() should be compatible with WPO\WC\PDF_Invoices\Documents\Order_Document::get_formatted_number($document_type) in C:\laragon\www\mohamad\wp-content\plugins\wcpdf-quotation\class-wcpdf-quotation-document.php on line 164`

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -494,10 +494,38 @@ public function get_data( $key, $document_type = '', $order = null, $context = '
 	public function get_number( $document_type = '', $order = null, $context = 'view'  ) {
 		return $this->get_data( 'number', $document_type, $order, $context );
 	}
+	
+	public function get_formatted_number( $document_type ) {
```
MohamadNateqi, line 498: Would it be good to apply a filter on the return value of these new functions?

includes/documents/abstract-wcpdf-order-document.php
alexmigf: If we change this, the Pro extension should trigger the same error because it declares the function with the document type arg. What do you think of changing all the methods that `echo` values prefixing with `display_`?

`$this->number()` => `$this->display_number()`

This will keep both functions in place for backwards compatibility, at least for some versions.

I think adding the `display_` helps to understand better what they do as well.

Let me know your thoughts.

MohamadNateqi: I like the idea, it's very good!
Adding the 'display_' prefix is a much better naming convention. And also, adding the `$formatted` arg makes it simpler.

The only issue with this new naming convention I can think of is that we have several other functions named like that, such as `header_logo` and `shop_name`. Changing them can create compatibility issues; not changing them would create two groups of naming (The best option here, I think); and duplicating them with the new name, I am not sure we can remove them soon.

What do you think of keeping the function name the same (like the `number()`) and merging the `get_formatted_number` method with it? Is it necessary to have that logic in another method?

alexmigf: The `number()` function has the same problem I think, because it requires the `$document_type` as well. So the affecting functions currently are:

`get_formatted_number`
`get_formatted_date`
`number`
`date`

alexmigf: To be honest I don't think we need the `formatted` functions, but on the other side the `number` and `date` will trigger the same error after passing the `formatted` ones.

> Changing them can create compatibility issues; not changing them would create two groups of naming (The best option here, I think); and duplicating them with the new name, I am not sure we can remove them soon.

A: Indeed, we need to keep both names for a while, to avoid compatibility issues on custom templates. I think we could create maybe a page under the "Advanced" tab to alert for legacy functions/hooks present in the templates, so they can fix before they go away.

MohamadNateqi: I saw that there is just one override for both the `number` and the `date` method, which is in the `Pro_Document` class, and both include a `$document_type` arg. So, would adding them to the `Order_Document` class create any issues?

> I think we could create maybe a page under the "Advanced" tab to alert for legacy functions/hooks present in the templates, so they can fix before they go away.

Based on the support emails after removing legacy support, I feel most of the customers don't know how to fix their template, they either don't have enough coding knowledge, or someone else did that for them. So I think just a few customers will use that, and we still need to fix it ourselves. Of course, I like the idea of having this page, and it's useful, but what if we add it to the website documents?

alexmigf: > So, would adding them to the `Order_Document` class create any issues?

A: Only in the Quotation extension, but maybe we should go that way and fix the extension. If a user reports the issue we send them the extension updated.

MohamadNateqi: > Only in the Quotation extension

Sorry I couldn't understand the issue. If we only add those two methods to the `Order_Document`, what makes an issue in this extension? Since I couldn't find the `number` and `date` methods in this extension.

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1160,35 +1160,60 @@ public function get_formatted_item_price ( $item, $type, $tax_display = '' ) {
...
-		$third_party_invoice_number = apply_filters( 'woocommerce_invoice_number', null, $this->order_id );
-		if ($third_party_invoice_number !== null) {
-			return $third_party_invoice_number;
-		}
-
-		if ( $invoice_number = $this->get_number('invoice') ) {
-			return $formatted_invoice_number = $invoice_number->get_formatted();
+		wcpdf_log_error( 'The method get_invoice_number() is deprecated since version 3.7.3, please use the method get_number() instead.' );
+		
+		if ( is_callable( array( $this, 'get_formatted_number' ) ) ) {
```
MohamadNateqi, line 1170: `get_formatted_number` -> `get_number`

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1160,35 +1160,60 @@ public function get_formatted_item_price ( $item, $type, $tax_display = '' ) {
...
+	/**
+	 * Legacy function (v3.7.2 or inferior)
+	 * Use $this->get_date() instead.
+	 */
 	public function get_invoice_date() {
-		if ( $invoice_date = $this->get_date('invoice') ) {
-			return $invoice_date->date_i18n( wcpdf_date_format( $this, 'invoice_date' ) );
+		wcpdf_log_error( 'The method get_invoice_date() is deprecated since version 3.7.3, please use the method get_date() instead.' );
+		
+		if ( is_callable( array( $this, 'get_formatted_date' ) ) ) {
```
MohamadNateqi, line 1198: `get_formatted_date` -> `get_date`

MohamadNateqi (changes requested):

MohamadNateqi (approved):
LGTM!

MohamadNateqi:
@alexmigf Do you have any idea how the customer could raise this error? Since it seems they didn't have the Pro extension.

alexmigf:
Not yet, it's still in discussion in the ticket.

MohamadNateqi:
> I think we should go forward with this improvements which simplifies the number and date document functions. What you think?

Totally agree. It's a good improvement.

PR #662, opened 2023-11-22
Fix PHP 8.1 preg_replace warning

MohamadNateqi:
`Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /wp-includes/kses.php on line 1744
`


alexmigf (approved):
LGTM!

PR #664, opened 2023-11-25
Disable packing slip attachment

### LINKED ISSUE #644: Packing slips are attached to emails based on prior settings in the Pro extension
Normally, the free version does not allow for the packing slip to be attached to emails, on the invoice. However, if the Professional extension is activated with a setting to attach the packing slip, and then the Professional extension is deactivated, the packing slip will be sent using those settings that were stored by the Professional extension.

We should check if the Professional extension is active before a packing slip is attached to an email notification.

dwalkerpriv:
closes #644 

We recently found a 9 year old bug where the packing slips are attached to the emails in the free version, when only the invoices are attached to the emails in the free version. The bug is cause by a user enabling an option in the "Attach to" setting from the Professional extension. If the Professional extension is deactivated, that "Attach to" email setting is still saved and the packing slip is thus attached to the email.

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -133,6 +133,11 @@ class Main {
 		$attach_to_document_types = $this->get_documents_for_email( $email_id, $order );
 		foreach ( $attach_to_document_types as $output_format => $document_types ) {
 			foreach ( $document_types as $document_type ) {
+				//If Professional extension is not active, only attach the invoice
+				if ( ! function_exists( 'WPO_WCPDF_Pro' ) && $document_type !== 'invoice' ) {
+					return $attachments;
+				};
+				
 				$email_order    = apply_filters( 'wpo_wcpdf_email_attachment_order', $order, $email, $document_type );
 				$email_order_id = $email_order->get_id();
 


YordanSoares (approved):
LGTM! 🚀

includes/class-wcpdf-main.php
```diff
@@ -133,6 +133,11 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
 		$attach_to_document_types = $this->get_documents_for_email( $email_id, $order );
 		foreach ( $attach_to_document_types as $output_format => $document_types ) {
 			foreach ( $document_types as $document_type ) {
+				//If Professional extension is not active, only attach the invoice
+				if ( ! function_exists( 'WPO_WCPDF_Pro' ) && $document_type !== 'invoice' ) {
```
alexmigf, line 137: WPCS:

```php
'invoice' !== $document_type
```

includes/class-wcpdf-main.php
```diff
@@ -133,6 +133,11 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
...
+					return $attachments;
```
alexmigf, line 138: I think it should be `continue` here.

alexmigf (changes requested):

MohamadNateqi (approved):
LGTM!

PR #669, opened 2023-12-05
Fix: Invoice number race condition

### LINKED ISSUE #619: Avoid double invoice creation on race condition related with Stripe payments
Some users have reported race conditions on orders paid with Stripe, which create two PDF invoices for the same order (the first is overwritten by the second).

See: 
- https://wordpress.org/support/topic/missing-4-invoices-95-99-102-and-139/#post-17098468
- https://wordpress.org/support/topic/duplicate-invoice-ids-are-generated/#post-17049280

alexmigf:
closes #619 

This PR attempts to fix the race condition issue spotted using Stripe plugin.

It changes the lock time to `60` seconds, to avoid expiring before the code being finished, and now also adds a lock to the email attachment function as well, which is the primary way of getting duplicated numbers.

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -184,7 +184,7 @@ class Main {
 		if ( $lock->lock( $this->lock_retries ) ) {
 			
 			if ( $this->lock_log_enabled ) {
-				$lock->log( sprinf( 'Lock acquired for attach document to email for order ID# %s.', $order_id ), 'info' );
+				$lock->log( sprintf( 'Lock acquired for attach document to email for order ID# %s.', $order_id ), 'info' );
 			}
 		
 			foreach ( $attach_to_document_types as $output_format => $document_types ) {
@@ -258,12 +258,12 @@ class Main {
 			$lock_release = $lock->release();
 			
 			if ( $lock_release && $this->lock_log_enabled ) {
-				$lock->log( sprinf( 'Lock released for attach document to email for order ID# %s.', $order_id ), 'info' );
+				$lock->log( sprintf( 'Lock released for attach document to email for order ID# %s.', $order_id ), 'info' );
 			}
 		
 		} else {
 			if ( $this->lock_log_enabled ) {
-				$lock->log( sprinf( 'Couldn\'t get the lock for attach document to email for order ID# %s.', $order_id ), 'critical' );
+				$lock->log( sprintf( 'Couldn\'t get the lock for attach document to email for order ID# %s.', $order_id ), 'critical' );
 			}
 		}
 
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -33,7 +33,7 @@ class Main {
 		// semaphore
 		$this->lock_name    = 'wpo_wcpdf_main_lock';
 		$this->lock_context = array( 'source' => 'wpo-wcpdf-main' );
-		$this->lock_time    = apply_filters( 'wpo_wcpdf_main_lock_time', 300 );
+		$this->lock_time    = apply_filters( 'wpo_wcpdf_main_lock_time', 60 );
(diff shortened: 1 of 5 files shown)


includes/class-wcpdf-main.php
```diff
@@ -131,70 +143,80 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
 		}
 
 		$attach_to_document_types = $this->get_documents_for_email( $email_id, $order );
+		$lock_name                = $this->lock_name . '_attach_document_to_email_' . $order_id;
```
MohamadNateqi, line 146: Does it also need to contain the document type and maybe the output type?

alexmigf: I added the order number to avoid having concurrent orders being processed at the same time, for example, if 2 customers are buying at the exact same time.

I don't think that the document type or the output type helps here.

alexmigf: > Also, Is it better to put the if above the foreach?

Yes, maybe it's better.

alexmigf:
@MohamadNateqi check now

MohamadNateqi (changes requested):

includes/class-wcpdf-main.php
alexmigf: Regarding the first comment [here](https://wordpress.org/support/topic/missing-4-invoices-95-99-102-and-139/), I think we should avoid that.

MohamadNateqi: Can you help me understand something here? It seems a little confusing to me.

Consider, for example, ten requests at the same time (that calls the `attach_document_to_email` method ten times). One of them passes through the `if ( $lock->lock( $this->lock_retries ) ) {`, for the other nine, since the request won't freeze here until the lock gets released, then all of them fail. Am I right?

alexmigf: Well, only if there are many at the same time and the process takes too much, which seems unlikely no?

includes/class-wcpdf-main.php
```diff
@@ -131,72 +171,82 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
...
-					wcpdf_log_error( $e->getMessage(), 'critical', $e );
-					continue;
+						
 				}
 			}
+			
+			$lock->release();
+		
+		} else {
+			$lock->log( sprintf( 'Couldn\'t get the lock for the order ID# %s!', $email_order_id ), 'critical' );
```
MohamadNateqi, line 249: `$email_order_id` -> `$order_id`

includes/class-wcpdf-main.php
MohamadNateqi: > which seems unlikely no?

Yes, I think it hardly can happen. Maybe only for big stores with lots of orders, and still unlikely.
Thanks for the explanation.

MohamadNateqi (approved):
LGTM!

includes/class-wcpdf-main.php
alexmigf: Yes, my idea is avoiding having 2 or more orders processing the attachments at the same time to avoid concurrent document creations, which should avoid issues with the numbers and assignments.

MohamadNateqi (changes requested):

MohamadNateqi (approved):
All good!

MohamadNateqi (commented):
The last commit was great.

What do you think of checking that setting here and sending an empty array if it's not checked, instead of adding that in the Semaphore class? Then the `log()` method can only check the emptiness of that property.
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/842bc343f720fcb6fb6949cb35f8edfc3a66193f/includes/class-wcpdf-main.php#L174

MohamadNateqi (approved):
Perfect!

PR #674, opened 2023-12-06
Tweak: re-adds the legacy hook `woocommerce_invoice_number` for backwards compatibility

### LINKED ISSUE #673: Re-add legacy hook `woocommerce_invoice_number` for backwards compatibility
Re-add legacy hook `woocommerce_invoice_number` for backwards compatibility.

Forum thread: https://wordpress.org/support/topic/missing-woocommerce_invoice_number-hook/

alexmigf:
closes #673

### DIFF
--- a/includes/documents/class-wcpdf-invoice.php
+++ b/includes/documents/class-wcpdf-invoice.php
@@ -102,7 +102,7 @@ class Invoice extends Order_Document_Methods {
 			try {
 				// If a third-party plugin claims to generate invoice numbers, trigger this instead
 				if ( apply_filters( 'woocommerce_invoice_number_by_plugin', false ) || apply_filters( 'wpo_wcpdf_external_invoice_number_enabled', false, $this ) ) {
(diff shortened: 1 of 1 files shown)


MohamadNateqi (commented):
Alex, that hook used to be in a function only responsible for displaying the invoice number. The current function you added is to set/initiate the invoice number.
Would it be possible to lead to unexpected issues?

alexmigf:
I know that, actually I did debugging and testing before adding it here. The number will be formatted here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/d8e87a5200b8a9ef06658724f66508b76da5f5a3/includes/documents/class-wcpdf-invoice.php#L114-L118

The output for the example from the user:

Screenshot from 2023-12-06 16-34-58

The number is set to the document class data, but nothing touches the database, so I think it's safe. Do you see any potential problem here?

Of course, should be better to use plain numbers with the filter, but seems to work fine.

MohamadNateqi (approved):
Then it seems good!

PR #678, opened 2023-12-11
Fix a bug in "Full" document access

MohamadNateqi:
When the access type is set to "Full," using bulk actions to generate an invoice for a single order results in a permission error. Still, it's working correctly if several orders have been chosen.


alexmigf (approved):
LGTM!

PR #679, opened 2023-12-11
Tweak: includes `$args` in filter hook `wpo_wcpdf_filename` parameters

alexmigf:
This new parameter is required to pass the `output` key, which can be `ubl`.


MohamadNateqi (approved):
LGTM!
I think it also needs to be updated in the Pro extension.

alexmigf:
> I think it also needs to be updated in the Pro extension.

Yes, I'm working there, that's why I needed this parameter. But thank you for the reminder.

PR #681, opened 2023-12-14
Fix a bug in `get_payment_date()`

MohamadNateqi:
If there is no paid date for an order, the `get_date_paid()` returns null, and it can lead to this error:
`
Fatal error: Call to a member function getTimestamp() on null
C:\laragon\www\mohamad\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\documents\abstract-wcpdf-order-document-methods.php (478)
`

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -475,7 +475,7 @@ abstract class Order_Document_Methods extends Order_Document {
 			$payment_date = $this->order->get_date_paid();
 		}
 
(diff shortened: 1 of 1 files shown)


MohamadNateqi:
@alexmigf, We also may need to check the emptiness in templates like here:

https://github.com/wpovernight/woocommerce-pdf-ips-templates/blob/5b08c2636db4e88bd1e96aca88dffe408a02c780/templates/Simple%20Premium/receipt.php#L83C30

alexmigf:
@MohamadNateqi I agree, can you open an issue there and made the fix?

alexmigf (approved):
LGTM!

Issue #442, opened 2022-10-31
Restore 'VAT' and 'Tax rate' strings


YordanSoares:
Since we merged #419, the translations for the 'VAT' and 'Tax rate' columns stopped working. This happened because we removed the strings from the free main plugin, where they were not used. However, this makes that the translations disappeared for users of Premium Templates 2.19.0 and older.

PR #683, opened 2024-01-02
Declare new "Payment Date:" string

YordanSoares:
Currently, we are using the "Payment Date:" string within the Receipt document in both Professional and Premium Templates extensions, but it has not yet been declared within our free main plugin.

### DIFF
--- a/languages/strings.php
+++ b/languages/strings.php
@@ -2,6 +2,7 @@
 exit;
 __( 'VAT', 'woocommerce-pdf-invoices-packing-slips' );
 __( 'Tax rate', 'woocommerce-pdf-invoices-packing-slips' );
+__( 'Payment Date:', 'woocommerce-pdf-invoices-packing-slips' );
 __( 'Payment method', 'woocommerce-pdf-invoices-packing-slips' );
 __( 'Shipping method', 'woocommerce-pdf-invoices-packing-slips' );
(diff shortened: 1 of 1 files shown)


languages/strings.php
```diff
@@ -2,6 +2,7 @@
 exit;
 __( 'VAT', 'woocommerce-pdf-invoices-packing-slips' );
 __( 'Tax rate', 'woocommerce-pdf-invoices-packing-slips' );
+__( 'Payment Date:', 'woocommerce-pdf-invoices-packing-slips' );
```
alexmigf, line 5: I think the `:` is not necessary, which means that the template strings are wrong too and need to be fixed. Also `Date` could be `date`.

alexmigf (changes requested):

languages/strings.php
YordanSoares: @alexmigf done in [80c9622](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/683/commits/80c9622819d399257b66d3f836a9c162449d64c0)

alexmigf (approved):
LGTM!

languages/strings.php
alexmigf: I don't think this strings are for templates: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/442

PR #685, opened 2024-01-03
Fix: potential SQL injection bug in Number Tools

### LINKED ISSUE #684: SQL Injection vulnerability
In the Tools > Number:

> The `orderby` and `order` parameters are not properly sanitized in the SQL query, resulting in a potential injection vulnerability in the database. Since it utilizes the GET request method, a malicious user can exploit this by creating a specially crafted HTML page and then sending a request to the vulnerable endpoint (similar to CSRF attack).

alexmigf:
closes #684

### DIFF
--- a/includes/tables/class-wcpdf-number-store-list-table.php
+++ b/includes/tables/class-wcpdf-number-store-list-table.php
@@ -169,8 +169,8 @@ class Number_Store_List_Table extends \WP_List_Table {
 	 * @since 2.0
 	 * @return int Current page number
 	 */
-	public function get_paged() {
-		return isset( $_GET['paged'] ) ? absint( $_GET['paged'] ) : 1;
+	public function get_paged( $request ) {
+		return isset( $request['paged'] ) ? absint( $request['paged'] ) : 1;
 	}
 
 	/**
@@ -179,8 +179,8 @@ class Number_Store_List_Table extends \WP_List_Table {
 	 * @since 2.0
 	 * @return mixed string If search is present, false otherwise
 	 */
-	public function get_search() {
-		return ! empty( $_GET['s'] ) ? urldecode( trim( $_GET['s'] ) ) : false;
+	public function get_search( $request ) {
+		return ! empty( $request['s'] ) ? sanitize_text_field( $request['s'] ) : false;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/tables/class-wcpdf-number-store-list-table.php
```diff
@@ -193,14 +193,15 @@ public function get_search() {
...
 		$results                        = array();
-		$paged                          = $this->get_paged();
+		$paged                          = $this->get_paged( $request );
 		$offset                         = $this->per_page * ( $paged - 1 );
-		$search                         = $this->get_search();
-		$table_name                     = isset( $_GET['table_name'] ) ? sanitize_text_field( $_GET['table_name'] ) : null;
-		$order                          = isset( $_GET['order'] ) ? sanitize_text_field( $_GET['order'] ) : 'DESC';
-		$orderby                        = isset( $_GET['orderby'] ) ? sanitize_text_field( $_GET['orderby'] ) : 'id';
+		$search                         = $this->get_search( $request );
+		$table_name                     = isset( $request['table_name'] ) ? sanitize_text_field( $request['table_name'] ) : null;
```
MohamadNateqi, line 202: Do you think it's better not to take the table name from users? I don't see why we need to have this option here.

includes/tables/class-wcpdf-number-store-list-table.php
```diff
@@ -193,14 +193,15 @@ public function get_search() {
...
+		$order                          = isset( $request['order'] ) && in_array( $request['order'], array( 'DESC', 'ASC' ) ) ? sanitize_text_field( $request['order'] ) : 'DESC';
+		$orderby                        = isset( $request['orderby'] ) && in_array( $request['orderby'], array( 'id' ) ) ? sanitize_text_field( $request['orderby'] ) : 'id';
```
MohamadNateqi, line 204: Same for this, since it's always `id`.

alexmigf: For now it is, but it's ready to accept others.

includes/tables/class-wcpdf-number-store-list-table.php
alexmigf: Because you can select the table from the UI.

MohamadNateqi (approved):
LGTM!

includes/tables/class-wcpdf-number-store-list-table.php
alexmigf: Please check my last commit.

MohamadNateqi (approved):
Great change!
LGTM!

PR #700, opened 2024-01-25
Fix sorting orders by invoice number and date issue on the orders page

### LINKED ISSUE #696: Improve the sorting of the invoice number column
Currently, we order the invoice number column by `meta_value`, which work well with some formatted numbers, e.g. `INV-000123`:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/37c7d60fda123b03845ed704592978adc5b7f38a/includes/class-wcpdf-admin.php#L433

However, it doesn't work correctly with invoice number without a format, where the sorting behaves in an unexpected way, e.g. sorting the numbers like 1, 11, 111, 2, 22, 3, 33, 33, etc.

It can be fixed with this code snippet:

```php
/**
 * PDF Invoices & Packing Slips for WooCommerce:
 * Change the orderby value to 'meta_value_num' to sort invoice numbers without a format
 */
add_filter( 'wpo_wcpdf_invoice_number_column_orderby', function() { return 'meta_value_num'; } );
```

But I think we should try to improve this behavior by default, making the sorting smarter, so the code snippet above is no longer necessary.

### LINKED ISSUE #699: Sort orders by invoice# issue when HPOS is on
[This customer](https://secure.helpscout.net/conversation/2480132681/57572/) has reported that with HPOS activated, they can no longer sort the orders by invoice numbers on the Orders page.

MohamadNateqi:
close #699,
close #696

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -1374,12 +1374,14 @@ class Admin {
 
 	public function adjust_order_list_query_args( array $order_query_args ): array {
 		if ( 'invoice_number_column' === $order_query_args['orderby'] ) {
+			$is_numeric = $this->is_invoice_number_numeric();
+
 			$order_query_args['meta_query'] = array(
 				'invoice_number_column' => array(
 					'key'     => '_wcpdf_invoice_number',
 					'compare' => '!=',
 					'value'   => '0',
-					'type'    => 'NUMERIC',
+					'type'    => $is_numeric ? 'NUMERIC' : 'CHAR',
 				),
 			);
 		}
@@ -1402,7 +1404,22 @@ class Admin {
 			return;
 		}
 
-		$query->set( 'orderby', 'meta_value_num' );
+		$query->set( 'orderby', $this->is_invoice_number_numeric() ? 'meta_value_num' : 'meta_value' );
+	}
+
+	/**
+     * Determines if the invoice number is numeric.
+     * It evaluates the presence of non-numeric characters in the prefix and suffix of the invoice number.
+     *
+	 * @return bool
+	 */
+	private function is_invoice_number_numeric() {
+		$invoice          = wcpdf_get_invoice( null );
+		$invoice_settings = $invoice->get_settings();
+		$is_numeric       = ( empty( $invoice_settings['number_format']['prefix'] ) || ctype_digit( $invoice_settings['number_format']['prefix'] ) ) &&
+		                    ( empty( $invoice_settings['number_format']['suffix'] ) || ctype_digit( $invoice_settings['number_format']['suffix'] ) );
+
+		return apply_filters( 'wpo_wcpdf_order_list_invoice_number_column_is_numeric', $is_numeric );
 	}
 
 }
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -30,7 +30,7 @@ class Admin {
 			add_filter( 'manage_woocommerce_page_wc-orders_sortable_columns', array( $this, 'invoice_columns_sortable' ) ); // WC 7.1+
             // Fix sortable issue when HPOS is on.
 			add_filter( 'woocommerce_shop_order_list_table_sortable_columns', array( $this, 'add_invoice_column_to_sortable_columns' ) );
-			add_filter( 'woocommerce_order_list_table_prepare_items_query_args', array( $this, 'modify_woo_query' ) );
+			add_filter( 'woocommerce_order_list_table_prepare_items_query_args', array( $this, 'adjust_order_list_query_args' ) );
 
 			add_filter( 'manage_edit-shop_order_columns', array( $this, 'add_invoice_columns' ), 999 );
 			add_action( 'manage_shop_order_posts_custom_column', array( $this, 'invoice_columns_data' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


MohamadNateqi:
@alexmigf This PR still needs to be completed, but can you run a test to see if the provided solution works correctly on your end? It's fixing only the invoice number for now.

alexmigf:
@MohamadNateqi date sort doesn't seem to be working for me, and the number seems to work but keeps the same issue has before:

Screenshot from 2024-01-25 13-36-47

Do you think you could fix it? I think it was not possible at the time, but can't recall with precision.

MohamadNateqi:
@alexmigf The date column should be fixed in the latest commit.
Also, the sorting issue for numbers should be resolved.
Ready to review.

includes/class-wcpdf-admin.php
```diff
@@ -28,6 +28,10 @@ public function __construct()	{
 			add_filter( 'manage_woocommerce_page_wc-orders_columns', array( $this, 'add_invoice_columns' ), 999 ); // WC 7.1+
 			add_action( 'manage_woocommerce_page_wc-orders_custom_column', array( $this, 'invoice_columns_data' ), 10, 2 ); // WC 7.1+
 			add_filter( 'manage_woocommerce_page_wc-orders_sortable_columns', array( $this, 'invoice_columns_sortable' ) ); // WC 7.1+
+            // Fix sortable issue when HPOS is on.
```
alexmigf, line 31: HPOS is now the default for new stores, so maybe we shouldn't call this a fix?

alexmigf:
Can you also fix the non HPOS number sorting?

Screenshot from 2024-01-25 15-29-27

includes/class-wcpdf-admin.php
MohamadNateqi: Good point. I have removed it here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/700/commits/3b58258afb8207515131f32d04b2a85f4796f50f.

MohamadNateqi:
> Can you also fix the non HPOS number sorting?

It should be fixed with the latest commit. Can you give it a try, please?

alexmigf (approved):
Fantastic job!

includes/class-wcpdf-admin.php
```diff
@@ -1360,6 +1364,47 @@ public function add_invoice_number_value_to_order_export( $export_item, $item )
...
+
+		return $order_query_args;
+	}
+
+	public function sort_orders_by_numeric_invoice_number( $query ): void {
+		if ( ! is_admin() || ! $query->is_main_query() || 'shop_order' !== $query->get( 'post_type' ) && '_wcpdf_invoice_number' !== $query->get( 'meta_key' ) ) {
+			return;
+		}
+
+		$query->set( 'orderby', 'meta_value_num' );
```
YordanSoares, line 1405: I think we should wrap `meta_value_num` within the `wpo_wcpdf_invoice_number_column_orderby` filter hook, to let the customer switch to `meta_value`, if they are using a prefix/suffix, e.g. `INV-000123/2024`.

YordanSoares (changes requested):

includes/class-wcpdf-admin.php
```diff
@@ -1402,7 +1404,22 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+
+	/**
+     * Determines if the invoice number is numeric.
+     * It evaluates the presence of non-numeric characters in the prefix and suffix of the invoice number.
+     *
+	 * @return bool
+	 */
+	private function is_invoice_number_numeric() {
+		$invoice          = wcpdf_get_invoice( null );
+		$invoice_settings = $invoice->get_settings();
```
alexmigf, line 1418: Maybe you could use:

`$invoice_settings = WPO_WCPDF()->settings->get_document_settings( 'invoice' );`

includes/class-wcpdf-admin.php
```diff
@@ -1402,7 +1404,22 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+		$is_numeric       = ( empty( $invoice_settings['number_format']['prefix'] ) || ctype_digit( $invoice_settings['number_format']['prefix'] ) ) &&
+		                    ( empty( $invoice_settings['number_format']['suffix'] ) || ctype_digit( $invoice_settings['number_format']['suffix'] ) );
+
+		return apply_filters( 'wpo_wcpdf_order_list_invoice_number_column_is_numeric', $is_numeric );
```
alexmigf, line 1422: I think the filter name could be only: `wpo_wcpdf_invoice_number_is_numeric`

alexmigf (changes requested):

dwalkerpriv (approved):
Works well for both the invoice number and invoice date 👍

alexmigf (approved):

YordanSoares (commented):
@MohamadNateqi they last improvements work great! I just wonder if they work with non-HPOS stores, do they?

MohamadNateqi:
@YordanSoares, It should work correctly for both. I added two methods to fix the numbering issue: one for non-HPOS and one for HPOS stores, and having non-numeric characters in numbers is considered for both.

YordanSoares (approved):
Thanks for confirming @MohamadNateqi! I tested anyway, just in case: Great work!

Please note that with your latest improvements, we can also close #699, so _we free two birds with one key_* 😉

\* This is a version of the famous saying, following the [PETA](https://www.peta.org/teachkind/lesson-plans-activities/animal-friendly-idioms/) animal-friendly idiom spirit.

MohamadNateqi:
@YordanSoares, Thanks for testing.
I think you wanted to mention another issue since the one you mentioned is linked to this PR.

YordanSoares:
@MohamadNateqi my bad! I wanted to mention #696.

MohamadNateqi:
@YordanSoares, Thanks. Killing two... sorry, free two birds with one key indeed. 👍

Issue #125, opened 2020-12-08
Try if Symfony Polyfill fixes missing characters if the MBString extension is disabled


alexmigf:
Try to test if the Symfony Polyfill library would solve the missing characters when the MBString extension is disabled. 

Pay attention that this library only supports PHP >= 7.1

Library: https://github.com/symfony/polyfill-mbstring
Example issue: https://wordpress.org/support/topic/german-characters-wrong-in-pdf-auo-umlaute/

Spreeuw:
It looks like dompdf 1.0 may have added a new MB String function (`mb_substitute_character`), seeing this error:
https://wordpress.org/support/topic/error-when-changing-the-statue-of-an-order/
![02 11 2021-18 45](https://user-images.githubusercontent.com/3217773/107693006-d72b1900-6cad-11eb-92bc-cb2fd8d92f3d.png)

Yet another reason to bring in the polyfill. I'm not sure about how to best handle the PHP7.1 requirement though, that would mean we can't load it via composer and need to use our own autoloader for this and only load it for PHP7.1+.

Spreeuw:
closed by #223

PR #703, opened 2024-01-29
#125 regular price conversion

BrunoPavlinic98:
closes https://github.com/wpovernight/woocommerce-pdf-ips-templates/issues/125

I have implemented a function that will check if the order currency is the same as the default currency set in Woo Commerce, and will convert the prices to the order currency if order and default currencies are not the same. The prices will remain the same if:

The order currency and default currency are the same.
The order currency does not exist.
The order currency is not set up in the Woo Commerce Multi Currency settings.
This conversion should only affect the PDF values. The order, product, and order item data should not be touched.

I have tested this with the following multi currency plugin - https://docs.villatheme.com/?item=woocommerce-multi-currency

Reproduction steps:

Add a Curcy WooCommerce Multi Currency plugin.
In the plugin settings, set 2 different currencies, one of them being the default one. Wait till the "Rate + Exchange Fee" is properly set up. It takes time for the plugin to retrieve this data, and if you click "Save" before the rates are set, the rates for all currencies will be 1 and the conversion won't make sense.
Add a new order, and try to create a pdf invoice for it. You will also need to setup that the invoice has the "Regular price" field.
The invoice should have the a regular price the same as the product regular price.
Change the default currency.
The regular price should be converted to the order currency.

The function works only for Curcy WooCommerce Multi Currency plugin, but can be extended to other plugins as well.

### DIFF
--- a/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
+++ b/includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
@@ -65,9 +65,9 @@ class Third_Party_Plugins {
 		}
 
 		// Currency converter implementation.
-		add_action( 'wpo_wcpdf_process_template_order', array( $this, 'add_convert_to_order_currency_filter' ), 11, 2 );
-		add_action( 'wpo_wcpdf_after_html', array( $this, 'remove_convert_to_order_currency_filter' ), 11, 0 );
-		add_action( 'wpo_wcpdf_after_pdf', array( $this, 'remove_convert_to_order_currency_filter' ), 11, 0 );
+		add_action( 'wpo_wcpdf_process_template_order', array( $this, 'add_convert_price_to_order_currency_filter' ), 11, 2 );
+		add_action( 'wpo_wcpdf_after_html', array( $this, 'remove_convert_price_to_order_currency_filter' ), 11, 0 );
+		add_action( 'wpo_wcpdf_after_pdf', array( $this, 'remove_convert_price_to_order_currency_filter' ), 11, 0 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


BrunoPavlinic98:
After speaking with @YordanSoares, I was informed that Curcy plugin is the most used plugin for multicurrency features.
Here is the list of other plugins I have tested:
https://wordpress.org/plugins/wc-multi-currency/
https://wordpress.org/plugins/premmerce-woocommerce-multi-currency/
https://hr.wordpress.org/plugins/woocommerce-multilingual/

The listed plugins work in a different way than Curcy. They either store the order item prices in order item metadata table, with prices being converted, or they change the product price once a default currency is change. This is an issue because we can't know which prices need conversion and which are already converted on our invoices and packing slips. In each of the listed plugins, the product regular price was not converted.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -311,6 +316,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+			return $this->convert_to_order_currency( $raw_price, $order_id );
+		}, 999, 1 );
+	}
+
+	public function remove_convert_to_order_currency_filter() {
+		remove_all_filters( 'raw_woocommerce_price', 999 );
+	}
+
+	public function convert_to_order_currency( $raw_price, $order_id ) {
+		$order = wc_get_order( $order_id );
```
alexmigf, line 332: You should check if the  order object exists.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -311,6 +316,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+		$price = $raw_price;
+
+		// Filter for enabling or disabling conversion to order currency.
+		$should_convert = apply_filters( 'wpo_wcpdf_should_convert_to_order_currency', true );
+
+		if($should_convert) {
```
alexmigf, line 338: You can add the filter directly here:

```php
if ( apply_filters( 'wpo_wcpdf_should_convert_price_to_order_currency', '__return_true' ) ) {
```

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -311,6 +316,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+			// CURCY - Multi Currency for WooCommerce (https://villatheme.com/) implementation
+			if ( function_exists( 'wmc_get_price' ) && method_exists( $order, 'get_currency' ) && function_exists( 'get_woocommerce_currency' ) && get_woocommerce_currency() != $order->get_currency() ) {
```
alexmigf, line 340: `get_woocommerce_currency() !== $order->get_currency()`

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -311,6 +316,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+		add_filter( 'raw_woocommerce_price', function( $raw_price ) use ( $order_id ) {
+			return $this->convert_to_order_currency( $raw_price, $order_id );
+		}, 999, 1 );
+	}
+
+	public function remove_convert_to_order_currency_filter() {
+		remove_all_filters( 'raw_woocommerce_price', 999 );
+	}
+
+	public function convert_to_order_currency( $raw_price, $order_id ) {
```
alexmigf, line 331: Maybe this should be renamed to `convert_price_to_order_currency`?

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -311,6 +316,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+		}
+
+		/**
+		 * Filters the price converted to order currency.
+		 *
+		 * @param float $price     Price after conversion.
+		 * @param float $raw_price Price before conversion.
+		 * @param int   $order_id  The order id from order that is being processed.
+		 */
+		return apply_filters( 'wpo_wcpdf_convert_to_order_currency', $price, $raw_price, $order_id );
```
alexmigf, line 352: This too? `wpo_wcpdf_convert_price_to_order_currency`

alexmigf: And all the others missing `price` in the name.

alexmigf (changes requested):

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
alexmigf: No, it can be.

BrunoPavlinic98:
I agree with all comments and I have fixed them.

PR #705, opened 2024-01-31
Fix: renumbering documents not working

### LINKED ISSUE #704: Renumbering documents tool not working
A [user](https://wordpress.org/support/topic/renumber-existing-documents-not-working/#post-17385859) reported that the renumbering documents tool is not working properly.

alexmigf:
closes #704

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -262,7 +262,7 @@ abstract class Order_Document {
 		}
 	}
 	
-	public function add_number( $force_new_number = false ) {
+	public function initiate_number( $force_new_number = false ) {
 		$lock            = new Semaphore( $this->lock_name, $this->lock_time, $this->lock_loggers, $this->lock_context );
 		$document_number = $this->exists() ? $this->get_data( 'number' ) : null;
 		$document_number = ! empty( $document_number ) && $force_new_number ? null : $document_number;
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -134,12 +134,6 @@ abstract class Order_Document {
 	 * @param  int|object|WC_Order $order Order to init.
 	 */
 	public function __construct( $order = 0 ) {
-		// semaphore
-		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
-		$this->lock_time    = apply_filters( 'wpo_wcpdf_document_semaphore_lock_time', 60 );
-		$this->lock_retries = apply_filters( 'wpo_wcpdf_document_semaphore_lock_retries', 0 );
-		$this->lock_loggers = apply_filters( 'wpo_wcpdf_document_semaphore_lock_loggers', isset( WPO_WCPDF()->settings->debug_settings['semaphore_logs'] ) ? array( wc_get_logger() ) : array() );
-		
 		if ( is_numeric( $order ) && $order > 0 ) {
 			$this->order_id = $order;
 			$this->order    = wc_get_order( $this->order_id );
@@ -152,12 +146,18 @@ abstract class Order_Document {
 		$this->slug = ! empty( $this->type ) ? str_replace(  '-', '_', $this->type ) : '';
 		
 		// output formats
-		$this->output_formats = apply_filters( "wpo_wcpdf_{$this->type}_output_formats", array( 'pdf' ), $this  );
+		$this->output_formats = apply_filters( "wpo_wcpdf_{$this->slug}_output_formats", array( 'pdf' ), $this  );
 
 		// load data
 		if ( $this->order ) {
 			$this->read_data( $this->order );
 		}
+		
+		// semaphore
+		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
+		$this->lock_time    = apply_filters( "wpo_wcpdf_{$this->slug}_semaphore_lock_time", 60 );
+		$this->lock_retries = apply_filters( "wpo_wcpdf_{$this->slug}_semaphore_lock_retries", 0 );
+		$this->lock_loggers = apply_filters( 'wpo_wcpdf_document_semaphore_lock_loggers', isset( WPO_WCPDF()->settings->debug_settings['semaphore_logs'] ) ? array( wc_get_logger() ) : array() );
 
 		// load settings
 		$this->init_settings_data();
@@ -268,7 +268,6 @@ abstract class Order_Document {
 		$document_number = ! empty( $document_number ) && $force_new_number ? null : $document_number;
 		
 		if ( $lock->lock( $this->lock_retries ) && empty( $document_number ) ) {
-			
 			$lock->log( "Lock acquired for the {$this->slug} number init.", 'info' );
 			
 			try {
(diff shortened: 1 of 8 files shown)


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -103,6 +104,13 @@ abstract class Order_Document {
 	 * @param  int|object|WC_Order $order Order to init.
 	 */
 	public function __construct( $order = 0 ) {
+		// semaphore
+		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
```
MohamadNateqi, line 108: Dynamic class property.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -224,6 +232,59 @@ public function save_settings( $latest = false ) {
...
+			$lock->log( "Lock acquired for the {$this->slug} number init.", 'info' );
+			
+			try {
+				// If a third-party plugin claims to generate document numbers, trigger this instead
+				if ( apply_filters( "woocommerce_{$this->slug}_number_by_plugin", false ) || apply_filters( "wpo_wcpdf_external_{$this->slug}_number_enabled", false, $this ) ) {
+					$document_number = apply_filters( "woocommerce_generate_{$this->slug}_number", $document_number, $this->order );  // legacy (backwards compatibility)
+					$document_number = apply_filters( "woocommerce_{$this->slug}_number", $document_number, $this->order->get_id() ); // legacy (backwards compatibility)
+					$document_number = apply_filters( "wpo_wcpdf_external_{$this->slug}_number", $document_number, $this );
+				} elseif ( isset( $this->settings['display_number'] ) && 'order_number' === $this->settings['display_number'] && ! empty( $this->order ) ) {
+					$document_number = $this->order->get_order_number();
```
MohamadNateqi, line 252: Do you think we can change this part?

Order numbers are always present in the Woo database, and we don't need to save them in our table. So we can always have the document separate numbering system if no third party changes it.

We can easily display the order number if the user wants that as the document number in the method related to display.

MohamadNateqi (changes requested):

includes/documents/abstract-wcpdf-order-document.php
alexmigf: This will involve more work that could bring more issues, because there is a class for the numbers to apply formatting, and I want to avoid too much drastic changes. I think we should keep this simple for now.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -103,6 +134,12 @@ abstract class Order_Document {
 	 * @param  int|object|WC_Order $order Order to init.
 	 */
 	public function __construct( $order = 0 ) {
+		// semaphore
+		$this->lock_name    = "wpo_wcpdf_{$this->slug}_semaphore_lock";
+		$this->lock_time    = apply_filters( 'wpo_wcpdf_document_semaphore_lock_time', 60 );
```
MohamadNateqi, line 139: Maybe it's good to also send the document slug/type as a parameter to this filter?

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -542,6 +545,16 @@ public function legacy_addon_notices() {
 			}
 		}
 	}
+	
+	public function init_number_declaration_error_custom_message( $message, $error ) {
+		if ( isset( $error['message'] ) && false !== strpos( $error['message'], 'Declaration of WPO\WC\PDF_Invoices\Documents\Pro_Document::init_number() must be compatible with' ) ) {
+			$message  = '<p>' . __( 'A plugin extending the document abstract class of the PDF Invoices & Packing Slips for WooCommerce plugin version 3.7.8 or higher may either not be running the latest version or require additional code improvements. Please check the error associated with the affected plugin below.', 'woocommerce-pdf-invoices-packing-slips' ) . '</p>'; 
+			$message .= '<p>' . wp_kses_post( $error['message'] ) . '</p>';
+			$message .= '<p><a href="' . esc_url( '' ) . '">' . __( 'Troubleshoot the issue with this helpful guide!', 'woocommerce-pdf-invoices-packing-slips' ) . '</a></p>';
```
alexmigf, line 553: We need to create a documentation page to help customers update their Proposal plugin from FTP, because if the site crashes is the only way, I think renaming the plugin doesn't give them access to the backend either.

MohamadNateqi: @alexmigf, It seems renaming will work since I could run the site again with just renaming.

MohamadNateqi (changes requested):

includes/documents/class-wcpdf-invoice.php
```diff
@@ -82,65 +73,24 @@ public function init() {
...
-			if ( $lock->release() ) {
-				$lock->log( 'Lock released for the invoice number init.', 'info' );
-			}
-			
-		} else {
-			$lock->log( 'Couldn\'t get the lock for the invoice number init.', 'critical' );
-		}
-		
-		return $invoice_number;
+		return $this->add_number();
```
MohamadNateqi, line 93: What do you think about adding a warning to logs to inform about deprecation?

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -552,11 +552,31 @@ private function renumber_or_delete_document( $document, $delete_or_renumber ) {
...
+							// packing slip
+							if ( 'packing-slip' === $document->get_type() && is_callable( array( WPO_WCPDF_Pro()->functions, 'init_packing_slip_number' ) ) ) {
+								WPO_WCPDF_Pro()->functions->init_packing_slip_number( $document );
+								$return = true;
+							// other documents
+							} elseif ( is_callable( array( $document, 'init_number' ) ) ) {
+								$document->init_number();
+								$return = true;
+							}
+						// current
```
MohamadNateqi, line 568: Can this part be removed?
I thought maybe we don't need to call the `init_packing_slip_number()` or `init_number()` while we have this update. We can call the new `add_number()` method.

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -552,11 +552,31 @@ private function renumber_or_delete_document( $document, $delete_or_renumber ) {
 		if ( $document && $document->exists() ) {
 			switch ( $delete_or_renumber ) {
 				case 'renumber':
-					if ( is_callable( array( $document, 'init_number' ) ) ) {
-						$document->init_number();
-						$return = true;
-					} elseif ( 'packing-slip' === $document->get_type() && is_callable( array( WPO_WCPDF_Pro()->functions, 'init_packing_slip_number' ) ) ) {
-						WPO_WCPDF_Pro()->functions->init_packing_slip_number( $document );
+					// pro and third party documents
+					if ( 'invoice' !== $document->get_type() ) {
```
MohamadNateqi, line 556: Can we also remove this condition?
So the whole renumber case would be like this: if the `add_number()` is callable, call it, else if `init_number()` is callable, call it.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -224,6 +261,58 @@ public function save_settings( $latest = false ) {
 			}
 		}
 	}
+	
+	public function add_number( $force_new_number = false ) {
```
MohamadNateqi, line 265: What do you think about `initiate_number()` for the method name?

MohamadNateqi (changes requested):
Can be removed:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/705/files#diff-19a452d809d082bbe584c079e40cf32b00abd705fa2e4e60c6ec4f4cf6e1efd6L21-L25

MohamadNateqi (approved):
Perfect!

YordanSoares (approved):
LGTM! Tested with legacy and HPOS order data storage.

dwalkerpriv (approved):
Deleting numbers, renumbering, recreating and renumbering documents after deleting them works great for all documents 👍

dpeyou (approved):
It appears ready to be released into the wild :+1:

PR #706, opened 2024-02-05
Fix: disables Order Proposal plugin in versions below 2.0.2 to force update

alexmigf:
Regarding this future implementation: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/705

### DIFF
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -3,7 +3,7 @@
  * Plugin Name:          PDF Invoices & Packing Slips for WooCommerce
  * Plugin URI:           https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/
  * Description:          Create, print & email PDF or UBL Invoices & PDF Packing Slips for WooCommerce orders.
- * Version:              3.7.7
+ * Version:              3.7.8-beta-1
  * Author:               WP Overnight
  * Author URI:           https://www.wpovernight.com
  * License:              GPLv2 or later
@@ -21,7 +21,7 @@ if ( ! class_exists( 'WPO_WCPDF' ) ) :
 
 class WPO_WCPDF {
 
-	public $version              = '3.7.7';
+	public $version              = '3.7.8-beta-1';
 	public $version_php          = '7.2';
 	public $version_woo          = '3.0';
 	public $version_wp           = '4.4';
@@ -77,6 +77,7 @@ class WPO_WCPDF {
 		add_action( 'admin_notices', array( $this, 'mailpoet_mta_detected' ) );
 		add_action( 'admin_notices', array( $this, 'rtl_detected' ) );
 		add_action( 'admin_notices', array( $this, 'legacy_addon_notices' ) );
+		add_action( 'admin_notices', array( $this, 'order_proposal_below_2_0_2_notice' ) );
 		
 		// deactivate legacy extensions if activated
 		register_activation_hook( __FILE__, array( $this, 'deactivate_legacy_addons' ) );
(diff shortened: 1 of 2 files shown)


MohamadNateqi:
What do you think about always checking for the Proposal version? Not only in the upgrade, to avoid crashing stores?

alexmigf:
I think the upgrade routine runs also when doing new installs.

MohamadNateqi:
The only time it can crash the site is when they install an old version of the Order Proposal while they have the latest version of the PDF Invoice plugin. Right?
In this case, the only way to prevent activating that is to always check the Proposal version. Do you think it's a good idea to do this?

alexmigf:
How do you propose to do that?

MohamadNateqi:
We can add a function to the "plugins_loaded" hook to check dependent plugins. What do you think?

alexmigf:
What that differs from including in the upgrade routine?

MohamadNateqi:
If the customer doesn't have the Order Proposal plugin installed and then installs it when they have the latest version of PDF Invoices, the upgrade routine won't happen, and this could lead to the crash you mentioned [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/705#discussion_r1478394392). I think this is the only scenario that could lead to a crash.
So, I suggested directly checking the installed version of the Order Proposal to prevent this scenario.

alexmigf:
If they install the Proposal plugin I think they will do it with the latest version, but doesn't hurt adding that as well.

MohamadNateqi:
Would it solve the crash issue you mentioned? Or are there also other scenarios that could lead to crash stores?

alexmigf:
The only scenarios I'm seeing now are:

* The Proposal plugin running below 2.0.2
* Other third party plugin that extends our document class not declaring he `init_number` correctly. We need to check compatible plugins about this, specially the [Italian addon](https://github.com/alexmigf/woocommerce-pdf-italian-add-on) and provide them with changes if necessary.
* Fix here also https://github.com/wpovernight/wcpdf-quotation

I think a few users might receive the fatal error, that's why I'm providing a custom message with a documentation link so they could fix on their end if that happens.

MohamadNateqi:
So It seems like there is no way to avoid this situation!
Then maybe not so much necessary to add my suggestion [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/706#issuecomment-1927374823).

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -542,6 +543,34 @@ public function legacy_addon_notices() {
...
+		if ( get_transient( $transient_name ) ) {			
+			ob_start();
+			?>
+			<div class="notice notice-warning">
+				<p><?php _e( 'Attention! We detected that your WooCommerce Order Proposal plugin is running a version below 2.0.2. In the next version of the PDF Invoices & Packing Slips for WooCommerce plugin, we are making a drastic change that requires at least version 2.0.2 of the Proposals plugin. To avoid clashes, we\'ve temporarily deactivated it—feel free to update to version 2.0.2 or higher. Thank you!', 'woocommerce-pdf-invoices-packing-slips' ); ?></p>
+				<p><a href="<?php echo esc_url( wp_nonce_url( add_query_arg( array( $query_arg => true ) ), 'hide_order_proposal_below_2_0_2_notice_nonce' ) ); ?>"><?php _e( 'Hide this message', 'woocommerce-pdf-invoices-packing-slips' ); ?></a></p>
+			</div>
+			<?php
+			echo wp_kses_post( ob_get_clean() );
+		}
```
MohamadNateqi, line 560: Should it check the Order Proposal plugin version before displaying this notice?

dwalkerpriv (approved):
Appears appropriately and does disable the plugin 👍 

Screenshot 2024-02-05 152734

alexmigf:
I have changed my mind, I think we should play safe and keep the original function for now. Closing.

PR #712, opened 2024-02-09
New: function `wcpdf_get_document_output_format_extension`

alexmigf:


includes/wcpdf-functions.php
```diff
@@ -252,6 +252,21 @@ function wcpdf_get_document_file( object $document, string $output_format = 'pdf
...
 }
 
+/**
+ * Get the document output format extension
+ *
+ * @param  string $output_format
+ * @return string
+ */
+function wcpdf_get_document_output_format_extension( string $output_format ): string {
+	$output_formats = array(
```
MohamadNateqi, line 262: Would it be useful to have a filter for this?

MohamadNateqi (approved):
LGTM!

includes/wcpdf-functions.php
alexmigf: I thought about that, but we have them hard coded in other places, and I'm not seeing use for that since we don't support other formats currently (I also don't see other formats that we could provide in the future).

MohamadNateqi: All good then.

PR #713, opened 2024-02-09
New: function `wcpdf_error_handling`

alexmigf:
Based on @brualio suggestion via Slack DM to avoid having a fatal error when calling this function by passing a `$document` value has `false` on `wcpdf_get_document_file`. This way we could determine how the function should handle the errors.

### DIFF
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -218,6 +218,7 @@ function wcpdf_ubl_headers( $filename, $size ) {
  * 
  * @param  object $document
  * @param  string $output_format
+ * @param  string $error_handling
  * @return string
  */
 function wcpdf_get_document_file( object $document, string $output_format = 'pdf', $error_handling = 'exception' ): string {
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -220,11 +220,12 @@ function wcpdf_ubl_headers( $filename, $size ) {
  * @param  string $output_format
  * @return string
  */
-function wcpdf_get_document_file( object $document, string $output_format = 'pdf' ): string {
+function wcpdf_get_document_file( object $document, string $output_format = 'pdf', $error_handling = 'exception' ): string {
 	$default_output_format = 'pdf';
 	
 	if ( ! $document ) {
-		throw new \Exception( 'Invalid document argument provided' );
+		$error_message = 'No document object provided.';
+		return wcpdf_error_handling( $error_message, $error_handling, true, 'critical' );
 	}
 	
 	if ( empty( $output_format ) ) {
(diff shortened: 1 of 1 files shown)


MohamadNateqi:
> I think the fatal error I was getting was because `false` can't be an object. Maybe changing `object` to `mixed` would be better?

@brualio, We mainly added type hints to enforce type safety. If `false` is being passed, it shows the need for better error handling or validation before calling this function. It's not on our part, but the caller should ensure that the correct parameters are passed to the function, and our objective is to raise an error if the wrong type has been passed.

Therefore, I think we better keep the type to `object`, which could make our code safe and easy to understand.

includes/wcpdf-functions.php
```diff
@@ -350,6 +355,38 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+/**
+ * Error handling function
+ *
+ * @param string $message
+ * @param string $error_handling
+ * @param bool   $log_error
+ * @param string $log_level
+ * @return mixed
+ */
+function wcpdf_error_handling( $message, $error_handling = 'exception', $log_error = false, $log_level = 'error' ) {
```
MohamadNateqi, line 367: Type hints.

includes/wcpdf-functions.php
```diff
@@ -350,6 +355,38 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+			break;
+		case 'output':
+			wcpdf_output_error( $message, $log_level );
+			break;
+		case 'return':
+			return $message;
+			break;
+		case 'false':
+			return false;
+			break;
```
MohamadNateqi, line 384: I think it can be removed since it is the same as the return at the end of the function.

includes/wcpdf-functions.php
```diff
@@ -350,6 +355,38 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
+	switch ( $error_handling ) {
+		case 'exception':
+			throw new \Exception( $message );
+			break;
+		case 'output':
+			wcpdf_output_error( $message, $log_level );
+			break;
+		case 'return':
+			return $message;
+			break;
```
MohamadNateqi, line 381: Maybe we can remove this part and only have two types of handling: exception and output. And if so, the `$error_handling` could change to a `bool` type. What do you think?

includes/wcpdf-functions.php
MohamadNateqi: Would it be better to have the log error `true` as the default value?

includes/wcpdf-functions.php
```diff
@@ -350,6 +355,38 @@ function wcpdf_output_error( $message, $level = 'error', $e = null ) {
...
 }
 
+/**
+ * Error handling function
+ *
+ * @param string $message
+ * @param string $error_handling
+ * @param bool   $log_error
+ * @param string $log_level
+ * @return mixed
```
MohamadNateqi, line 365: Also adding `@throws Exception` might be good.

MohamadNateqi (changes requested):

includes/wcpdf-functions.php
alexmigf: I think I prefer keeping it as string, this way we could extend in the future.

MohamadNateqi (approved):
LGTM!

brualio:
LGTM! 🎆

brualio (approved):
LGTM! 🎆

PR #717, opened 2024-02-14
Changed default nonce name for wizard dismissal

### LINKED ISSUE #714: Bug: "I am the wizard" button is not working!
The banner displayed when the plugin is installed for the first time includes a button labeled "I am the wizard." However, this button does not dismiss the banner.

image

brualio:
Closes #714 

The default nonce name `_wpnonce` was causing the conflict, so it was changed to `_wpdismissnonce`.

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -89,9 +89,9 @@ class Admin {
 		if ( get_option( 'wpo_wcpdf_review_notice_dismissed' ) !== false ) {
 			return;
 		} else {
-			if ( isset( $_REQUEST['wpo_wcpdf_dismiss_review'] ) && isset( $_REQUEST['_wpnonce'] ) ) {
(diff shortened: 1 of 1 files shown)


brualio:
By the way, there's another dialog that should show up after 100 PDF Downloads. It might be a good idea to update it too, but not sure if here or in another issue.

MohamadNateqi:
> By the way, there's another dialog that should show up after 100 PDF Downloads. It might be a good idea to update it too, but not sure if here or in another issue.

I also agree with this.

alexmigf:
@brualio I agree too, can you do that implementation as well?

brualio:
@alexmigf @MohamadNateqi Alright! The code has been updated.

MohamadNateqi (approved):
LGTM!

alexmigf (approved):
LGTM!

Issue #719, opened 2024-02-15
"Disable for free orders" option does not affect existing free orders.


YordanSoares:
This user reported that switching the "Disable for free orders" option does not affect existing free orders.

I have tested myself and managed to reproduce the scenario: It is affecting only new orders.

PR #720, opened 2024-02-15
Check document settings for the "Disable for free orders" option

YordanSoares:
This PR switches the condition for the "Disable for free orders" option to the document settings instead of the historical setting (stored within the order data at the moment of its creation).

Closes: #719


MohamadNateqi (approved):
LGTM!

alexmigf (approved):
LGTM!

PR #722, opened 2024-02-16
Update `get_tax_rate` method to use Woo getters

### LINKED ISSUE #716: Using getters to access Woo class properties
The `get_tax_rate()` in the `Order_Document_Methods` class directly accesses class properties of the `WC_Order_Item` child classes of Woocommerce, which it needs to use the getter provided. This way of accessing properties for `WC_Order_Item_Coupon` can raise this log:
`The WC_Order_Item_Coupon::offsetGet function is deprecated since version 4.4.0.`

This first happened in this [PR](https://github.com/wpovernight/woocommerce-pdf-ips-templates/pull/281#issuecomment-1943549593).

MohamadNateqi:
close #716 

The Getters method exists in version `3.0.0`:
https://github.com/woocommerce/woocommerce/blob/3.0.0/includes/class-wc-order-item-coupon.php
https://github.com/woocommerce/woocommerce/blob/3.0.0/includes/class-wc-order-item-fee.php
https://github.com/woocommerce/woocommerce/blob/3.0.0/includes/class-wc-order-item-product.php
https://github.com/woocommerce/woocommerce/blob/3.0.0/includes/class-wc-order-item-shipping.php
https://github.com/woocommerce/woocommerce/blob/3.0.0/includes/class-wc-order-item-tax.php

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -640,12 +640,12 @@ abstract class Order_Document_Methods extends Order_Document {
 	 * @param WC_Order|WC_Order_Refund $order WC_Order
 	 * @param bool $force_calculation force calculation of rates rather than retrieving from db
 	 *
-	 * @return string $tax_rates imploded list of tax rates
+	 * @return string|float $tax_rates imploded list of tax rates
 	 */
-	public function get_tax_rate( object $item, $order, bool $force_calculation = false ): string {
+	public function get_tax_rate( object $item, $order, bool $force_calculation = false ) {
 		$line_tax = is_callable( array( $item, 'get_total_tax' ) ) ? $item->get_total_tax() : 0;
 
-		if ( 0 === $line_tax ) {
+		if ( 0 == $line_tax ) {
 			return '-'; // no need to determine tax rate.
 		}
 
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -635,29 +635,27 @@ abstract class Order_Document_Methods extends Order_Document {
 
 	/**
 	 * Get the tax rates/percentages for an item
-	 * @param  object $item order item
-	 * @param  object $order WC_Order
-	 * @param  bool $force_calculation force calculation of rates rather than retrieving from db
+	 *
+	 * @param object $item order item
+	 * @param WC_Order|WC_Order_Refund $order WC_Order
+	 * @param bool $force_calculation force calculation of rates rather than retrieving from db
+	 *
 	 * @return string $tax_rates imploded list of tax rates
 	 */
-	public function get_tax_rate( $item, $order, $force_calculation = false ) {
-		$tax_data_container = ( $item['type'] == 'line_item' ) ? 'line_tax_data' : 'taxes';
-		$tax_data_key       = ( $item['type'] == 'line_item' ) ? 'subtotal' : 'total';
-		$line_total_key     = ( $item['type'] == 'line_item' ) ? 'line_total' : 'total';
-		$line_tax_key       = ( $item['type'] == 'shipping' ) ? 'total_tax' : 'line_tax';
-
-		$tax_class          = isset($item['tax_class']) ? $item['tax_class'] : '';
-		$line_tax           = $item[$line_tax_key];
-		$line_total         = $item[$line_total_key];
-		$line_tax_data      = $item[$tax_data_container];
+	public function get_tax_rate( object $item, $order, bool $force_calculation = false ): string {
+		$tax_class     = is_callable( array( $item, 'get_tax_class' ) ) ? $item->get_tax_class() : '';
+		$line_tax      = is_callable( array( $item, 'get_total_tax' ) ) ? $item->get_total_tax() : 0;
+		$line_total    = is_callable( array( $item, 'get_total' ) ) ? $item->get_total() : 0;
+		$line_tax_data = is_callable( array( $item, 'get_taxes' ) ) ? $item->get_taxes() : 0;
+		$tax_data_key  = ( 'line_item' === $item->get_type() ) ? 'subtotal' : 'total';
 
 		// first try the easy wc2.2+ way, using line_tax_data
-		if ( !empty( $line_tax_data ) && isset($line_tax_data[$tax_data_key]) ) {
-			$tax_rates = array();
+		if ( ! empty( $line_tax_data ) && isset( $line_tax_data[ $tax_data_key ] ) ) {
+			$tax_rates  = array();
+			$line_taxes = $line_tax_data[ $tax_data_key ];
 
-			$line_taxes = $line_tax_data[$tax_data_key];
 			foreach ( $line_taxes as $tax_id => $tax ) {
-				if ( isset($tax) && $tax !== '' ) {
+				if ( isset( $tax ) && ! empty ( $tax ) ) {
 					$tax_rate = $this->get_tax_rate_by_id( $tax_id, $order );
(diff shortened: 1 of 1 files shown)


includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -674,32 +672,31 @@ public function get_tax_rate( $item, $order, $force_calculation = false ) {
 				}
 			}
 
-			$tax_rates = implode(', ', $tax_rates );
-			return $tax_rates;
+			return implode( ', ', $tax_rates );
 		}
 
-		if ( $line_tax == 0 ) {
+		if ( 0 === $line_tax ) {
```
alexmigf, line 678: The `get_total_tax()` function returns a `string`, but maybe we should also consider that the value could be a `float`?

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -635,29 +635,27 @@ public function get_order_items() {
...
+		$tax_class     = is_callable( array( $item, 'get_tax_class' ) ) ? $item->get_tax_class() : '';
+		$line_tax      = is_callable( array( $item, 'get_total_tax' ) ) ? $item->get_total_tax() : 0;
+		$line_total    = is_callable( array( $item, 'get_total' ) ) ? $item->get_total() : 0;
+		$line_tax_data = is_callable( array( $item, 'get_taxes' ) ) ? $item->get_taxes() : 0;
+		$tax_data_key  = ( 'line_item' === $item->get_type() ) ? 'subtotal' : 'total';
 
 		// first try the easy wc2.2+ way, using line_tax_data
-		if ( !empty( $line_tax_data ) && isset($line_tax_data[$tax_data_key]) ) {
-			$tax_rates = array();
+		if ( ! empty( $line_tax_data ) && isset( $line_tax_data[ $tax_data_key ] ) ) {
```
alexmigf, line 653: Maybe we don't need this code part anymore?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document-methods.php
MohamadNateqi: I think we always have a `%` sign. In line `662` and also in `679`.

MohamadNateqi: @alexmigf Do you think I should consider `float` too?

alexmigf: In the test I made the `$line_tax` is a `string`with a decimal value in there, no percentage there, so I think it's safer to consider also the `float`.

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -635,71 +635,41 @@ public function get_order_items() {
 
 	/**
 	 * Get the tax rates/percentages for an item
-	 * @param  object $item order item
-	 * @param  object $order WC_Order
-	 * @param  bool $force_calculation force calculation of rates rather than retrieving from db
-	 * @return string $tax_rates imploded list of tax rates
+	 *
+	 * @param object $item order item
+	 * @param WC_Order|WC_Order_Refund $order WC_Order
```
alexmigf, line 640: I think we could use `\WC_Abstract_Order` instead?

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -635,71 +635,41 @@ public function get_order_items() {
...
-				}
-			}
-
-			// apply decimal setting
-			if ( function_exists( 'wc_get_price_decimal_separator' ) ) {
-				foreach ( $tax_rates as &$tax_rate ) {
-					$tax_rate = ! empty( $tax_rate ) ? str_replace( '.', wc_get_price_decimal_separator(), strval( $tax_rate ) ) : $tax_rate;
-				}
-			}
+	public function get_tax_rate( object $item, $order, bool $force_calculation = false ) {
```
alexmigf, line 645: Type hint for `$order` and return.

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -635,71 +635,41 @@ public function get_order_items() {
...
-	 * @param  object $item order item
-	 * @param  object $order WC_Order
-	 * @param  bool $force_calculation force calculation of rates rather than retrieving from db
-	 * @return string $tax_rates imploded list of tax rates
+	 *
+	 * @param object $item order item
+	 * @param WC_Order|WC_Order_Refund $order WC_Order
+	 * @param bool $force_calculation force calculation of rates rather than retrieving from db
+	 *
+	 * @return string|float $tax_rates imploded list of tax rates
```
alexmigf, line 643: I think it will be always `string`?

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -635,71 +635,41 @@ public function get_order_items() {
...
+		if ( 0 == $line_tax ) {
+			return '-'; // no need to determine tax rate.
 		}
 
-		if ( $line_tax == 0 ) {
-			return '-'; // no need to determine tax rate...
-		}
+		$tax_rates = '';
 
 		if ( ! apply_filters( 'wpo_wcpdf_calculate_tax_rate', false ) ) {
```
alexmigf, line 654: Maybe we could return earlier if this filter returns `true`?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document-methods.php
MohamadNateqi: Yeah, it seems good. I thought maybe being more explicit was better.

includes/documents/abstract-wcpdf-order-document-methods.php
MohamadNateqi: Yes. I didn't understand the previous comment well and thought it was about this. Fixed.

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -637,40 +637,38 @@ public function get_order_items() {
...
+	 * @param WC_Abstract_Order $order WC_Order
 	 * @param bool $force_calculation force calculation of rates rather than retrieving from db
 	 *
-	 * @return string|float $tax_rates imploded list of tax rates
+	 * @return string $tax_rates imploded list of tax rates
 	 */
-	public function get_tax_rate( object $item, $order, bool $force_calculation = false ) {
+	public function get_tax_rate( object $item, object $order, bool $force_calculation = false ): string {
+		if ( apply_filters( 'wpo_wcpdf_calculate_tax_rate', false ) ) {
+			return '';
```
alexmigf, line 647: Maybe `'-'`?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document-methods.php
alexmigf: You are returning it just below: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/722/files/71f8cd06f5db0f205acf55083a26e6e50e093563..dd560f14b7bd8c330534e39015a0e1f0c4204f24#diff-3c27d95148592338471e6af2989845953abad29560a97a4e73b989aaa9d1253fR653

alexmigf (approved):
LGTM!

includes/documents/abstract-wcpdf-order-document-methods.php
MohamadNateqi: Yes, I assumed that the purpose of this filter was to bypass the method's calculation in favor of an alternative calculation method hence it should return nothing. This was just a thought because I couldn't find any examples of its use.

PR #723, opened 2024-02-16
New: bumps WooCommerce min version to 3.3

alexmigf:
Following the same bump for the next Pro extension release, I think we should also do the same here.

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -953,23 +953,6 @@ class Admin {
 	}
 
 	/**
-	 * Add actions to menu, legacy method
-	 */
-	public function bulk_actions_js() {
-		if ( $this->is_order_page() ) {
-			?>
-			<script type="text/javascript">
-			jQuery(document).ready(function() {
-				<?php foreach (wcpdf_get_bulk_actions() as $action => $title) { ?>
-				jQuery('<option>').val('<?php echo esc_attr( $action ); ?>').html('<?php echo esc_attr( $title ); ?>').appendTo("select[name='action'], select[name='action2']");
-				<?php }	?>
-			});
-			</script>
-			<?php
-		}
-	}
-
-	/**
 	 * Save invoice number date
 	 */
 	public function save_invoice_number_date( $order_id, $order ) {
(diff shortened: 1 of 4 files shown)


MohamadNateqi (changes requested):
These can also be removed:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/96fef46aa1a28b8979680fea2dc2eb0dd717be21/includes/class-wcpdf-admin.php#L47-L49

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/96fef46aa1a28b8979680fea2dc2eb0dd717be21/includes/documents/abstract-wcpdf-order-document-methods.php#L791-L793

MohamadNateqi (approved):
LGTM!

MohamadNateqi:
This also needs to be updated:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/74b54ff386be9b65b7695e4cadb3bd98e4895631/woocommerce-pdf-invoices-packingslips.php#L12

MohamadNateqi:
I understood that the `get_order_discount` method used below no longer exists in the Woo plugin. We need to remove them in this PR or maybe another:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/96fef46aa1a28b8979680fea2dc2eb0dd717be21/includes/documents/abstract-wcpdf-order-document-methods.php#L965-L972
All of the usage of this method which is in the Template plugin is only for getting `total`.

<hr>

Maybe this also should be removed:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/96fef46aa1a28b8979680fea2dc2eb0dd717be21/includes/documents/abstract-wcpdf-order-document-methods.php#L228-L232

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/96fef46aa1a28b8979680fea2dc2eb0dd717be21/includes/documents/abstract-wcpdf-order-document-methods.php#L241-L244

alexmigf:
> Maybe this also should be removed:

I think we should keep, it's using getters.

alexmigf:
> I understood that the get_order_discount method used below no longer exists in the Woo plugin. We need to remove them in this PR or maybe another:

Done.

MohamadNateqi (approved):
Perfect!

PR #733, opened 2024-02-22
New: implements date type selector on Danger tools

### LINKED ISSUE #732: New date selector for the danger zone tools
We added a new filter hook (`wpo_wcpdf_process_danger_zone_tools_date_type`) to set the date type for the danger zone tools (#731).

However, I think that adding a new date selector will improve the user experience even more.

alexmigf:
closes #732


includes/wcpdf-functions.php
```diff
@@ -507,3 +507,23 @@ function wcpdf_safe_redirect_or_die( $url = '', $message = '' ) {
...
+ * 
+ * @param array $wp_query_args
+ * @param array $query_args
+ * @param object $order_store_cpt
+ *
+ * @return array
+ */
+function wpo_wcpdf_custom_document_date_query_var( array $wp_query_args, array $query_vars, $order_store_cpt ): array {
+	foreach ( WPO_WCPDF()->documents->get_documents() as $document ) {
+		if ( isset( $query_vars[ "wcpdf_{$document->slug}_date" ] ) && '' !== $query_vars[ "wcpdf_{$document->slug}_date" ] ) {
```
MohamadNateqi, line 522: I think the `empty()` method can be used instead.

includes/wcpdf-functions.php
```diff
@@ -507,3 +507,23 @@ function wcpdf_safe_redirect_or_die( $url = '', $message = '' ) {
...
+/**
+ * Allows `wc_get_orders()` queries to include document dates
+ * 
+ * @param array $wp_query_args
+ * @param array $query_args
+ * @param object $order_store_cpt
+ *
+ * @return array
+ */
+function wpo_wcpdf_custom_document_date_query_var( array $wp_query_args, array $query_vars, $order_store_cpt ): array {
```
MohamadNateqi, line 520: Maybe type hinting the object as `\WC_Order_Data_Store_CPT`?

MohamadNateqi (changes requested):
I renumbered the invoice for a specified date; it renumbered 44 documents.
Then I ran this query and only got 3 rows:
image

Then, I disabled HPOS and ran the renumbering tool. It couldn't find any document to renumber.
Again, this query returned 3 rows:
image

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -493,16 +493,15 @@ private function reset_settings( $data ) {
...
 			$message = __( 'One or more request parameters missing.', 'woocommerce-pdf-invoices-packing-slips' );
-			wp_send_json_error( compact( $message ) );
+			wp_send_json_error( compact( 'message' ) );
 		}
 	
-		$from_date          = date_i18n( 'Y-m-d', strtotime( $request['date_from'] ) );
-		$to_date            = date_i18n( 'Y-m-d', strtotime( $request['date_to'] ) );
-		$date_type          = apply_filters( 'wpo_wcpdf_process_danger_zone_tools_date_type', 'date_created', $request );
+		$from_date          = strtotime( $request['date_from'] );
+		$to_date            = strtotime( $request['date_to'] );
```
MohamadNateqi, line 504: HPOS is disabled. I have 3 orders from the 19th to the 22nd.
Since the time is not given to the `$to_date`, it considers the time as `00:00:00`, and so a document that I have in "February 22, 2024, 1:26:34 PM" will not be considered to be renumbered. Maybe it's good to add `. ' 23:59:59'` at the end of this?

MohamadNateqi (changes requested):
When the HPOS is on, it renumbers 81 rows which is not correct.

alexmigf:
> When the HPOS is on, it renumbers 81 rows which is not correct.

Was this related to [this](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/733#discussion_r1499613812) comment?

MohamadNateqi:
> Is this related to https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/733#discussion_r1499613812 comment?

I think not. The issue on that comment was for when the HPOS was off, and it renumbered 2 rows instead of 3, which shows the query was working correctly and the issue was only with the `$to_date`.

MohamadNateqi (changes requested):
It's working correctly now. But what do you think about creating an instance of the `WC_Order_Data_Store_CPT` class to use the `parse_date_for_wp_query` method?

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -520,10 +519,45 @@ public function ajax_process_danger_zone_tools() {
...
+			$message = __( 'Wrong date type selected.', 'woocommerce-pdf-invoices-packing-slips' );
+			wp_send_json_error( compact( 'message' ) );
+		}
+		
+		$args[ $date_arg ] = $from_date . '...' . $to_date;
+		
+		if ( WPO_WCPDF()->order_util->custom_orders_table_usage_is_enabled() ) {
+			$args = wpo_wcpdf_parse_document_date_for_wp_query( $args, $args );
+		} else {
+			add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', 'wpo_wcpdf_parse_document_date_for_wp_query', 10, 2 );
```
MohamadNateqi, line 551: I think we can directly call that method for both types and remove this filter here.

alexmigf: The `meta_query` was introduced only on `6.9.0` I believe ( I thought it was with HPOS), so I think it will not work on older versions of Woo.

MohamadNateqi: Isn't that PR for before releasing the HPOS?
Also, I tested with WooCommerce 6.8.0, and the `wpo_wcpdf_parse_document_date_for_wp_query` was working correctly.

alexmigf: Fixed.

MohamadNateqi (approved):
LGTM!

YordanSoares (changes requested):
Some remarks after running the tests:

1. When activated HPOS, the "Document date" option seems not to be respecting the date but applying to all the documents:

   For instance, I tried setting a date in the future:
   
   image

   ...and it apply the renumbering to all the documents:

   image
2. The same is happening to the "All" option for the deleting numbering tool. In this case, it is happening for both legacy and HPOS.
3. Finally, the "All" option for the deleting numbering tool is not including the generated **credit notes**.

YordanSoares (commented):
@alexmigf everything is working well now!

I'm just being a bit picky here, but I just realized that you can select the "Document date" for the "All" option within the deleting exiting document tool, if you select the date type first:

delete-all-documents

However, it returns the "Wrong data type selected" error. In other words, this is just a UI issue.

This is an edge case (I don't think people normally* will select the date type first), but I wanted to mention it, just in case.

\*_Fun fact: I often read magazines and newspapers from back to front!_ 😅

YordanSoares (approved):
LGTM! 🚀

PR #734, opened 2024-02-23
Fix: JS error `$(...).pointer` is not a function

alexmigf:
Fixes console error:

> Uncaught TypeError: "$(...).pointer" is not a function


MohamadNateqi (approved):
LGTM!

PR #737, opened 2024-02-28
Fix "get_tax_rate" issue when the tax class is "inherit"

### LINKED ISSUE #736: Fix "get_tax_rate" issue when tax class is "inherit"
For shipping item, the value of the `get_tax_class()` method is "inherit". Sending this string as the tax class to the `get_rates()` method will return an empty array instead of returning taxes correctly.

This shipping row indicates the tax rate and it is not divided in the same way as the other rows:
image

MohamadNateqi:
close #736


alexmigf (approved):
LGTM!

PR #739, opened 2024-02-29
Add hooks to allow adding new tabs and their contents

### LINKED ISSUE #738: Add hooks to the "Settings_Debug" class

MohamadNateqi:
close #738

### DIFF
--- a/includes/settings/class-wcpdf-settings-debug.php
+++ b/includes/settings/class-wcpdf-settings-debug.php
@@ -22,13 +22,6 @@ class Settings_Debug {
 	}
 
 	public function __construct()	{
-		$this->sections = array(
-			'settings' => __( 'Settings', 'woocommerce-pdf-invoices-packing-slips' ),
-			'status'   => __( 'Status', 'woocommerce-pdf-invoices-packing-slips' ),
-			'tools'    => __( 'Tools', 'woocommerce-pdf-invoices-packing-slips' ),
-			'numbers'  => __( 'Numbers', 'woocommerce-pdf-invoices-packing-slips' ),
-		);
-		
 		add_action( 'admin_init', array( $this, 'init_settings' ) );
 		add_action( 'wpo_wcpdf_settings_output_debug', array( $this, 'output' ), 10, 1 );
 		
@@ -38,7 +31,7 @@ class Settings_Debug {
 
 	public function output( $active_section ) {
 		$active_section = ! empty( $active_section ) ? $active_section : 'settings';
-		$sections       = apply_filters( 'wpo_wcpdf_settings_debug_sections', $this->sections, $active_section );
+		$sections       = $this->get_settings_sections();
 
 		?>
 		<div class="wcpdf_debug_settings_sections">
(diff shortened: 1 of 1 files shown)


includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -38,12 +38,13 @@ public function __construct()	{
 
 	public function output( $active_section ) {
 		$active_section = ! empty( $active_section ) ? $active_section : 'settings';
-		
+		$sections       = apply_filters( 'wpo_wcpdf_settings_debug_sections', $this->sections, $active_section );
```
alexmigf, line 41: Maybe this filter should be in the constructor?

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -66,6 +67,8 @@ public function output( $active_section ) {
 				$this->display_numbers();
 				break;
 		}
+
+		do_action( 'wpo_wcpdf_settings_debug_after_output', $active_section );
```
alexmigf, line 71: Maybe you could include this in the `switch` statement?

```php
default:
	do_action( 'wpo_wcpdf_settings_debug_section_output', $active_section );
	break;
```

alexmigf (changes requested):

includes/settings/class-wcpdf-settings-debug.php
MohamadNateqi: Since the constructor method will run on the `plugins_loaded` hook, If I put the filter there, it won't work. Can a filter work in this condition?

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a60611f709fd4952f5643e1cd23bbbde2967f56d/woocommerce-pdf-invoices-packingslips.php#L73

includes/settings/class-wcpdf-settings-debug.php
MohamadNateqi: I thought for more flexibility and usability, it could be better to put it outside of the switch statement. Is it better as a switch case?

alexmigf: You can have both.

includes/settings/class-wcpdf-settings-debug.php
MohamadNateqi: @alexmigf Sorry, I think maybe you missed my previous comment, so I'm mentioning you again.

alexmigf: I see. You could create a method to get them with the filter instead of setting in the constructor.

alexmigf: Because otherwise, if you call the `$this->sections` in other place the filter doesn't apply.

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -66,6 +59,9 @@ public function output( $active_section ) {
 			case 'numbers':
 				$this->display_numbers();
 				break;
+			default:
+				do_action( 'wpo_wcpdf_settings_debug_section_output', $active_section );
+				break;
 		}
 
 		do_action( 'wpo_wcpdf_settings_debug_after_output', $active_section );
```
alexmigf, line 67: Maybe you could add all the sections here as well? `$sections`

alexmigf (approved):
LGTM!

includes/settings/class-wcpdf-settings-debug.php
MohamadNateqi: Done.

Issue #740, opened 2024-03-01
Wrong date for PDF filename when 2 or more documents are exported using the bulk actions.


YordanSoares:
[This user](https://wordpress.org/support/topic/incorrect-date-when-downloading-pdf/) reported when exporting PDF documents in bulk, using the bulk actions, the filename is not displaying the correct date for them when they export the document in the morning (they are from Australia).

PR #741, opened 2024-03-01
Replace `date()` with `date_i18n()` to set the date for the PDF filename (bulk action export)

YordanSoares:
Closes: #740


alexmigf (approved):
LGTM!

alexmigf:
@YordanSoares can you do also the same for the Pro extension?

PR #744, opened 2024-03-06
New: updates composer libraries, specially `php-svg-lib` because of a security issue

### LINKED ISSUE #743: Dompdf's usage of vulnerable version of phenx/php-svg-lib leads to restriction bypass and potential RCE
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/security/dependabot/7

alexmigf:
closes #743

This PR updates this third party libraries:

- phenx/php-font-lib (0.5.4 => 0.5.6)
- phenx/php-svg-lib (0.5.1 => 0.5.2)
- sabberworm/php-css-parser (8.4.0 => v8.5.1)

This is specially important because of a vulnerability found in `php-svg-lib` and reported [here](https://github.com/dompdf/dompdf/issues/3393).

In terms of testing we just need to make sure that the PDFs are still working as normal, including SVG images.


dwalkerpriv (approved):
Works in all the templates in the custom blocks and as a logo. 👍 

Screenshot 2024-03-06 134456


Screenshot 2024-03-06 134330

alexmigf:
@YordanSoares can you also test this?

YordanSoares (approved):
LGTM!

However, I just wanted to mention that some SVGs are not displayed in the correct size (this happens in the latest release as well), even if you set a specific width for them:

image
_In the screenshot above, the shop logo and a couple of the social media logos after the customer notes, are very tiny (the Instagram logo can't be seen at all)._

However, this is my HTML output for the same document:

image

I think it depends on the SVG code, but it's good to know that this format is still limited supported.

alexmigf:
@YordanSoares I think we could try to fix that in a new issue. Can you open it?

PR #746, opened 2024-03-11
Fix: Incorrect sorting orders for some column

### LINKED ISSUE #745: Sort filtered results on the Orders page
On the Orders page when filtering by an order status, the results are out date sequence. clicking on the sort tabs does nothing.

Related forum ticket
https://wordpress.org/support/topic/issue-with-order-list-sequence/

Once they have disabled our plugin the issue is resolved. additionally, when all plugins except WC and our plugin is disabled and using a default theme, the issue is still there. Only when our plugin is disabled does the issue go away.

MohamadNateqi:
close #745 

The issue only happens when the HPOS is disabled.

### DIFF
--- a/includes/class-wcpdf-admin.php
+++ b/includes/class-wcpdf-admin.php
@@ -1377,7 +1377,7 @@ class Admin {
 	}
 
 	public function sort_orders_by_numeric_invoice_number( $query ): void {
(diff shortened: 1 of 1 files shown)


alexmigf (approved):
LGTM!

dwalkerpriv (approved):
I didn't have the issue before, but there was not negative change with this PR. Sorting works 👍

YordanSoares (approved):
LGTM!

PR #747, opened 2024-03-13
HPOS: Add new filter to include invoice number in search queries

them-es:
When HPOS is enabled the filter `woocommerce_order_table_search_query_meta_keys` needs to be executed.
The other filter is only relevant when CPT orders are in use.

References:
https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableSearchQuery.php#L241-L254
https://stackoverflow.com/a/77752011


alexmigf (approved):
LGTM!

MohamadNateqi (approved):
LGTM!

alexmigf:
@them-es thank you for the suggestion, appreciated.

PR #751, opened 2024-03-18
Fix not removing documents meta when renew a subcription

### LINKED ISSUE #688: WooCommerce subscription compatibility
We've had a few threads/tickets (about 3) of users that are using the WooCommerce Subscription plugin who do not get their invoice numbers incremented on a subscription renewal. One user does not get their order date updated with the renewal date.

This issue only happens with the initially created subscription that has been renewed, it seems. An example from the forums https://wordpress.org/support/topic/subscriptions-renewal-order-issue/

A user mentioned this in a test

> If we check a subscription that is active since 2022, the first renewal (2023) has a new invoice number. The second renewal has the invoice number of the initial order (2022). So it seems like the bug is happening with subscriptions with 2 or more renewal orders attached to this subscription.

MohamadNateqi:
close #688 

This is related to this [ticket](https://secure.helpscout.net/conversation/2536211939/59199?viewId=7725013).

Issue description: The customer renewing their subscription. For some of the subscriptions, the renewed order showed that the invoice had been generated, but with the wrong invoice number (usually the parent order of the subscription).

It was discovered that the subscription post contained some meta data that was being copied to the renewed order, including the invoice number and date. The reason for this meta data being there is not clear, but we had a method developed to prevent such meta data from being copied to the renewed order, to prevent such issues.

However, we encountered a problem with this method as there was a typo in the hook name, which prevented it from working correctly. Additionally, the old deprecated hook used to send the `$meta` argument with a different data structure than the new hook, so it needs to be updated to be compatible with the new data structure.

The old deprecated hook had this data structure for the `$meta` argument: 
```php
array(..., array( "meta_key":"_wcpdf_invoice_number","meta_value":"158" ), ...)
```
And this is for the new hook:
```php
array(..., "_wcpdf_invoice_number":"158", ...)
```


MohamadNateqi:
Both customers confirmed their issue has been fixed with this change.

alexmigf (approved):
LGTM!

PR #754, opened 2024-03-20
Translatable shop logo

BrunoPavlinic98:


alexmigf (changes requested):

assets/js/media-upload.js
```diff
@@ -5,11 +5,24 @@ jQuery(document).ready(function($) {
 		
 	// Uploading files
 	var file_frame;
+	let $settings_wrapper;
```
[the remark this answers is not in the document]
BrunoPavlinic98: @alexmigf This needs to stay because we need the `$settings_wrapper` outside of `on` click scope so that it is properly updated on each '.wpo_upload_image_button' click. If I remove this and update the  `$settings_wrapper` to be let as you stated in the comment below, application will not work properly - it will replace the first language value with the second one when I update the second language value.

includes/settings/class-wcpdf-settings-callbacks.php
```diff
@@ -714,6 +717,23 @@ public function validate( $input ) {
...
+	 * Appends language at the end of the setting provided, in case the setting is translatable
+	 * and it does not have a language set.
+	 *
+	 * @param string $setting Settings field that needs a language.
+	 * @param array  $args Setting arguments.
+	 *
+	 * @return void
+	 */
+	public function append_language( string $setting, array $args ): string {
+		if ( isset( $args['translatable'] ) && isset( $args['lang'] ) && 'default' !== $args['lang'] && true == $args['translatable'] && ! str_ends_with( $setting, "[{$args['lang']}]" ) ) {
```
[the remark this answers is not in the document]
BrunoPavlinic98: @alexmigf good catch, I did not pay attention to that. I will try to think about that in the future.
I have made the same thing with `strlen` and `substr` methods, which are okay with versions PHP 4, PHP 5, PHP 7, PHP 8.

includes/settings/class-wcpdf-settings-callbacks.php
```diff
@@ -714,6 +717,23 @@ public function validate( $input ) {
...
+	 * Appends language at the end of the setting provided, in case the setting is translatable
+	 * and it does not have a language set.
+	 *
+	 * @param string $setting Settings field that needs a language.
+	 * @param array  $args Setting arguments.
+	 *
+	 * @return void
+	 */
+	public function append_language( string $setting, array $args ): string {
+		if ( isset( $args['translatable'] ) && isset( $args['lang'] ) && 'default' !== $args['lang'] && true == $args['translatable'] && ! ( substr( $setting, -strlen( "[{$args['lang']}]" ) ) === "[{$args['lang']}]" ) ) {
```
alexmigf, line 731: Maybe `true === $args['translatable']`?

BrunoPavlinic98: @alexmigf I left == because sometimes, `$args['translatable']` comes as a string `'true'`, and sometimes as a boolean. With ==, both cases are covered.

alexmigf: Just checking in—did you notice it uses `===` over [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e6496688b9484860f232eae1ce4fab90c4380cc0/includes/settings/class-wcpdf-settings-callbacks.php#L201)?

includes/settings/class-wcpdf-settings-callbacks.php
```diff
@@ -714,6 +717,23 @@ public function validate( $input ) {
...
 	}
+
+	/**
+	 * Appends language at the end of the setting provided, in case the setting is translatable
+	 * and it does not have a language set.
+	 *
+	 * @param string $setting Settings field that needs a language.
+	 * @param array  $args Setting arguments.
+	 *
+	 * @return void
```
alexmigf, line 728: Maybe `string`?

alexmigf (changes requested):

BrunoPavlinic98:
> Maybe we could replace [this](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/754/files#diff-810ed0f5ac5bea1698528d5035c7f8885e56809bbbee68602da884b33d9426beL46) with `jquery-blockui`?

I agree! I did not know about this block UI plugin, looks good!

includes/settings/class-wcpdf-settings-callbacks.php
BrunoPavlinic98: @alexmigf I did not notice that, interesting. I am not sure why sometimes it returns argument as a string and sometimes as a bool. But nevertheless, I think that true == should be OK for this case. What do you think?

includes/settings/class-wcpdf-settings-callbacks.php
BrunoPavlinic98: Ah yes, I forgot to change that! Thanks.

includes/settings/class-wcpdf-settings-callbacks.php
alexmigf: I think you need to find why that is happening.

BrunoPavlinic98: @alexmigf this is happening because PHP decodes bool values as strings via AJAX. We can solve this by checking if the translation is string and if yes, convert it to boolean, however, I think that == true is the best possible solution in this case because it will check if translation is true or "true", and both of those indicate that we want to translate the setting.
Do you have an alternative solution?

alexmigf: Yes, you could check if the value is a `string` and apply this function `wc_string_to_bool()`, otherwise just deal with the `boolean`.

alexmigf:
I'm getting this error in the `media_upload()` function after saving logos for 2 languages with this branch and then get back to the `main` branch:

> PHP Warning:  Array to string conversion

BrunoPavlinic98:
> I'm getting this error in the `media_upload()` function after saving logos for 2 languages with this branch and then get back to the `main` branch:
> 
> > PHP Warning:  Array to string conversion


> I'm getting this error in the `media_upload()` function after saving logos for 2 languages with this branch and then get back to the `main` branch:
> 
> > PHP Warning:  Array to string conversion

@alexmigf Can you please share more detailed error reproduction steps? I have tried the following with no errors:

1. Saved logos for all available languages on `375` branch.
2. Switched to `main` branch.
3. Reloaded the settings page - /wp-admin/admin.php?page=wpo_wcpdf_options_page
4. Saved the settings again.
5. Loaded some invoice.

alexmigf:
I'm still getting the error, steps to reproduce:

> Free PDF plugin on this branch
> Pro on this branch: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/387
> Enable WPML
> Upload 2 or more logos to different languages
> Disable WPML
> Switch free PDF branch to `main`
> Switch Pro extension to `main`

Error trace:

Screenshot from 2024-04-02 16-40-35

Note: you might need to add a version compatibility check in the Pro extension, because the changes you made in the `media_upload()` function aren't present in the older versions of the free plugin (current stable version and below).

alexmigf (changes requested):

alexmigf (changes requested):
If I set logos for languages using WPML, and after that I disable WPML, I still have the logo showing in the settings but in the PDF is not being displayed:

Screenshot from 2024-04-05 10-20-11

Even if I save the settings the logo doesn't display in the PDF, I need to remove it and set it again to work.

MohamadNateqi (changes requested):
I have a high-resolution image (1000 by 570 pixels - 965dpi), and this is what I see:
image
It is correct on the main branch.

<hr>

On the main branch, I removed general settings from the database (`delete_option( 'wpo_wcpdf_settings_general' );`). Then, I set up a logo and saved the settings. After enabling Polylang and switching to this PR branch, I refreshed the General tab page, but I couldn't see the logo I saved. I think it still needs to display the previous logo, but all tabs are empty:
image

BrunoPavlinic98:
@alexmigf @MohamadNateqi I think I managed to solve the issues you had. Please let me know what you think.

includes/settings/class-wcpdf-settings-callbacks.php
```diff
@@ -714,6 +717,23 @@ public function validate( $input ) {
...
+	 * and it does not have a language set.
+	 *
+	 * @param string $setting Settings field that needs a language.
+	 * @param array  $args Setting arguments.
+	 *
+	 * @return string
+	 */
+	public function append_language( string $setting, array $args ): string {
+		if ( isset( $args['translatable'] ) && isset( $args['lang'] ) && 'default' !== $args['lang'] && true == $args['translatable'] && ! ( substr( $setting, -strlen( "[{$args['lang']}]" ) ) === "[{$args['lang']}]" ) ) {
+			return $setting .= "[{$args['lang']}]";
```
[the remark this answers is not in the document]
BrunoPavlinic98: @MohamadNateqi Can you be more specific, I am not sure I understand what do you mean.
If you mean that it is not necessary to use append language function on the `$setting_name `variable, this is necessary because after I set the new image as logo, the media_callback function will not automatically assign the language to the settings name, and therefore, the setting will be saved in the database in the wrong way.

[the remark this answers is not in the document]
MohamadNateqi: Sure, sorry.
I meant you can directly return the value, I mean like this:
```php
return "[{$args['lang']}]";
```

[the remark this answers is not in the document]
BrunoPavlinic98: @MohamadNateqi But then, only the language will be returned. I need to return the setting name with the language appended at the end. I think this needs to stay as is. Am I missing something?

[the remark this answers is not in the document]
MohamadNateqi: Sorry, I missed `$setting` in my previous code. I meant this:
```php
return $setting . "[{$args['lang']}]";
```
But not important, no need to change.

alexmigf (approved):
LGTM!

assets/js/media-upload.js
```diff
@@ -88,5 +109,5 @@ jQuery(document).ready(function($) {
 		$preview.remove();
 		$( this ).remove();
 		$( '.attachment-resolution, .attachment-resolution-warning' ).remove();
```
MohamadNateqi, line 111: I think this should also be updated. When I remove the image for one language, the notice will be removed in all tabs.

MohamadNateqi (changes requested):

assets/js/media-upload.js
BrunoPavlinic98: @MohamadNateqi Nice catch, well done! I have updated the code.

MohamadNateqi (approved):
LGTM!

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -818,8 +818,14 @@ public function has_header_logo() {
 	 * Return logo id
 	 */
 	public function get_header_logo_id() {
-		if ( !empty( $this->settings['header_logo'] ) ) {
-			return apply_filters( 'wpo_wcpdf_header_logo_id', $this->settings['header_logo'], $this );
+		if ( ! empty( $this->settings['header_logo'] ) ) {
+			$header_logo_id = $this->get_settings_text( 'header_logo', '', false );
+
+			if ( empty( $header_logo_id ) ) {
+				$header_logo_id = $this->settings['header_logo'];
```
alexmigf, line 825: I think `$this->settings['header_logo']` can be an `array`? And in that case this will throw error [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f312b8657ee4ec91a9caa940a8ce62160882a668/includes/documents/abstract-wcpdf-order-document.php#L850)?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document.php
BrunoPavlinic98: > I think `$this->settings['header_logo']` can be an `array`? And in that case this will throw error [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/f312b8657ee4ec91a9caa940a8ce62160882a668/includes/documents/abstract-wcpdf-order-document.php#L850)?

This should not happen because `$this->get_settings_text( 'header_logo', '', false );` should handle the array version of `$this->settings['header_logo'];`. If that fails, and `$header_logo_id` is still empty, this means that we don't have a language version of logo, and therefore, the value should not be an array.

@alexmigf can you please describe what are the steps you have taken to get the error described above?

alexmigf: I see. I'm not able to reproduce the error above again.

Maybe we could remove this part? Since it's handled by the `get_settings_text` function.

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/cf4d8cfdfb921bf8d8590ac82a236ca75a13521a/includes/documents/abstract-wcpdf-order-document.php#L824-L826

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -818,8 +818,14 @@ public function has_header_logo() {
...
+			}
+			
+			return apply_filters( 'wpo_wcpdf_header_logo_id', $header_logo_id, $this );
```
alexmigf, line 828: Maybe this should return `int` if not `empty` and `false` otherwise?

includes/documents/abstract-wcpdf-order-document.php
alexmigf: Actually, after remove the 2 language logos I had, disable WPML and change this branch to `main` I'm now getting this error:

Screenshot from 2024-04-10 12-15-40

After I save a new logo the error is gone.

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document.php
BrunoPavlinic98: @alexmigf I have removed it, and after testing, I realized I need to make a slight change to the `get_settings_text` method. It should be working correctly now.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -819,10 +819,10 @@ public function has_header_logo() {
 	 */
 	public function get_header_logo_id() {
 		if ( ! empty( $this->settings['header_logo'] ) ) {
-			$header_logo_id = $this->get_settings_text( 'header_logo', '', false );
+			$header_logo_id = intval( $this->get_settings_text( 'header_logo', '', false ) );
```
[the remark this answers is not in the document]
BrunoPavlinic98: @alexmigf I have used `absint` instead, however, I think that other checks are not needed. If `absint` gets an invalid or empty string as an input, it will return 0. And if `$header_logo_id` is 0, we will assign it to false because then, the logo is not set.

What do you think?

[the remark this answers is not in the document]
alexmigf: Seems fine to me.

BrunoPavlinic98:
> Great work!

Thank you for the great review!

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -884,6 +888,8 @@ public function get_settings_text( $settings_key, $default = false, $autop = tru
 		// fallback to first array element if default is not present
 		} elseif( ! empty( $setting ) && is_array( $setting ) ) {
 			$text = reset( $setting );
+		} else {
+			$text = $setting;
```
MohamadNateqi, line 892: May I ask why this part has been added here?

BrunoPavlinic98: @MohamadNateqi yes sure. The `get_settings_text` method gets the text in the specific language. In case we don't have a multilingual setup, the `$this->get_setting` method won't return an array, but one string value. Therefore, in case the `$setting` is not an array, assign it to the `$text` variable. In case the `$text` variable stays empty all the way, we will assign it to the default value provided through function parameter.

In short, I have modified this method so that it works in cases multilingual setups are not enabled.

MohamadNateqi: Thanks.

MohamadNateqi (approved):
Perfect!

PR #755, opened 2024-03-21
Fix: issues rendering the number tab results 

### LINKED ISSUE #752: Numbers tab exhausting memory limit
A [customer](https://wordpress.org/support/topic/numbers-tab-not-loading-error/#post-17509432) reported issues when trying to load the Numbers tab. Note that this user has 26k orders, which might be the reason for the memory exhausting issue.

> An error of type E_ERROR was caused in line 1168 of the file /var/www/website.com/wp-content/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableDataStore.php. Error message: Allowed memory size of 536870912 bytes exhausted (tried to allocate 266240 bytes)

alexmigf:
closes #752 

This pull request aims to enhance the performance of the number tables query by caching the data in an option. This approach ensures that the data can be reused in subsequent requests, thus reducing the load on the store.

Additionally, I've introduced a new selector that allows for limiting the query results. This is particularly useful if there's a need to retrieve, for instance, only the most recent 100 numbers.

The implementation also includes functionality to track and display the last fetch date. This feature is instrumental in determining whether there's a need for a fresh fetch of data.

The cached data is designed to be deletable, ensuring flexibility and independence across different tables.

Furthermore, I've made minor stylistic adjustments to the table, especially within the `type` column. For unassigned numbers, the column now displays `gapped`. In the context of `credit-note` tables, it also shows the refund order ID, enhancing clarity and information accessibility.

The functionality for number filtering and search has been preserved.

Screenshot from 2024-03-21 16-20-13


BrunoPavlinic98 (changes requested):

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -904,6 +906,144 @@ private function get_settings_sections(): array {
...
+		$total_limit   = 'all' !== $limit ? absint( $limit ) : 0;
+		$total_fetched = absint( $total_fetched ?? 0 );
+		$chunk_size    = 0 !== $total_limit && $total_limit < 100 ? $total_limit : absint( $chunk_size ?? 100 );
+		$offset        = absint( $offset ?? 0 );
+		$option_name   = "wpo_wcpdf_number_data::{$table_name}";
+		$results       = get_option( $option_name, array() );
+		$hook          = 'wpo_wcpdf_number_table_data_fetch';
+
+		while ( true ) {
+			$query         = $wpdb->prepare( "SELECT * FROM {$table_name} ORDER BY {$orderby} {$order} LIMIT %d OFFSET %d", $chunk_size, $offset );
```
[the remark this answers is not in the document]
alexmigf: You can only use placeholder for data values: https://developer.wordpress.org/reference/classes/wpdb/prepare/

BrunoPavlinic98 (approved):
LGTM!

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -904,6 +906,144 @@ private function get_settings_sections(): array {
...
+		}
+	}
+	
+	public function ajax_numbers_data() {
+		check_ajax_referer( 'wpo_wcpdf_debug_nonce', 'nonce' );
+		
+		$request = stripslashes_deep( $_POST );
+		
+		if ( isset( $request['action'] ) && 'wpo_wcpdf_numbers_data' === $request['action'] && isset( $request['operation'] ) && isset( $request['table_name'] ) ) {
+			extract( $this->filter_fetch_request_data( $request ) );
```
MohamadNateqi, line 964: WPCS - using `extract()`

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -904,6 +906,144 @@ private function get_settings_sections(): array {
...
+	public function delete_number_table_data( $table_name ) {
+		if ( empty( $table_name ) ) {
+			return;
+		}
+		
+		delete_option( "wpo_wcpdf_number_data::{$table_name}" );
+		delete_option( "wpo_wcpdf_number_data::{$table_name}::last_time" );
+	}
+	
+	public function search_number_in_table_data( $table_name, $search ) {
```
MohamadNateqi, line 996: I think it would be better if the search happens directly in the database instead of fetched data.
For a customer with hundreds of orders, it might need to fetch all orders and then perform the search, and also, if they fetch some orders and search and get no result, it could be confusing for them as they think they don't have the searched number.

includes/tables/class-wcpdf-number-store-list-table.php
```diff
@@ -189,45 +189,36 @@ public function get_search( $request ) {
...
-		// MySQL int range
-		$options = array(
-			'options' => array(
-				'min_range' => 1,
-				'max_range' => 4294967295
-			)
-		);
+		if ( as_has_scheduled_action( 'wpo_wcpdf_number_table_data_fetch' ) ) {
+			return array();
+		}
```
MohamadNateqi, line 209: Can be merged.

MohamadNateqi (changes requested):

includes/settings/class-wcpdf-settings-debug.php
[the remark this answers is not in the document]
alexmigf: `$total_limit` can be `0`

includes/settings/class-wcpdf-settings-debug.php
MohamadNateqi: Great idea!

includes/settings/class-wcpdf-settings-debug.php
```diff
@@ -904,6 +906,233 @@ private function get_settings_sections(): array {
...
+		
+		$data = $this->filter_fetch_request_data( $input_data );
+		
+		if ( empty( $data['table_name'] ) ) {
+			return;
+		}
+		
+		$total_limit   = 'all' !== $data['limit'] ? absint( $data['limit'] ) : 0;
+		$total_fetched = absint( $total_fetched ?? 0 );
+		$chunk_size    = 0 !== $total_limit && $total_limit < $chunk_size ? min( $total_limit - $offset, $chunk_size ) : absint( $chunk_size ?? 100 );
```
MohamadNateqi, line 941: Considering the `$total_limit` is `110`, and the default `$chunk_size` is `100`, for the first call to this method, it'll get 100 rows, and for the second call to this method, it'll get 100 rows instead of just 10, since `$total_limit < $chunk_size`.

MohamadNateqi (changes requested):

MohamadNateqi (approved):
LGTM!

includes/views/advanced-numbers.php
```diff
@@ -32,13 +32,75 @@
...
+				<p><?php _e( 'Given the potential impact of querying a large volume of orders on site performance, it\'s essential to fetch data each time you need the most current information. This procedure ensures that the site remains efficient and responsive, even when handling substantial order quantities.', 'woocommerce-pdf-invoices-packing-slips' ); ?></p>
+				<?php if ( ! empty( $last_fetch ) ) : ?>
+					<p><strong><?php _e( 'Last fetch', 'woocommerce-pdf-invoices-packing-slips' ); ?>: </strong><code><?php echo date_i18n( 'Y-m-d H:i:s', $last_fetch ); ?></code></p>
+					<?php if ( $last_fetch > strtotime( 'today 23:59:59' ) ) : ?>
+						<div class="notice notice-warning inline"><p><?php _e( 'The displayed data may not be current. To ensure you have the most recent information, you might want to fetch updated data.', 'woocommerce-pdf-invoices-packing-slips' ); ?></p></div>
+					<?php endif; ?>
+				<?php endif; ?>
+				<p>
+					<?php
+						$limit_options = array(
```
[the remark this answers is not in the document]
alexmigf: I agree, maybe it's more intuitive for the user.

MohamadNateqi (approved):
LGTM!

dwalkerpriv:
The update seems to be working so far. I've noticed these documents in the drop down. Did we add these in the update?

Screenshot 2024-03-27 134725

alexmigf:
@dwalkerpriv no, those are tables that were created in your local setup, probably on tests you made for customers, for example using code snippets.

dwalkerpriv (approved):

includes/tables/class-wcpdf-number-store-list-table.php
```diff
@@ -55,54 +55,71 @@ public function __construct() {
...
+								unset( $document_types[ $key ] );
+							}
+						}
+					}
+					
+					$document_type = reset( $document_types );
+					
+					if ( ! empty( $document_type ) ) {
+						$document_slug = str_replace( '-', '_', $document_type );
+						$saved_number  = $order->get_meta( "_wcpdf_{$document_slug}_number", true );
```
Terminator-Barbapapa, line 83: All my invoices are marked as gapped as `_wcpdf_{$document_slug}_number` returns the formatted number and `$item->id` returns the unformatted number.

Terminator-Barbapapa (changes requested):
I'm not sure if using AS is the best way of doing this. As the page has to be reloaded to see actual results, which is not a good UX.

alexmigf:
> I'm not sure if using AS is the best way of doing this. As the page has to be reloaded to see actual results, which is not a good UX.

What alternatives would you recommend? I believe that in large stores, waiting for page data can also be a lengthy and unpleasant experience.

alexmigf:
@dpeyou 

> I am not getting the textbox to limit the fetch.

It has been replaced with the date interval selectors that you now see.

> The time of the fetch is in UTC, but without indication it is UTC time. That could be a source of confusion.

The `date_i18n` function is being used without specifying the GMT timezone. In this case, it defaults to using the WordPress site's timezone settings.

dpeyou (approved):
From a functional standpoint, I cannot find an issue :smile:

YordanSoares (approved):
LGTM! The query worked as expected, although I'm agree with Michael that would be nice if the result could be displayed without the need to reload the plage. That said, I also think that this could be improved later :)

alexmigf:
> ...although I'm agree with Michael that would be nice if the result could be displayed without the need to reload the page.

I understand. We've handled similar implementations before, like with the Bulk Export feature. The new code uses the Action Scheduler to efficiently manage stores with a large number of orders. This method was chosen to avoid the drawbacks of using AJAX, particularly in larger stores. With AJAX, users would have to wait, and if they closed the tab, the script would stop running, leading to a poor user experience. By using the Action Scheduler, users can leave the page and come back later to view the results if the process takes too long.

alexmigf:
We've had a customer using this feature for a few weeks now, and the feedback has been positive. They can easily view numbers for a month or even the entire year without any issues. While I agree that the UX could be improved, I believe it's better to merge this now and let customers use it, as their feedback will be valuable. We can always revisit and refine it later.

PR #757, opened 2024-03-28
Improvements for "PDF document data" metabox

### LINKED ISSUE #756: Improvements for "PDF document data" metabox
This https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/374 had the intention to solve the issue of not displaying generated document data in the PDF document data metabox, but it doesn't fix the underlying issue

This issue can still be reproduced when, for instance, three documents are created automatically for a specific order. Changing the priority doesn't make a difference because the issue is that these documents are being generated via AS. When changing the order status via the order edit page, these documents haven't been generated yet when the page reloads.

The only solution would be to check for pending AS actions on page loading. And display a message like 'Invoice is being generated in the background. Please reload the page to see the document data.'

It would be even better to run a callback when the action is finished and load the document data via AJAX.

This issue also has been mentioned here: https://github.com/wpovernight/woocommerce-pdf-ips-pro/issues/388.

### LINKED ISSUE #760: When an invoice is deleted, any associated receipts become inaccessible.
When an invoice is deleted, any associated receipts become inaccessible.

Screenshot from 2024-04-02 17-50-11

Screenshot from 2024-04-02 17-51-10

MohamadNateqi:
close #756
close #760 

**Note**: For the Pro extension compatibility, this branch should be used: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392.

<hr>

This PR improves the way document data are displayed in the metabox. It includes a check for pending jobs and displays a message if the document is still in the queue.

It also uses WordPress heartbeat to update document data via AJAX, eliminating the need to refresh the page. However, for this to work correctly with Pro documents, the Pro plugin requires some improvements that have been done in the PR mentioned above.

For test purposes, the below snippet can be used to stop the queue from running jobs:
```php
function wpo_disable_default_runner() {
	if ( class_exists( 'ActionScheduler' ) ) {
		remove_action( 'action_scheduler_run_queue', array( ActionScheduler::runner(), 'run' ) );
	}
}

add_action( 'init', 'wpo_disable_default_runner', 10 );
```
Then, the job can be manually run to test the AJAX update.

<hr>

This PR also fixes the issue that prevents displaying a dependent document if it already exists. If the invoice document is a requirement for another document (such as the credit note), removing the invoice from the "PDF document data" meta box will prevent the other dependent document from being displayed. For example, when a credit note or receipt document exists, removing the invoice prevents displaying those documents, even though they've already been generated.

This issue happens on both the AJAX request and page refresh/load (AJAX request also [gets the whole page content](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/14c6a767fc2c557c5e83806c373554e0eccf0f14/assets/js/order-script.js#L134) the same as page load and refreshes the meta box with new content) because the invoice is a requirement for those documents, and the `is_allow()` method checks for that and returns `false` when it cannot find an invoice.

A [new filter](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/files#diff-0d342e77085acd3923be230ea36e354d74d4232e687d0bc4f8aafe3eecc3fe0aR722) has been added to address this issue, which needs [this filter](https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392/files#diff-1aaf16af4dbfbd03af9365d3d5500e1e5e75a84b1e49aa025a71eb5b38875578R1366) in the Pro extension.


alexmigf (changes requested):

includes/class-wcpdf-admin.php
```diff
@@ -1391,14 +1262,87 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+	/**
+	 * Append ready documents to the heartbeat response.
+	 *
+	 * @param array $response
+	 * @param array $data
+	 *
+	 * @return array
+	 */
+	public function handle_heartbeat_received( array $response, array $data ): array {
+		if ( empty( $data['wpo_pending_documents'] ) || empty( $data['wc-refresh-order-lock'] ) ) {
```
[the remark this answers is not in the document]
MohamadNateqi: I used the `||` in the condition to prevent the execution of the rest of the function if either is empty. 
So this should early return if the `$data['wc-refresh-order-lock']` is empty and the `$data['wpo_pending_documents']` is not. Wouldn't it?

includes/views/document-data-metabox.php
```diff
@@ -0,0 +1,141 @@
...
+			<?php endif; ?>
+		</h4>
+
+		<!-- Read only -->
+		<div class="read-only">
+			<?php if ( ! $document->exists() && $in_process ) : ?>
+			<p>
+				<?php
+				// translators: document title
+				printf( esc_html__( 'The %s is being generated in the background. Please reload the page to see the document data.', 'woocommerce-pdf-invoices-packing-slips' ), $document->get_title() );
```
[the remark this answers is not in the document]
MohamadNateqi: This should be fixed with this PR: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392.
Can you please try using the above PR again?

includes/class-wcpdf-admin.php
```diff
@@ -75,6 +75,9 @@ public function __construct()	{
 		add_filter( 'woocommerce_rest_prepare_report_orders', array( $this, 'add_invoice_number_to_order_report' ) );
 		add_filter( 'woocommerce_report_orders_export_columns', array( $this, 'add_invoice_number_header_to_order_export' ) );
 		add_filter( 'woocommerce_report_orders_prepare_export_item', array( $this, 'add_invoice_number_value_to_order_export' ), 10, 2 );
+
+		add_filter( 'heartbeat_received', array( $this, 'handle_heartbeat_received' ), 10, 2 );
```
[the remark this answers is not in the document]
MohamadNateqi: Do you mean you had the metabox without any data?

It's weird because the [`document-data-metabox.php`](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/files#diff-0d342e77085acd3923be230ea36e354d74d4232e687d0bc4f8aafe3eecc3fe0aR807) should be loaded on the page load anyway and display something.

Could you help me replicate it?

includes/class-wcpdf-admin.php
```diff
@@ -1300,47 +1300,57 @@ public function handle_heartbeat_received( array $response, array $data ): array
...
 	}
 
+	/**
+     * Updates documents data in the "PDF document data" meta box if the generation in the background is finished.
+     *
+	 * @return void
+	 */
 	public function ajax_fetch_pdf_document_data(): void {
-		if ( isset( $_REQUEST['nonce'] ) && wp_verify_nonce( $_REQUEST['nonce'], 'generate_wpo_wcpdf' ) ) {
+		if ( ! isset( $_REQUEST['nonce'] ) || wp_verify_nonce( $_REQUEST['nonce'], 'generate_wpo_wcpdf' ) ) {
```
alexmigf, line 1309: I think should be `security` instead of `nonce` and `! wp_verify_nonce`?

includes/class-wcpdf-admin.php
```diff
@@ -1300,47 +1300,57 @@ public function handle_heartbeat_received( array $response, array $data ): array
...
 				continue;
 			}
 
-			$document = wcpdf_get_document( $document_type, wc_get_order( $_REQUEST['order_id'] ) );
+			$document = wcpdf_get_document( $document_type, wc_get_order( $order_id ) );
 			if ( $document && $document->exists() ) {
-				ob_start();
 				$this->output_number_date_edit_fields( $document, $data[ $document_type ] );
-				$documents_data[ $document_type ] = ob_get_clean();
+				$documents_data[ $document_type ] = ob_get_contents();
```
alexmigf, line 1341: I think you missed my comment here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757#discussion_r1548165448

includes/class-wcpdf-admin.php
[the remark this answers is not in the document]
alexmigf: You're right, my bad.

alexmigf (changes requested):

includes/class-wcpdf-admin.php
MohamadNateqi: Good catch. Thanks.

includes/class-wcpdf-admin.php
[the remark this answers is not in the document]
MohamadNateqi: Please let me know if this has been fixed by this commit: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/commits/acc87cb0ddd1de0aac3c2d489192ef09d5f49bdc.

includes/class-wcpdf-admin.php
```diff
@@ -790,150 +793,48 @@ public function get_current_values_for_document( $document, $data ) {
...
-					</p>
-					<?php endif; ?>
-				</div>
 
-				<!-- Document Notes -->
-				<?php if( array_key_exists( 'notes', $data ) ) : ?>
+		$independent_documents = apply_filters( 'wpo_wcpdf_document_data_meta_box_independent_documents', array( 'invoice', 'packing-slip' ) );
```
alexmigf, line 801: I think this could be problematic when you have other documents being generated on the background, which hides them from the document data while are being processing.

includes/class-wcpdf-admin.php
```diff
@@ -1386,19 +1287,95 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+	/**
+	 * Append ready documents to the heartbeat response.
+	 *
+	 * @param array $response
+	 * @param array $data
+	 *
+	 * @return array
+	 */
+	public function handle_heartbeat_received( array $response, array $data ): array {
+		if ( empty( $data['wpo_pending_documents'] ) || empty( $data['wc-refresh-order-lock'] ) ) {
```
alexmigf, line 1312: This doesn't seem to run for me because in my case I'm not getting the key `wc-refresh-order-lock` but `wp-refresh-post-lock` instead.

alexmigf: Also, maybe `wpo_wcpdf_pending_documents` would be more appropriate?

alexmigf (changes requested):
For some reason I can't delete the Receipt:

Screenshot from 2024-04-10 11-26-56

It may not seem directly relevant, but it could be worth investigating.

The time inputs width is strange:

Screenshot from 2024-04-10 11-31-28

MohamadNateqi:
@alexmigf 
> For some reason I can't delete the Receipt:

It seems the document was removed, but it didn't refresh the content.
I couldn't replicate the issue. Can you tell me the steps to take to replicate it?

> It may not seem directly relevant, but it could be worth investigating.
> 
> The time inputs width is strange:
> 
> Screenshot from 2024-04-10 11-31-28

This issue happens on Firefox. Fixed.

includes/class-wcpdf-admin.php
MohamadNateqi: Thanks for catching that. This should fix the issue: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/commits/ab89e8c5e4d56b333c7b8849b6d3c582368319e0.

includes/class-wcpdf-admin.php
```diff
@@ -1386,19 +1286,98 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+			return $response;
+		}
+
+		$order_id          = $data['wc-refresh-order-lock'] ?? $data['wp-refresh-post-lock']['post_id'];
+		$pending_documents = $data['wpo_wcpdf_pending_documents'];
+
+		foreach ( $pending_documents as $document_type ) {
+			$document = wcpdf_get_document( $document_type, wc_get_order( $order_id ) );
+			if ( $document && $document->exists() ) {
+				$response['wpo_ready_documents']['document_types'][] = $document_type;
```
alexmigf, line 1324: Maybe this can be also renamed? `wpo_wcpdf_ready_documents`

includes/class-wcpdf-admin.php
```diff
@@ -1386,19 +1286,98 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+	 */
+	public function handle_heartbeat_received( array $response, array $data ): array {
+		if (
+			empty( $data['wpo_wcpdf_pending_documents'] ) ||
+			( empty( $data['wc-refresh-order-lock'] ) && empty( $data['wp-refresh-post-lock']['post_id'] ) )
+		) {
+			return $response;
+		}
+
+		$order_id          = $data['wc-refresh-order-lock'] ?? $data['wp-refresh-post-lock']['post_id'];
```
alexmigf, line 1318: Maybe you should check if you have an ID before continue?

alexmigf:
> I couldn't replicate the issue. Can you tell me the steps to take to replicate it?

Having the free and Pro branches set for this, first generate the Invoice and Receipt, and after that delete the Invoice, if you try to delete the Receipt after you can't:

Screenshot from 2024-04-12 09-43-29

includes/class-wcpdf-admin.php
```diff
@@ -790,150 +793,47 @@ public function get_current_values_for_document( $document, $data ) {
...
+				) );
+			}
+		}
 
-				<?php endif; ?>
-				<!-- / Document Notes -->
+		$independent_documents = apply_filters( 'wpo_wcpdf_document_data_meta_box_independent_documents', array( 'invoice', 'packing-slip' ) );
+		$data                  = apply_filters( 'wpo_wcpdf_document_data_meta_box_document_data_fields', $data, $document );
+		$data                  = $this->get_current_values_for_document( $document, $data );
+		$in_process            = as_has_scheduled_action( 'wpo_wcpdf_generate_document_on_order_status', array(
```
alexmigf, line 820: This function is only available since AS 3.3.0, maybe you should check the function existence and use another as a fallback.

includes/class-wcpdf-admin.php
```diff
@@ -790,150 +793,47 @@ public function get_current_values_for_document( $document, $data ) {
...
+
+		// Prevent displaying setup for documents that are not independent, and don't exist, or are not in process,
+		// like documents that require an invoice first and don't exist already or are not in process.
+		if ( ! $in_process && ! $document->exists() && ! in_array( $document->get_type(), $independent_documents, true ) ) {
+			return;
+		}
 
-			</section>
+		// Allow preventing document output.
+		if ( apply_filters( 'wpo_wcpdf_document_data_meta_box_allow_document_output', false, $data, $in_process, $document ) ) {
```
alexmigf, line 832: I find the naming convention of the filter hook `wpo_wcpdf_document_data_meta_box_allow_document_output` a bit misleading. Given its name, one might expect that setting the filter to `true` would enable the document output process. However, in practice, if the filter returns `true`, the function exits early (`return;`), effectively preventing any document output. This behavior seems counterintuitive based on the filter's name, suggesting that allowing document output actually prevents it.

alexmigf (changes requested):

MohamadNateqi:
@alexmigf I still cannot replicate that issue with the Receipt document.
Am I doing this correctly?

47a3dd9041

includes/class-wcpdf-admin.php
alexmigf: Both are good to me, maybe the first one, but I don't have a strong opinion.

alexmigf (commented):
Screenshot from 2024-04-12 15-44-32

I believe it's imperative that the issue of duplicate third-party documents is resolved within our free plugin. Relying on third parties to address this problem on their own is not ideal; we should take the initiative to fix it ourselves.

MohamadNateqi:
> Screenshot from 2024-04-12 15-44-32
> 
> I believe it's imperative that the issue of duplicate third-party documents is resolved within our free plugin. Relying on third parties to address this problem on their own is not ideal; we should take the initiative to fix it ourselves.

Since I thought there might not be third-parties other than our plugins that adds documents data, I thought it might be okay to change it in this way.

In this case, we need to go back to the old way of displaying documents data, I mean each third-party should be responsible to display their documents data.
I'll revert changes then, unless you have any other ideas.

alexmigf:
We have Italian Addon from here: https://github.com/LDAV/woocommerce-pdf-italian-add-on

Additionally, I believe there is another one, I don't recall which one now.

Maybe we should revert for now, and concentrate in the pending jobs situation when are generated with AS. We can improve the displaying later on.

MohamadNateqi:
@alexmigf I reverted those changes.

includes/class-wcpdf-admin.php
```diff
@@ -1386,19 +1275,103 @@ public function sort_orders_by_numeric_invoice_number( $query ): void {
...
+	 */
+	public function handle_heartbeat_received( array $response, array $data ): array {
+		if (
+			empty( $data['wpo_wcpdf_pending_documents'] ) ||
+			( empty( $data['wc-refresh-order-lock'] ) && empty( $data['wp-refresh-post-lock']['post_id'] ) )
+		) {
+			return $response;
+		}
+
+		$order_id = $data['wc-refresh-order-lock'] ?? $data['wp-refresh-post-lock']['post_id'];
```
alexmigf, line 1307: Maybe this should be sanitized?

includes/class-wcpdf-admin.php
```diff
@@ -1154,6 +1040,9 @@ public function ajax_crud_document() {
 		);
 
 		try {
+			// Allow getting documents without invoice.
+			add_filter( 'wpo_wcpdf_allow_document_without_invoice', '__return_true' );
```
alexmigf, line 1044: Maybe this could be done differently? For example a filter with an array of Invoice depending document types? Also you don't offer any variable to check which document is affected by this filter, so maybe this should be after the line `1046`.

includes/class-wcpdf-admin.php
```diff
@@ -1154,6 +1040,9 @@ public function ajax_crud_document() {
...
+
 			$document = wcpdf_get_document( $document_type, wc_get_order( $order_id ) );
 
 			if( ! empty( $document ) ) {
```
alexmigf, line 1048: WPCS

includes/class-wcpdf-admin.php
```diff
@@ -710,33 +713,17 @@ public function data_input_box_content( $post_or_order_object ) {
...
-					'label' => __( 'Notes (printed in the invoice):', 'woocommerce-pdf-invoices-packing-slips' ),
-				),
+		$invoice      = wcpdf_get_document( 'invoice', $order );
+		$packing_slip = wcpdf_get_document( 'packing-slip', $order );
 
-			);
-			// output
-			$this->output_number_date_edit_fields( $invoice, $data );
+		if ( $invoice ) {
+			$this->output_number_date_edit_fields( $invoice );
```
alexmigf, line 722: Maybe this could be inside a `foreach` loop?

alexmigf (commented):
I'm having the "Packing Slip" duplicated, not sure why.

Screenshot from 2024-04-16 11-55-06

includes/class-wcpdf-admin.php
MohamadNateqi: > Maybe this could be done differently? For example a filter with an array of Invoice depending document types?

I'm sorry, I couldn't fully understand your approach. Could you please explain more?
I added this filter here so third parties, such as our Pro extension, can use it to affect the result of the `is_allow()` method. Do you mean to change [this filter](https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392/files#diff-1aaf16af4dbfbd03af9365d3d5500e1e5e75a84b1e49aa025a71eb5b38875578R1366) from boolean to an array containing depended documents, and then we make that array empty in the free plugin?

MohamadNateqi:
> I'm having the "Packing Slip" duplicated, not sure why.
> 
> Screenshot from 2024-04-16 11-55-06

Could you please check if your Pro extension is on the related branch?

alexmigf:
It's in `main`.

MohamadNateqi:
It should be on the [pdf-document-data](https://github.com/wpovernight/woocommerce-pdf-ips-pro/tree/pdf-document-data) branch, since as I mentioned [here](https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392#issuecomment-2055961786), I moved packing slip from the Pro extension to the Free plugin.

includes/class-wcpdf-admin.php
alexmigf: > Do you mean to change [this filter](https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392/files#diff-1aaf16af4dbfbd03af9365d3d5500e1e5e75a84b1e49aa025a71eb5b38875578R1366) from boolean to an array containing depended documents, and then we make that array empty in the free plugin?

Yes.

alexmigf:
Seems to working fine now regarding the Packing Slip.

includes/class-wcpdf-admin.php
MohamadNateqi: I updated that hook here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/commits/55b47a8df4af86bbad9bd30a21d01e3e1a458768, and also in the Pro branch.

> Also you don't offer any variable to check which document is affected by this filter, so maybe this should be after the line 1046.

I couldn't understand this. Could you please explain?

alexmigf: The implementation seems a bit strange to me, because the filter has to be placed in Pro when should be the contrary.

MohamadNateqi: I think I saw this reverse use of filters a few times, maybe even in WooCommerce. This is one way we can inform third parties that we need to get the document anyway.

Another implementation that I could think of is to declare a new property for invoice requirement, for example, a boolean `$invoice_required` property in the `Order_Document` class, and a method like this inside this class:

```php
public function is_invoice_required(): bool {
	return apply_filters( 'wpo_wcpdf_document_is_invoice_required', $this->invoice_required, $this );
}
```

So, this method could be used inside the `is_pro_document_allowed()` method or by third parties, and we can change it using the provided filter to display the document in the meta box.

Or we can also generalize this by changing the `$invoice_required` property to an array property `$prerequisite_documents`, and change that method to something like this:
```php
public function get_prerequisites(): array {
	return apply_filters( 'wpo_wcpdf_document_prerequisites', $this->$prerequisite_documents, $this );
}
```

What do you think?

alexmigf: WPML is one of those examples, but if we go that way we need a proper documentation about that and also custom documents.

MohamadNateqi: I agree that this reverse use of filters might be confusing and difficult for third parties to understand. So, let's go with the last method.

Among the options for name, I think both `get_prerequisite_documents()` and `get_base_documents()` are good.
Since this requirement means the document can exist but not be displayed, I thought maybe the 'prerequisite' could be a good word.
Which one do you prefer?

alexmigf: Your suggestion `get_prerequisite_documents` is good.

MohamadNateqi: Updated with the new implementation: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/commits/a6c59279b703c8ab239f0f297dfa727037e53ba8.

includes/class-wcpdf-admin.php
```diff
@@ -1038,8 +1038,8 @@ public function ajax_crud_document() {
 		);
 
 		try {
-			// Allow getting documents without invoice.
-			add_filter( 'wpo_wcpdf_require_invoice_documents', '__return_empty_array()' );
+			// Allow getting documents even if their prerequisite documents doesn't exist.
+			add_filter( 'wpo_wcpdf_prerequisite_documents', '__return_empty_array()' );
```
alexmigf, line 1042: Maybe here you can call the function?

alexmigf (changes requested):

MohamadNateqi:
@alexmigf Maybe it's better to have these lines in the free plugin:
https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/392/files#diff-1aaf16af4dbfbd03af9365d3d5500e1e5e75a84b1e49aa025a71eb5b38875578R1354-R1358

What do you think?

alexmigf:
@MohamadNateqi yes, I agree.

includes/class-wcpdf-admin.php
alexmigf: No, calling the function here or another function with the code you mentioned below that was in Pro. I don't think you need to set the array as empty because it's an empty array property already?

MohamadNateqi: > I don't think you need to set the array as empty because it's an empty array property already?

I think it might not be empty. For example, for the receipt document when "Require invoice" is checked, the prerequisite documents is not empty, and if we don't make it empty, the issue you mentioned [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757#issuecomment-2051309553) could happen after removing the invoice.

What function do you think should be called here?

alexmigf: I think you can call the `is_allowed()` and include a new `$context` arg in it? This way you can create a filter inside it to allow or not in different contexts.

MohamadNateqi: The `wcpdf_get_document()` method, which will be called in the next lines to get the document, will call the `is_allowed()` method again separately, so I'm not sure if calling the `is_allowed()` before that could help in this implementation.

Since the `is_allowed()` method checks for different conditions and here, we just needed to modify prerequisite documents, I thought maybe applying this filter could be good. But if modifying prerequisites using the filter is not preferred, then as you mentioned, we can add a new `$context` arg to the `is_allowed()` method and a new filter inside it, So we can use that filter here before the `wcpdf_get_document()` to change the result of `is_allowed()`.

What do you think?

alexmigf: I suggested previously to create a separated function for this check: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/commits/48d03715bf5e586a79098d22eb181d9045ef5c95#diff-b4b45c08877ccc6f215e534426a81b3edd82ce371cfc3fce7420e62e279b3c1eR538-R545

This way you don't need to rely on `is_allowed()` I think?

alexmigf: I think you're right.

MohamadNateqi: I created a new method, `do_prerequisites_exist(),` but this method still needs to be used inside `wcpdf_get_document()` to prevent the generation of a document if prerequisites do not exist. So, again, this approach ended up with adding a filter to change the `do_prerequisites_exist()` method value to allow displaying documents in meta box.

Please let me know what other solution you suggest.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -535,16 +535,22 @@ public function is_allowed(): bool {
 			}
 		}
 
-		// Check document prerequisites
+		return apply_filters( 'wpo_wcpdf_document_is_allowed', $allowed, $this );
+	}
+
+	public function do_prerequisites_exist(): bool {
```
alexmigf, line 541: Do we need this function? Can't we use only the `get_prerequisite_documents()`?

includes/class-wcpdf-admin.php
```diff
@@ -1039,7 +1044,7 @@ public function ajax_crud_document() {
 
 		try {
 			// Allow getting documents even if their prerequisite documents doesn't exist.
-			add_filter( 'wpo_wcpdf_prerequisite_documents', '__return_empty_array()' );
+			add_filter( 'wpo_wcpdf_document_do_prerequisites_exist', '__return_true' );
```
alexmigf, line 1047: Maybe something like this would work below line `1049`? Because we are not forcing the document generation at this stage, so it's safe to call the document first.

```php
if ( false === apply_filters( 'wpo_wcpdf_bypass_prerequisites_restriction', true, 'ajax_crud_document', $document ) && ! empty( $document->get_prerequisite_documents() ) ) {
	return;
}
```

This can be also applied to the `data_input_box_content()` inside the loop I think.

And maybe inside `is_allowed()`?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document.php
MohamadNateqi: I thought the suggestion [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757#discussion_r1568778838) was to create a new method and move checking prerequisites condition to it from the `is_allowed()` method.

Maybe I didn't understand your suggestion correctly.

alexmigf: I'm not sure if the `get_prerequisite_documents()` existed at the time of my comment, but what I wanted to say is having one function to handle all the `prerequisite`, not having multiple functions. Sorry if it was not clear.

But let me know your opinion.

MohamadNateqi: I think it would be good to keep the `get_prerequisite_documents()` method as it is, since it acts as a getter for prerequisite documents.
I agree that the `do_prerequisites_exist()` method didn't make improvements and we just ended up with another method.
Maybe it's better to revert the recent change and go back to the previous implementation?

alexmigf: Ok, let's do that.

includes/class-wcpdf-admin.php
MohamadNateqi: We can add checking prerequisites inside the `is_allowed()` method, inside the condition you mentioned, but again we need to change the value that the `wpo_wcpdf_bypass_prerequisites_restriction` filter returns, before line `1049`, since otherwise `wcpdf_get_document()` method won't return the document.
So, I think it's again adding a filter above the line `1049`.

MohamadNateqi: In this commit: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/757/commits/08cff804ddb50a6a542f62f868a7a76c5fc59680, I added again checking prerequisites to `is_allowed()` method, but made it `false` as default, since otherwise it'll allow creation of documents without prerequisites, and also used the filter you mentioned.
Please let me know what you think.

alexmigf: You could avoid having it inside `is_allowed()`, and call it when you need to check the `is_allowed()` output. I think returning an empty array seems strange to me and a bit like hacking.

MohamadNateqi: If we don't add checking prerequisites to the `is_allowed()` method, the `wcpdf_get_document()` method will return a document even if the prerequisites of it don't exist. so I thought maybe it should be inside that as default?

> I think returning an empty array seems strange to me and a bit like hacking.

You're right.
In the recent commit, I used the filter you mentioned: `wpo_wcpdf_bypass_prerequisites_restriction`. What do you think about it?

MohamadNateqi: Yes, without sending the last parameter (`$init`) to the `wcpdf_get_document()` function, it won't generate it, but it'll still return the document. For example, a customer can manipulate the URL to get a receipt document. This is the URL to get the invoice:
`.../wp-admin/admin-ajax.php?action=generate_wpo_wcpdf&document_type=invoice&order_ids=1235...`
So they can change the `document_type` parameter and get the receipt. While the receipt is not actually saved, they can get a PDF. 
If it's OK, then I can change that filter value to `true` as default and add the condition you mentioned after the line `1049`.

MohamadNateqi: I understood your point.
So we add this condition to `data_input_box_content()`, `generate_document_ajax()`, and `is_allowed()` and make the filter value `true` by default, but for `generate_document_ajax()`, we make it `false` by default.

In this case, I like to propose adding the `do_prerequisites_exist()` method again, but with small changes:

```php
public function do_prerequisites_exist( string $context = '', bool $default = true ): bool {
	foreach ( $this->get_prerequisite_documents() as $type ) {
		$document = wcpdf_get_document( $type, $this->order );

		if ( ! $document || ! $document->exists() ) {
			return apply_filters( 'wpo_wcpdf_bypass_prerequisites_restriction', $default, $context, $this );
		}
	}

	return true;
}
```

What do you think?

alexmigf: In this case we don't need the filter do we? I think you might need to check if `get_prerequisite_documents()` is not empty?

MohamadNateqi: We can still keep the filter; it will allow us to bypass the `false` value, which I think would be good.

Another issue I see (with both approaches; using the condition you mentioned or the `do_prerequisites_exist()` function) is that if a customer changes the `document_type` in the URL, as I mentioned, we can check here to prevent displaying the document:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/main/includes/class-wcpdf-main.php#L545-L547

Like this:
```php
if ( $document && $document->do_prerequisites_exist( 'generate_document_ajax', false ) ) {
```

It'll prevent the document from being displayed but not prevent the document from being generated. So, the receipt will be generated but not displayed.

MohamadNateqi: This issue will happen everywhere when the third parameter of `wcpdf_get_document()` is `true`.

One way I can think of is to use the filter provided in the `do_prerequisites_exist()` method to change the `$bypass` value and also make the `$bypass` default value `false`. It'll be still the same as the previous solution; using a filter, but maybe seems less hacking? Since we don't change the prerequisites value.

alexmigf: I think you need to check the prerequisite before forcing the document generation here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/main/includes/class-wcpdf-main.php#L545

MohamadNateqi: Since a document object is required to call the `do_prerequisites_exist()`, to check the prerequisite before forcing the document generation, I must call the `wcpdf_get_document()` again without the third parameter. I'm not sure if this is a good approach.

alexmigf: You can get the document from here: `$document = WPO_WCPDF()->documents->get_document( $document_type, $order );`

MohamadNateqi: I added that to `is_allowed()` method like this:
```php
if ( ! $this->do_prerequisites_exist( $context ) ) {
	return false;
}
```

But since it'll bypass by default, it won't affect it. If we change the bypass to `false` by default, then we need to use the filter to change it anywhere we want, for example, for metabox.

I think checking prerequisites is a requirement for the `wcpdf_get_document()` method, so maybe it's better to keep the bypass `false` as default?

Also, we can use the `WPO_WCPDF()->documents->get_document()` method here, but I think it'll make the usage of `wcpdf_get_document()` and it should be something that should be considered for using of `wcpdf_get_document()` method.

If you agree, I will revert the related change to this matter and open another PR to address this issue, which is related to issue #760.

alexmigf: > If you agree, I will revert the related change to this matter and open another PR to address this issue, which is related to issue https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/760.

Yes, it's better, we continue from there.

MohamadNateqi:
I'm closing this PR to open new PRs to address the above issues separately.

PR #762, opened 2024-04-03
Add plugin dependency

### LINKED ISSUE #761: Adding plugin dependency
WordPress recently introduced [Plugin Dependencies](https://make.wordpress.org/core/2024/02/15/merge-announcement-plugin-dependencies/) "to make the process of installing/activating add-ons (dependents) and the plugins they rely on (dependencies) consistent and easy to use."

This is also mentioned in the article that:
"The feature does not intend to replace the need for defensive coding within dependent plugins, or ensure version compatibility with their dependencies; nor does it seek to mitigate the vast array of potential plugin interactions.".

"A plugin’s dependencies can be declared by the use of a new `Requires Plugins` header in the dependent plugin’s main file. This must contain a comma-separated list of WordPress.org slugs for its dependencies, such as jetpack (jetpack/jetpack.php is not supported)."

"Defining plugins not hosted on WordPress.org as dependencies will enforce and display the dependencies to the user. However, there will be no ability to install the missing plugins through the UI changes, and these will need to be installed manually."

<hr>

Declaring a plugin dependency for the **Bridge** plugin, dependent on **Foo** and **Bar**, places the following requirements:

- **Bridge** can only be installed once **Foo** and **Bar** have been installed.
- **Bridge** can only be activated once **Foo** and **Bar** have been activated.
- **Foo** and **Bar** can only be deactivated once **Bridge** has been deactivated.
- **Foo** and **Bar** can only be deleted once **Bridge** has been deleted.

<hr>

This is how it'll appear in the plugin repository:

image

This is how it'll appear in "Installed plugins":

image

MohamadNateqi:
close #761

This PR aims to add WooCommerce as the plugin dependency.

### DIFF
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -1,6 +1,7 @@
 <?php
 /**
  * Plugin Name:          PDF Invoices & Packing Slips for WooCommerce
+ * Requires Plugins:     woocommerce
  * Plugin URI:           https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/
  * Description:          Create, print & email PDF or UBL Invoices & PDF Packing Slips for WooCommerce orders.
  * Version:              3.8.0


alexmigf (approved):
LGTM!

Terminator-Barbapapa (approved):

PR #764, opened 2024-04-08
Fix: Semaphore attachments issue on concurrent refunds

### LINKED ISSUE #763: Semaphore issue with email attachments on concurrent refunds
A customer reported an issue with emails not being attached when doing concurrent refunds due to lock being used by another process.

alexmigf:
closes #763


MohamadNateqi (approved):
LGTM!

PR #765, opened 2024-04-15
New: DOMPDF upgrade to v2.0.5

alexmigf:


MohamadNateqi (approved):
Works correctly on my end with all templates for all documents.

alexmigf:
They have released `v2.0.7` today as well, I will close this in favor of that one.

PR #769, opened 2024-04-17
Fix: bug calling `is_allowed()` on `bool`

### LINKED ISSUE #766: Check if the document type exists to avoid a fatal error
[This user](https://wordpress.org/support/topic/fatal-error-call-to-a-member-function-is_allowed-on-bool-3) has reported that the document query returns a fatal error when the document type does not exist. I tried to reproduce the scenario and confirmed that the error happens. Here is the stack trace I received:

```
Fatal error: Call to a member function is_allowed() on bool

/home/example.com/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/wcpdf-functions.php (82)

#0 /home/example.com/public_html/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/class-wcpdf-main.php(545): wcpdf_get_document()
#1 /home/example.com/public_html/wp-includes/class-wp-hook.php(324): WPO\WC\PDF_Invoices\Main->generate_document_ajax()
#2 /home/example.com/public_html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
#3 /home/example.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action()
#4 /home/example.com/public_html/wp-admin/admin-ajax.php(192): do_action()
#5 {main}
```

alexmigf:
closes #766

### DIFF
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -79,7 +79,7 @@ function wcpdf_get_document( $document_type, $order, $init = false ) {
 
 			$document = WPO_WCPDF()->documents->get_document( $document_type, $order );
 
-			if ( ! $document->is_allowed() ) {
+			if ( ! $document || ! $document->is_allowed() ) {
 				return apply_filters( 'wcpdf_get_document', false, $document_type, $order, $init );
 			}
 


MohamadNateqi (approved):
LGTM!

Terminator-Barbapapa:
[IPS Free (bug): Check if the document type exists to avoid a fatal error](https://trello.com/c/i0AAynbj/171-ips-free-bug-check-if-the-document-type-exists-to-avoid-a-fatal-error)

PR #775, opened 2024-04-22
Fix: document date parse function being called on non HPOS stores

alexmigf:
This PR addresses an issue spotted on the Professional extension, which lead to issues in the Bulk Export when using the Document date.

The `meta_query` key used by `wc_get_orders`, and built by our function `wpo_wcpdf_parse_document_date_for_wp_query`, is only available on HPOS.


MohamadNateqi (approved):
LGTM!

Terminator-Barbapapa (approved):

PR #776, opened 2024-04-23
Apply footer CSS to the wrapper

### LINKED ISSUE #772: Apply footer CSS to the wrapper
As we've now introduced a new wrapper to our footer to be compatible with mPDF, we should apply the footer styling to the `htmlpagefooter` instead of onto `#footer`.

We should then also update the 'Repeating headers and footers' documentation article.

BrunoPavlinic98:
closes #772


MohamadNateqi (approved):
LGTM!

PR #783, opened 2024-04-30
New: PDFJS v3.11.174 upgrade

alexmigf:
PDFJS version 4 uses JavaScrfipt Modules, we might need to change our implementation a bit, we should do that in the near future: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/784


MohamadNateqi:
Getting this:

image

alexmigf:
I have updated both files, maybe it's cache? I'm not having that error.

MohamadNateqi (approved):
LGTM!

Sorry! Apparently, my browser refuses to clear the cache completely, even though I ask for it! I tried using another browser, and it is working well.

MohamadNateqi (approved):
LGTM!

PR #787, opened 2024-05-06
Fix: reverts the `customer_invoice` email title to the original Woo title

alexmigf:
Beside the slug include the key `customer_invoice`, this email just sends the order details. I think we should revert this.


MohamadNateqi:
May I ask why this has been reverted?

alexmigf:
@MohamadNateqi I forgot to add the description, done now.

MohamadNateqi (approved):
LGTM!

PR #795, opened 2024-05-16
Update language packs and translation template (up to 3.8.2)

### LINKED ISSUE #794: Update bundled language packs (.po files)
A user reported that a string was missing when they tried to translate it with Poedit, and I realized that they were using a .po files from our bundled language packs, which was outdated.

See: https://wordpress.org/support/topic/the-due-date-cannot-be-translated/

YordanSoares:
Closes: #794


alexmigf:
@YordanSoares Is this still relevant, or can we close it?

YordanSoares:
@alexmigf I've just checked, and all the included translation projects are very out of date. There are currently 1108 new strings missing from all of them:

image

I went ahead and updated all the projects locally, but this raised a new concern: the size of the `languages` folder has increased. To keep the file size down, I have purged unused translations, but it is still around 2 MB larger.

Here is the folder size before and after updating the translation projects:

* Before: 1,42 MB (1.497.473 bytes)
* After: 3,47 MB (3.649.443 bytes)

Unfortunately, at the moment, the completion percentage of the bundled translations is not high enough to move them to translate.wordpress.org, as the WP.org translation platform requires translations to be at least 90% complete (see number 5 [here](https://make.wordpress.org/meta/handbook/tutorials-guides/translations/)).

At this point, how would you recommend we proceed?

1. Close this issue for now and look into a workaround later.
2. Have me push these changes and provide the updated translation projects.
3. Something else?

alexmigf:
@YordanSoares Keep the current bundled translations for compatibility, but start migrating them to translate.wordpress.org. Don't add 2 MB of mostly untranslated PO data to the plugin just to keep the local projects synchronized. Prioritize getting our main locales onto WordPress.org language packs, and remove bundled translations progressively as each locale becomes available there.

YordanSoares:
@alexmigf it makes sense. Let's close this issue then 👍

PR #796, opened 2024-05-16
Add  `<p>` tag to allowed tag of the `wpo_wcpdf_sanitize_html_content()` method

MohamadNateqi:
The `<p>` tag is necessary to display order notes correctly, and it's safe to add this to the default allowed tags.


alexmigf (approved):
LGTM!

PR #799, opened 2024-05-17
Fix: the function `get_extensions_license_data` returning empty if no cached data found

### LINKED ISSUE #798: The function `get_extensions_license_data`returns empty if no cached data set
After some debugging I found that the function `get_extensions_license_data` returns empty if no cached data found. This should fix the Pro license issues found with the Cloud Storage feature.

alexmigf:
closes #798

### DIFF
--- a/includes/settings/class-wcpdf-settings-upgrade.php
+++ b/includes/settings/class-wcpdf-settings-upgrade.php
@@ -283,10 +283,10 @@ class Settings_Upgrade {
 	/**
 	 * Get extensions license data
 	 *
-	 * @param  string $type can be 'cached' or 'live'
+	 * @param string $type can be 'cached' or 'live'
 	 * @return array
 	 */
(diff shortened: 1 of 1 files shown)


includes/settings/class-wcpdf-settings-upgrade.php
```diff
@@ -283,20 +283,25 @@ public function clear_extensions_license_cache() {
 	/**
 	 * Get extensions license data
 	 *
-	 * @param  string $type    can be 'cached' or 'live'
+	 * @param  string $type can be 'cached' or 'live'
 	 * @return array
 	 */
 	public function get_extensions_license_data( $type = 'cached' ) {
```
MohamadNateqi, line 289: Could you also add the return type declaration?

MohamadNateqi (approved):
LGTM!

PR #801, opened 2024-05-21
New: adds semaphore to `wcpdf_get_document` when forcing init

alexmigf:


MohamadNateqi:
I again could create two records in DB:
image

Removing the `release()` prevented creating two records. Maybe you can remove that?

alexmigf:
I think not. Could you try again by increasing this? https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/801/files#diff-6853d85870a98f92b3e34cb73ec089472f35af0c81b73cfbdaf63ac2d876c0b2R74

MohamadNateqi:
Chaged it to `1000`, happend again.

MohamadNateqi:
Would increasing that number help if we use `release()`? Since I think the release removes the expiration time?

MohamadNateqi:
Logs:

image

image

> The release() method is used to "manually" release the lock as soon as possible, when the process ends before the lock expiration. If we don't use it, the lock only releases on expiration.

I guess the issue occurs when I click the update button multiple times. It seems that several threads are running simultaneously with a slight time difference, causing the first thread to release the lock and the second thread to acquire it, resulting in two records being created.

alexmigf:
Thank you, I will check that.

alexmigf:
@MohamadNateqi could you try my last commit?

MohamadNateqi:
DB:

image

Log file: [wpo-wcpdf-semaphore-2024-05-21.log](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/files/15390834/wpo-wcpdf-semaphore-2024-05-21.log)

I used to remove logs on each try, so the whole log is related to one attempt that created those records.

MohamadNateqi:
The issue could be reproduced in the below versions:
v3.8.0
v3.7.7
v3.7.5
v3.7.3
v3.7.0
v3.6.3

alexmigf:
The underlying issue of concurrent numbers has been persistent for a long time. I recall dealing with this issue even during the Ewout era, which led to David suggesting the use of his Semaphore class. We've been using this approach for some time now.

However, the issue continues to affect our plugin, as confirmed by @MohamadNateqi's testing on various versions, including `v3.6.3`.

We spent considerable time debugging this issue and exploring different solutions, such as MySQL LOCK, WP cache, Order object cache, transients, and several other methods to understand the problem and find the best approach.

### Concurrent Request Handling:

The primary challenge is that requests are made at the same second, leading to simultaneous attempts to create the document. Even with a forced delay before acquiring the lock, the order data isn't available before the `init()` and `save()` methods of the document are called. This prevents us from checking if the number was already assigned in concurrent requests.

### Race Condition Handling:

Introduced random delays to reduce race conditions if certain actions are detected.
Added filters to allow customization of lock times, race condition delays, and actions, making the function more flexible and adaptable.

### Document and Order Verification:

Added comprehensive checks to ensure the document is not created multiple times, including pre-lock checks, and direct database queries for number availability.

### Lock Expiration Strategy:

We discovered that allowing the lock to expire on its own helps prevent the issue. The downside is the uncertainty of how long the request will take, potentially causing minor execution delays. To fine-tune this, we check the current action and exclude sensitive documents, ensuring the delay only occurs during race conditions.

### Lock Management:

Improved lock handling by scheduling and unscheduling lock release actions using Action Scheduler.
Ensured that the lock release is handled properly, even in case of process interruptions.

### Scheduled Lock Release:

We introduced a mechanism to schedule an action for releasing the lock in case the process is abruptly stopped. This prevents the lock from getting stuck in the database and ensures it is released appropriately.

### Database Query for Number Availability:

We implemented a direct database query to check if a number is already assigned to the document and order. Since the number becomes available immediately after document initialization, this check effectively prevents new locks from being acquired unnecessarily.

### Extensibility:

Included filters to allow customization, ensuring the function can be extended to other situations where similar issues may arise.

## Conclusion:

The final function is robust and should effectively handle concurrent document generation requests. The combination of random delays, direct database checks, and scheduled lock release actions ensures that the function mitigates race conditions and prevents duplicate document creation.

MohamadNateqi:
I think there is still an issue.
I changed the status to completed, the document has been created, then removed the document, changed it to some other status, and then changed it back to completed. The document didn't generated this time, which I think is because it the lock was not released.

```php
2024-05-28T07:59:41+00:00 Info Lock (wpo_wcpdf_lock/wcpdf_init_document/invoice_with_order_1325, wp_options) acquired
2024-05-28T07:59:41+00:00 Info [17243952] Lock acquired to init document invoice with order ID# 1325.
2024-05-28T07:59:41+00:00 Info [17243952] Document init completed for invoice with order ID# 1325.
2024-05-28T07:59:41+00:00 Info [17243952] Document save completed for invoice with order ID# 1325.
2024-05-28T07:59:41+00:00 Info [17243952] Waiting lock to be released...
2024-05-28T08:00:17+00:00 Debug Lock option (wpo_wcpdf_lock/wcpdf_init_document/invoice_with_order_1325, wp_options) already existed in the database
2024-05-28T08:00:17+00:00 Info Lock (wpo_wcpdf_lock/wcpdf_init_document/invoice_with_order_1325, wp_options) could not be acquired (it is locked)
2024-05-28T08:00:17+00:00 Critical [55339775] Couldn't get the lock while initiating the document invoice with order ID# 1325.
```

PR #804, opened 2024-05-27
HPOS: add Invoice number to search type selector #748

### LINKED ISSUE #748: HPOS: add Invoice number to search type selector
Following [this PR](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/747) review, I notice that HPOS has a new selector in the order list search, maybe we could add the Invoice number option there:

Screenshot from 2024-03-13 16-15-31

BrunoPavlinic98:
closes #748


includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -317,6 +320,37 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
 	}
+
+	/**
+	 * Adds invoice number filter to the search filters available in the admin order search.
+	 * 
+	 * @param array $options List of available filters.
+	 * 
+	 * @return array
+	 */
+	function wcpdf_hpos_admin_search_filters ( array $options ): array {
```
MohamadNateqi, line 331: WPCS - No space after the function name.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -317,6 +320,37 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+	}
+	
+	/**
+	 * Modifies the arguments passed to `wc_get_orders()` to support 'invoice_number' order search filter.
+	 * 
+	 * @param array $order_query_args Arguments to be passed to `wc_get_orders()`.
+	 * 
+	 * @return array
+	 */
+	function wcpdf_invoice_number_query_args ( array $order_query_args ): array {
```
MohamadNateqi, line 344: WPCS - No space after the function name.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
MohamadNateqi, line 331: I think you can omit the `wcpdf_` prefix.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
MohamadNateqi, line 344: I think you can omit the `wcpdf_` prefix.

MohamadNateqi (changes requested):

MohamadNateqi (approved):
Working well on my test.

dwalkerpriv (approved):
All good on my end 👍

YordanSoares (approved):
LGTM @BrunoPavlinic98!

I just want to suggest a couple of small changes:

1. Moving the "Invoice number" option just before the "All" option, which I think makes more sense:
   
   image
2. To mention this new option in the order search selector in the "Enable invoice number search" setting, so the customers are aware that this setting can be turned off, since the global search is no longer required to search for the invoice number:
   
   image

dpeyou (approved):

BrunoPavlinic98:
Following @YordanSoares's comment, I have added a message `Enable invoice number search in the orders list` option description, and placed the invoice number select option above the "All" option.

@alexmigf @MohamadNateqi @dpeyou @dwalkerpriv do you agree with this message?

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -317,6 +320,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+	/**
+	 * Adds invoice number filter to the search filters available in the admin order search.
+	 * 
+	 * @param array $options List of available filters.
+	 * 
+	 * @return array
+	 */
+	function hpos_admin_search_filters( array $options ): array {
+		$all_element_position = array_search( 'all', array_keys( $options ) );
+		$all_element = array_slice( $options, $all_element_position, $all_element_position );
```
MohamadNateqi, line 333: It might be slightly faster if you preserve 'all' in a variable and then unset it. What do you think?

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -317,6 +320,41 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+		$options = array_slice( $options, 0, $all_element_position );
+		$options['invoice_number'] = __( 'Invoice number', 'woocommerce-pdf-invoices-packing-slips' );
```
MohamadNateqi, line 335: WPCS - Alignment.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -457,7 +457,7 @@ public function get_pdf_settings_fields( $option_name ) {
 				'args'			=> array(
 					'option_name'	=> $option_name,
 					'id'			=> 'invoice_number_search',
-					'description'   => __( 'Can potentially slow down the search process.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description'   => __( 'Can potentially slow down the search process. You can also search orders by invoice numbers using search type selector, which is faster than this option.', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi, line 460: Maybe you should mention something about the HPOS since the customer might not be aware that this is a feature of HPOS.

MohamadNateqi (changes requested):

includes/documents/class-wcpdf-invoice.php
BrunoPavlinic98: @MohamadNateqi I agree, I have changed the code!

includes/documents/class-wcpdf-invoice.php
```diff
@@ -457,7 +457,7 @@ public function get_pdf_settings_fields( $option_name ) {
 				'args'			=> array(
 					'option_name'	=> $option_name,
 					'id'			=> 'invoice_number_search',
-					'description'   => __( 'Can potentially slow down the search process.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description'   => __( 'Can potentially slow down the search process. You can also use the HPOS(High-Performance Order Storage) feature to search orders by invoice numbers using search type selector, which is faster than this option.', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi, line 460: Do you think it would also be good to add a link to this article? Maybe making the HPOS a hyperlink?
https://woocommerce.com/document/high-performance-order-storage/

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -317,6 +320,39 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+	 * Adds invoice number filter to the search filters available in the admin order search.
+	 * 
+	 * @param array $options List of available filters.
+	 * 
+	 * @return array
+	 */
+	function hpos_admin_search_filters( array $options ): array {
+		unset( $options['all'] );
+		$options['invoice_number'] = __( 'Invoice number', 'woocommerce-pdf-invoices-packing-slips' );
+		$options['all']            = __( 'All', 'woocommerce-pdf-invoices-packing-slips' );
```
MohamadNateqi, line 334: Maybe it would be better if you stored the "all" option and then brought it back here. In this case, if any changes happen to the "All" text (which is unlikely!), it'll not override that. 
I mean something like this:
```php
$all = $options['all'];
unset( $options['all'] );
$options['invoice_number'] = __( 'Invoice number', 'woocommerce-pdf-invoices-packing-slips' );
$options['all'] = $all;

```

Also, you used our text domain, which is different from the Woo.

What do you think?

BrunoPavlinic98: @MohamadNateqi I agree, with this approach we remove the possibility of some changes made by WooCoommerce that are not followed by our code.

includes/documents/class-wcpdf-invoice.php
BrunoPavlinic98: I agree, great suggestion.

MohamadNateqi (approved):
LGTM!

includes/documents/class-wcpdf-invoice.php
```diff
@@ -441,7 +441,7 @@ public function get_pdf_settings_fields( $option_name ) {
 				'args'			=> array(
 					'option_name'	=> $option_name,
 					'id'			=> 'invoice_number_search',
-					'description'   => __( 'Can potentially slow down the search process.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description'   => __( 'Can potentially slow down the search process. You can also use the <a href="https://woocommerce.com/document/high-performance-order-storage/" target="_blank">HPOS</a> feature to search orders by invoice numbers using search type selector, which is faster than this option.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 444: I think this setting can be disabled and omitted when running HPOS.

alexmigf: Actually, you could rely on it to enable the search in HPOS too?

BrunoPavlinic98: @alexmigf I don't understand this completely. You think that I should only show Invoice Number in HPOS selector in case this option is enabled?

alexmigf: Yes

BrunoPavlinic98: @alexmigf I agree. I have also modified this option's description so that it is not confusing for users.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -441,7 +441,7 @@ public function get_pdf_settings_fields( $option_name ) {
 				'args'			=> array(
 					'option_name'	=> $option_name,
 					'id'			=> 'invoice_number_search',
-					'description'   => __( 'Can potentially slow down the search process.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'description'   => __( 'Can potentially slow down the search process. You can also use the <a href="https://woocommerce.com/document/high-performance-order-storage/" target="_blank">HPOS</a> feature to search orders by invoice numbers using search type selector.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 444: > Can potentially slow down the search process.

I think this only applies to non-HPOS stores.

includes/compatibility/class-wcpdf-compatibility-third-party-plugins.php
```diff
@@ -356,6 +359,42 @@ function restore_wgm_thumbnails( $document_type, $document ) {
...
+
+	/**
+	 * Adds invoice number filter to the search filters available in the admin order search.
+	 * 
+	 * @param array $options List of available filters.
+	 * 
+	 * @return array
+	 */
+	function hpos_admin_search_filters( array $options ): array {
+		if ( isset( get_option( 'wpo_wcpdf_documents_settings_invoice', array() )['invoice_number_search'] ) ) {
```
alexmigf, line 371: There is a function you can use: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/3a63852e45c92fa2329251e0a5fb1d6bd0797c94/includes/class-wcpdf-admin.php#L402

alexmigf (changes requested):

includes/documents/class-wcpdf-invoice.php
alexmigf: Maybe one that could combine the 2 situations? Suggestion:

> The search process may be slower on non-HPOS stores. For a more efficient search, you can utilize the <a href="https://woocommerce.com/document/high-performance-order-storage/" target="_blank">HPOS</a> feature, allowing you to search orders by invoice numbers using the search type selector.

BrunoPavlinic98: @alexmigf great suggestion I agree with it!

alexmigf (approved):
Perfect!

PR #812, opened 2024-06-04
New: adds language to PDF HTML

alexmigf:
This PR adds the HTML language attribute to the PDF, which can help debug language issues in multilingual setups.

Screenshot from 2024-06-04 09-41-14

### DIFF
--- a/templates/Simple/html-document-wrapper.php
+++ b/templates/Simple/html-document-wrapper.php
@@ -1,6 +1,6 @@
 <?php if ( ! defined( 'ABSPATH' ) ) exit; // Exit if accessed directly ?>
 <!DOCTYPE html>
-<html <?php language_attributes(); ?>>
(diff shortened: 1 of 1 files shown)


MohamadNateqi (approved):
LGTM!

I think it always gets the language that has been set in the WordPress settings. One of our translation options is based on order/customer language; in this case, the customer language has no effect on this value. Right?
I was wondering if we should also consider updating this value using the filter provided before HTML generation?

MohamadNateqi (approved):
LGTM!

alexmigf:
> I think it always gets the language that has been set in the WordPress settings. One of our translation options is based on order/customer language; in this case, the customer language has no effect on this value. Right?
I was wondering if we should also consider updating this value using the filter provided before HTML generation?

Using WPML and Polylang this language will reflect the option set under the Pro tab because we switch the language before loading the HTML, for the other plugins we might need to use the filter to change it.

MohamadNateqi (approved):
LGTM!

PR #814, opened 2024-06-04
Wrap address around P tag

### LINKED ISSUE #813: Wrapping address around HTML tag
Following this [PR](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/792#issuecomment-2133263891), the address displayed in PDF needs to be around an HTML tag.

Before that PR, we had a `<p>` tag around it, but now it's a plain text that creates a white space before and after address since it has been printed in a new line:

image

MohamadNateqi:
closes #813


alexmigf (approved):
LGTM!

PR #824, opened 2024-06-19
Fix `normalize_settings_args` bug

### LINKED ISSUE #823: The `normalize_settings_args()` not returning array settings
A new condition introduced [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/754/files#diff-886edad7f2523187c7336a156f576b801fc1f8f1a9ed152061400b40d2300558R657-R658) to fix displaying the logo. The issue happened when you had already set a logo, but after activating a multilingual plugin, you didn't see the logo.

That fixed the logo issue but introduced another bug. For text input settings, that condition sets the current value of the field to an array (which is an array of values for each language, like `array( 'en' => 'something', 'nl' => 'something else')`, and the text input needs a string as the current value.

The second bug was fixed [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/816) by preventing arrays for the current value, but another [bug](https://github.com/wpovernight/woocommerce-pdf-ips-pro/issues/434) appeared. Some of the settings are arrays by default, such as the "Attach to" setting, and preventing arrays caused an issues displaying the current value of those settings.

MohamadNateqi:
close #823

### DIFF
--- a/includes/settings/class-wcpdf-settings-callbacks.php
+++ b/includes/settings/class-wcpdf-settings-callbacks.php
@@ -636,7 +636,7 @@ class Settings_Callbacks {
 						$args['current'] = $option[ $args['id'] ][ $args['lang'] ];
 					} elseif ( isset( $option[ $args['id'] ]['default'] ) ) {
 						$args['current'] = $option[ $args['id'] ]['default'];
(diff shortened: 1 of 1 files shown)


alexmigf (approved):
LGTM!

PR #825, opened 2024-06-19
New: includes ImageMagick in the Advanced Status

alexmigf:


includes/views/advanced-status.php
```diff
@@ -95,15 +101,15 @@
 		)
 	);
 }
-if ( ( $gm = extension_loaded( 'gmagick' ) ) || ( $im = extension_loaded( 'imagick' ) ) ) {
-	$server_configs['GMagick or IMagick']['value'] = ( $im ? 'IMagick '.phpversion( 'imagick' ) : 'GMagick '.phpversion( 'gmagick' ) );
+
+if ( extension_loaded( 'gmagick' ) || ( $im = extension_loaded( 'imagick' ) ) ) {
+	$server_configs['GMagick or IMagick']['value'] = ( $im ? 'IMagick ' . phpversion( 'imagick' ) : 'GMagick ' . phpversion( 'gmagick' ) );
```
MohamadNateqi, line 106: I think if the `extension_loaded( 'gmagick' )` returns `true` in the condition, then `$im` will be undefined.

MohamadNateqi (approved):
LGTM!

MohamadNateqi (approved):
Looks better :)

includes/views/advanced-status.php
```diff
@@ -55,10 +57,16 @@
 	),
 	'GMagick or IMagick' => array(
 		'required' => __( 'Better with transparent PNG images', 'woocommerce-pdf-invoices-packing-slips' ),
-		'value'    => null,
-		'result'   => extension_loaded( 'gmagick' ) || extension_loaded( 'imagick' ),
+		'value'    => $imagick ? 'IMagick ' . phpversion( 'imagick' ) : ( $gmagick ? 'GMagick ' . phpversion( 'gmagick' ) : '' ),
```
Terminator-Barbapapa, line 60: It seems in other cases we default to `null` but here to an empty string. Is that on purpose?

includes/views/advanced-status.php
```diff
@@ -55,10 +57,16 @@
...
+		'result'   => $gmagick || $imagick,
 		'fallback' => __( 'Recommended for better performances', 'woocommerce-pdf-invoices-packing-slips' ),
 	),
+	'ImageMagick' => array(
+		'required' => __( 'Required for IMagick', 'woocommerce-pdf-invoices-packing-slips' ),
+		'value'    => ( $imagick && class_exists( '\\Imagick' ) ) ? esc_attr( \Imagick::getVersion()['versionString'] ) : null,
+		'result'   => $imagick,
+		'fallback' => __( 'ImageMagick library, integrated via the IMagick PHP extension for advanced image processing capabilities', 'woocommerce-pdf-invoices-packing-slips' ),
+	),
```
Terminator-Barbapapa, line 69: I think this should only be visible when the Imagick extension is loaded.

Terminator-Barbapapa (changes requested):

includes/views/advanced-status.php
alexmigf: I'm not seeing that. Are you in the latest changes of the branch?

includes/views/advanced-status.php
alexmigf: Same.

includes/views/advanced-status.php
```diff
@@ -91,7 +91,23 @@
...
+if ( $imagick ) {
+	$gmagick_imagick_position = array_search( 'GMagick or IMagick', array_keys( $server_configs ) ) + 1;
+	$image_magick_config      = array(
+		'ImageMagick' => array(
+			'required' => __( 'Required for IMagick', 'woocommerce-pdf-invoices-packing-slips' ),
+			'value'    => ( $imagick && class_exists( '\\Imagick' ) ) ? esc_attr( \Imagick::getVersion()['versionString'] ) : null,
+			'result'   => $imagick,
+			'fallback' => __( 'ImageMagick library, integrated via the IMagick PHP extension for advanced image processing capabilities', 'woocommerce-pdf-invoices-packing-slips' ),
+		),
+	);
```
Terminator-Barbapapa, line 105: I think you forgot to remove this part from the `$server_configs` array. Now the ImagaMagick line will always show up, also if `$imagick` returns `false`.

Terminator-Barbapapa (changes requested):

includes/views/advanced-status.php
alexmigf: Sorry, fixed.

Terminator-Barbapapa (approved):
🚀

PR #831, opened 2024-07-08
Make totals labels translatable

### LINKED ISSUE #830: "VAT" label in totals section is not translatable
The "VAT" label displayed in the total sections cannot be translated. It has been tested with WPML, and translation was added to the translation strings.

image

MohamadNateqi:
close #830 

By the fix provided in this PR, the VAT label has been translated:

image

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document-methods.php
+++ b/includes/documents/abstract-wcpdf-order-document-methods.php
@@ -916,7 +916,8 @@ abstract class Order_Document_Methods extends Order_Document {
 		foreach ( $totals as $key => $total ) {
 			$label = $total['label'];
 			$colon = strrpos( $label, ':' );
-			if ( $colon !== false ) {
+
+			if ( false !== $colon ) {
 				$label = substr_replace( $label, '', $colon, 1 );
 			}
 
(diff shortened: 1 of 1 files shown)


includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -916,10 +916,10 @@ public function get_woocommerce_totals() {
 		foreach ( $totals as $key => $total ) {
 			$label = $total['label'];
 			$colon = strrpos( $label, ':' );
-			if( $colon !== false ) {
+			if ( $colon !== false ) {
 				$label = substr_replace( $label, '', $colon, 1 );
 			}
-			$totals[$key]['label'] = $label;
+			$totals[ $key ]['label'] = __( $label, 'woocommerce-pdf-invoices-packing-slips' );
```
alexmigf, line 922: You need to follow this approach (textdomain included): https://github.com/wpovernight/woocommerce-pdf-ips-templates/commit/4af49ec3491f23330888a1a7e8f6c31bb1e1d1b2#diff-e54d4a86c56437f82e078cf79de1abf3c9e5a4c18bb2a76170c08f09ec563d24R356-R365

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document-methods.php
MohamadNateqi: Thanks for the suggestion. Updated here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/831/commits/8be4bf92c5f856989ecc4fab41520e6729f6875e

alexmigf (approved):
LGTM!

PR #832, opened 2024-07-11
Tweak: improved Bulk document `get_html` method

alexmigf:
This PR aims to improve the Bulk document `get_html` method by placing the required hooks inside the orders loop. This is important because we detect and reset languages, in multilingual setups, using this hooks in each document here: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/397/files#diff-f5fe2ecddcf1afabc5e611c28a555e0d99f5f1745d4f2dac8af70dda17dadf86R27-R29

### DIFF
--- a/includes/documents/class-wcpdf-bulk-document.php
+++ b/includes/documents/class-wcpdf-bulk-document.php
@@ -117,31 +117,36 @@ class Bulk_Document {
 	}
 
 	public function get_html() {
-		// temporarily apply filters that need to be removed again after the html is generated
-		$html_filters = apply_filters( 'wpo_wcpdf_html_filters', array(), $this );
-		$this->add_filters( $html_filters );
-
-		do_action( 'wpo_wcpdf_before_html', $this->get_type(), $this );
+		do_action( 'wpo_wcpdf_before_bulk_html', $this );
 
 		$html_content = array();
 		foreach ( $this->order_ids as $key => $order_id ) {
 			do_action( 'wpo_wcpdf_process_template_order', $this->get_type(), $order_id );
 
-			$order = wc_get_order( $order_id );
+			$order    = wc_get_order( $order_id );
+			$document = wcpdf_get_document( $this->get_type(), $order, true );
 
-			if ( $document = wcpdf_get_document( $this->get_type(), $order, true ) ) {
+			// temporarily apply filters that need to be removed again after the html is generated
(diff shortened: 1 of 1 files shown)


MohamadNateqi (approved):
LGTM!

MohamadNateqi:
May I ask why you removed these codes:
```php
// temporarily apply filters that need to be removed again after the html is generated
$html_filters = apply_filters( 'wpo_wcpdf_html_filters', array(), $document );
$this->add_filters( $html_filters );

do_action( 'wpo_wcpdf_before_html', $document->get_type(), $document );
```

alexmigf:
@MohamadNateqi because `$document->get_html()` already call them.

MohamadNateqi (approved):
Ah yes, smart!
LGTM!

MohamadNateqi (approved):
LGTM!

includes/documents/class-wcpdf-bulk-document.php
```diff
@@ -117,31 +117,31 @@ public function get_pdf() {
 	}
 
 	public function get_html() {
-		// temporarily apply filters that need to be removed again after the html is generated
-		$html_filters = apply_filters( 'wpo_wcpdf_html_filters', array(), $this );
-		$this->add_filters( $html_filters );
-
-		do_action( 'wpo_wcpdf_before_html', $this->get_type(), $this );
```
Terminator-Barbapapa, line 124: I don't think we can simply remove this here because it is also called in `get_html()`, because the filter here let's you target the whole bulk document, and the filter in `get_html()` lets you target the single document. The `$this` variable returns something different for this filter than the filter within `get_html()`.

includes/documents/class-wcpdf-bulk-document.php
```diff
@@ -117,31 +117,31 @@ public function get_pdf() {
...
 
 		// get wrapper document & insert body content
 		$this->wrapper_document = wcpdf_get_document( $this->get_type(), null );
-		$html = $this->wrapper_document->wrap_html_content( $this->merge_documents( $html_content ) );
+		$html                   = $this->wrapper_document->wrap_html_content( $this->merge_documents( $html_content ) );
 
-		do_action( 'wpo_wcpdf_after_html', $this->get_type(), $this );
-
-		// remove temporary filters
-		$this->remove_filters( $html_filters );
```
Terminator-Barbapapa, line 144: Same here.

Terminator-Barbapapa (changes requested):

PR #833, opened 2024-07-12
Fix preview issue

### LINKED ISSUE #829: PDF preview is not reflecting the selected option in "Display number" setting
There are "Display invoice number" and "Display packing slip number" settings. One of the options of these settings is "Order number", but when selected, the "document number" is displayed in the preview instead of the "order number".

If a document is generated for that order, when the selected option is "Order number", the document displays the order number correctly. 
Also, the preview always displays the number that is displayed in the generated document.

MohamadNateqi:
close #829 

This PR aims to fix three issues on preview:

1. Number format: The document setting for the preview process was `null`, which is fixed, and now number formats are applied.
2. Date: The preview was not updated based on the `display_date` setting, which is fixed by introducing a new `initiate_date` and calling it during the preview process.
3. Number: The preview was not updated based on the `display_number` setting, which is fixed by introducing a new `generate_document_number` and calling it during the preview process.


includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -314,6 +308,46 @@ public function initiate_number( $force_new_number = false ) {
...
 	}
 
+	/**
+	 * Generate the document number.
+	 *
+	 * @param bool $increment_document_number
+	 *
+	 * @return mixed
+	 */
+	public function generate_document_number( bool $increment_document_number = false ) {
```
alexmigf, line 318: Maybe just `$increment` is enough?

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -262,7 +262,24 @@ public function save_settings( $latest = false ) {
 		}
 	}
 
-	public function initiate_number( $force_new_number = false ) {
+	public function initiate_date(): void {
```
alexmigf, line 265: Maybe you could add DocBlocks?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1282,36 +1282,25 @@ public function document_notes(): void {
...
 		if ( empty( $document_display_date ) ) {
-			$document_settings = $this->settings;
-			if( isset( $document_settings['display_date'] ) ) {
-				$document_display_date = $document_settings['display_date'];
-			}
-			else {
-				$document_display_date = 'invoice_date';
-			}
+			$document_settings     = $this->settings;
+			$document_display_date = $document_settings['display_date'] ?? 'document_date';
```
alexmigf, line 1291: Old Invoices still use `invoice_date`, you need to accommodate that too.

includes/documents/abstract-wcpdf-order-document-methods.php
```diff
@@ -1282,36 +1282,25 @@ public function document_notes(): void {
...
+		return $this->get_display_date_label( $document_display_date );
 	}
 
-	public function get_display_date_label( $date_string ) {
-
+	public function get_display_date_label( string $date_string ): string {
 		$date_labels = array(
-			'invoice_date'	=> __( 'Invoice Date' , 'woocommerce-pdf-invoices-packing-slips' ),
-			'order_date'	=> __( 'Order Date' , 'woocommerce-pdf-invoices-packing-slips' ),
+			'document_date' => __( 'Document Date', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 1299: I think having "Invoice Date" is preferable, maybe using the document object title?

alexmigf (changes requested):

alexmigf (approved):
LGTM!

alexmigf:
@MohamadNateqi the Premium Templates needs to be updated on the `document_date` change: https://github.com/search?q=repo%3Awpovernight%2Fwoocommerce-pdf-ips-templates+invoice_date&type=code

alexmigf:

MohamadNateqi:
@alexmigf Done.

alexmigf:
@MohamadNateqi can you fix the conflicts again please?

MohamadNateqi:
@alexmigf Done.

alexmigf:
@MohamadNateqi could you fix the conflicts here please?

includes/Documents/OrderDocument.php
```diff
@@ -279,6 +287,46 @@ public function initiate_number( $force_new_number = false ) {
...
 	}
 
+	/**
+	 * Generate the document number.
+	 *
+	 * @param bool $increment
+	 *
+	 * @return mixed
+	 */
+	public function generate_document_number( bool $increment = false ) {
```
alexmigf, line 297: Maybe `get_` would be more appropriated?

alexmigf (changes requested):

includes/Documents/OrderDocument.php
alexmigf: I could get on board with that, but in that case, we should remove the `$increment` parameter and the `get_next()` part, since it seems like two different concepts are being mixed here.

MohamadNateqi: Understand, you're right.

In this case, I think we have the following options:
1. Rename it to something like this: `get_or_generate_document_number`.
2. Using `get_` as you said.

Both options look good to me. Which do you prefer?

alexmigf: Why not 2 separated functions? Just an idea, but let me know.

MohamadNateqi: I may not fully share your vision, but two functions may create more hassle and be less clean than one.

Abstracting away the conditional boilerplate from the caller and having a single method that encapsulates both might be better.

Also, it might be simpler for third parties to interrupt the process of obtaining document numbers and modify the data since we're only incrementing the document number if third parties are not assigning a number already.

Although the one-function approach is simple, I agree that it violates SRP and does two jobs in one function.

alexmigf: We just an idea that came to my mind while writing my reply. Just `get_` makes more sense to me, because even if you are incrementing you are getting a new number.

includes/Documents/OrderDocument.php
```diff
@@ -294,7 +294,7 @@ public function initiate_number( bool $force_new_number = false ) {
 	 *
 	 * @return mixed
 	 */
-	public function generate_document_number( bool $increment = false ) {
+	public function get_document_number( bool $increment = false ) {
```
alexmigf, line 297: Maybe `$increment` could be `$generate`? What do you think?

alexmigf (changes requested):

PR #835, opened 2024-07-16
Fix: `get_header_logo_id()` return type fatal error

### LINKED ISSUE #834: get_header_logo_id(): Return value must be of type int, null returned
A customer reported the following error:

> Fatal error: WPO\WC\PDF_Invoices\Documents\Order_Document::get_header_logo_id(): Return value must be of type int, null returned

alexmigf:
closes #834

### DIFF
--- a/includes/documents/abstract-wcpdf-order-document.php
+++ b/includes/documents/abstract-wcpdf-order-document.php
@@ -847,7 +847,7 @@ abstract class Order_Document {
 	public function header_logo(): void {
 		$attachment_id = $this->get_header_logo_id();
 
-		if ( $attachment_id > 0 ) {
+		if ( ! is_null( $attachment_id ) && $attachment_id > 0 ) {
 			$company         = $this->get_shop_name();
 			$attachment_src  = wp_get_attachment_image_url( $attachment_id, 'full' );
 			$attachment_path = wp_normalize_path( realpath( get_attached_file( $attachment_id ) ) );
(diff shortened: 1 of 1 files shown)


MohamadNateqi (commented):
What do you think of checking the return value after applying the filter [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/835/files#diff-b4b45c08877ccc6f215e534426a81b3edd82ce371cfc3fce7420e62e279b3c1eR830) and then returning `0` if it's anything but numeric? This way we can keep the return type as `int`.

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -817,17 +817,19 @@ public function has_header_logo() {
...
 
 		if ( ! empty( $this->settings['header_logo'] ) ) {
 			$attachment_id  = $this->get_settings_text( 'header_logo', $header_logo_id, false );
 			$header_logo_id = is_numeric( $attachment_id ) ? absint( $attachment_id ) : 0;
 		}
+		
+		$header_logo_id = apply_filters( 'wpo_wcpdf_header_logo_id', $header_logo_id, $this );
 
-		return apply_filters( 'wpo_wcpdf_header_logo_id', $header_logo_id, $this );
+		return $header_logo_id ?? 0;
```
MohamadNateqi, line 832: Would it be good also to check if the value is numeric in case the filter returns any other types?

MohamadNateqi (approved):
LGTM!

MohamadNateqi (approved):
LGTM!

PR #838, opened 2024-07-19
Improve document numbers logging

### LINKED ISSUE #809: Improve document numbers logging
To enhance document management, we can create a log that registers everything related to the document number and maybe have separate logs for each document.

BrunoPavlinic98:
closes #809

### DIFF
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -321,10 +321,16 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
 /**
  * Logger function to capture errors thrown by this plugin, uses the WC Logger when possible (WC3.0+)
  */
-function wcpdf_log_error( $message, $level = 'error', $e = null ) {
+function wcpdf_log_error( $message, $level = 'error', $e = null, $document = null ) {
 	if ( function_exists( 'wc_get_logger' ) ) {
 		$logger  = wc_get_logger();
-		$context = array( 'source' => 'wpo-wcpdf' );
+		$source = 'wpo-wcpdf';
+
+		if($document) {
+			$source.= '-' . $document->get_type() . '-' . $document->get_number()->get_formatted();
+		}
+
+		$context = array( 'source' => $source );
 
 		if ( is_callable( array( $e, 'getFile' ) ) && is_callable( array( $e, 'getLine' ) ) ) {
 			$message = sprintf( '%s (%s:%d)', $message, $e->getFile(), $e->getLine() );
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -323,18 +323,18 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
  * 
  * @param string         $message
  * @param string         $level
- * @param \Exception     $e
+ * @param object         $e
  * @param Order_Document $document
  * 
  * @return void
  */
-function wcpdf_log_error( string $message, string $level = 'error', \Exception $e = null, WPO\WC\PDF_Invoices\Documents\Order_Document $document = null ): void {
+function wcpdf_log_error( string $message, string $level = 'error', object $e = null, WPO\WC\PDF_Invoices\Documents\Order_Document $document = null ): void {
 	if ( function_exists( 'wc_get_logger' ) ) {
 		$logger  = wc_get_logger();
-		$source = 'wpo-wcpdf';
+		$source  = 'wpo-wcpdf';
 
-		if($document) {
-			$source.= '-' . $document->get_type() . '-' . $document->get_number()->get_formatted();
+		if( $document ) {
+			$source .= '-' . $document->get_type();
 		}
 
 		$context = array( 'source' => $source );
--- a/includes/wcpdf-functions.php
+++ b/includes/wcpdf-functions.php
@@ -321,10 +321,10 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
 /**
  * Logger function to capture errors thrown by this plugin, uses the WC Logger when possible (WC3.0+)
  * 
- * @param string         $message
- * @param string         $level
- * @param object         $e
- * @param Order_Document $document
+ * @param string $message
+ * @param string $level
+ * @param object $e
+ * @param WPO\WC\PDF_Invoices\Documents\Order_Document $document
  * 
  * @return void
  */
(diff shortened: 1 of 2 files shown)


includes/wcpdf-functions.php
```diff
@@ -320,11 +320,24 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
...
+ * 
+ * @param string         $message
+ * @param string         $level
+ * @param \Exception     $e
+ * @param Order_Document $document
+ * 
+ * @return void
  */
-function wcpdf_log_error( $message, $level = 'error', $e = null ) {
+function wcpdf_log_error( string $message, string $level = 'error', \Exception $e = null, WPO\WC\PDF_Invoices\Documents\Order_Document $document = null ): void {
```
alexmigf, line 331: I think `$e` can be a lot more than only `\Exception`. Maybe we should not pass it here, it can be added to the `$message` where is called.

includes/wcpdf-functions.php
```diff
@@ -320,11 +320,24 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
...
 	if ( function_exists( 'wc_get_logger' ) ) {
 		$logger  = wc_get_logger();
-		$context = array( 'source' => 'wpo-wcpdf' );
+		$source = 'wpo-wcpdf';
+
+		if($document) {
```
alexmigf, line 336: WPCS: spacing

includes/wcpdf-functions.php
```diff
@@ -320,11 +320,24 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
...
+			$source.= '-' . $document->get_type() . '-' . $document->get_number()->get_formatted();
```
alexmigf, line 337: `$source .=`

alexmigf, line 337: This will create a log for each document number, I'm not sure if this is the best solution. I think having one for each document type would be preferable.

alexmigf (changes requested):

includes/wcpdf-functions.php
BrunoPavlinic98: @alexmigf I think that it won't hurt if we leave it here (but remove the `\Exception` parameter type). It find it useful in `catch` blocks where we have an Exception already defined and ready to be logged. What do you think?

includes/wcpdf-functions.php
BrunoPavlinic98: I agree that this will produce lots of different logs, and I agree that we could have one log for each document type. I did it this way because I thought that `separate logs for each document` in issue description means `1 document = 1 log`. But of course, `1 document type = 1 log` also makes sense.
@MohamadNateqi do you agree with `1 document type = 1 log` approach?

MohamadNateqi: @BrunoPavlinic98 I also agree `1 document type = 1 log` would be better.

includes/wcpdf-functions.php
alexmigf: Ok, let's try that. You can pass it as `object`.

includes/wcpdf-functions.php
```diff
@@ -320,11 +320,24 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
...
+ * @return void
  */
-function wcpdf_log_error( $message, $level = 'error', $e = null ) {
+function wcpdf_log_error( string $message, string $level = 'error', object $e = null, WPO\WC\PDF_Invoices\Documents\Order_Document $document = null ): void {
 	if ( function_exists( 'wc_get_logger' ) ) {
 		$logger  = wc_get_logger();
-		$context = array( 'source' => 'wpo-wcpdf' );
+		$source  = 'wpo-wcpdf';
+
+		if( $document ) {
```
alexmigf, line 336: `if (`

includes/wcpdf-functions.php
```diff
@@ -320,11 +320,24 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
...
+			$source .= '-' . $document->get_type();
+		}
+
+		$context = array( 'source' => $source );
 
 		if ( is_callable( array( $e, 'getFile' ) ) && is_callable( array( $e, 'getLine' ) ) ) {
 			$message = sprintf( '%s (%s:%d)', $message, $e->getFile(), $e->getLine() );
```
alexmigf, line 343: Not your fault, but maybe this should be:

```php
$message .= sprintf( ' (%s:%d)', $e->getFile(), $e->getLine() );
```

includes/wcpdf-functions.php
```diff
@@ -320,11 +320,24 @@ function wcpdf_deprecated_function( $function, $version, $replacement = null ) {
 
 /**
  * Logger function to capture errors thrown by this plugin, uses the WC Logger when possible (WC3.0+)
+ * 
+ * @param string         $message
+ * @param string         $level
+ * @param object         $e
+ * @param Order_Document $document
```
alexmigf, line 327: Should be `WPO\WC\PDF_Invoices\Documents\Order_Document`

alexmigf (changes requested):

alexmigf:
I'm not sure if this improvement is effective. It seems like all the errors we're seeing are common across every document, so we're ending up with different logs that show the same errors. I believe it might be more useful to have separate logs for main objects like settings, documents, and so on.

What do you think?

MohamadNateqi:
@alexmigf I also agree that this improvement does not seem to be so effective.

> I believe it might be more useful to have separate logs for main objects like settings, documents, and so on.

Could you please explain your idea more? How does it help in logging document numbers?

alexmigf:
To log only the document numbers, I'm uncertain if this implementation is correct. However, we do have the number tools table. What impact will this have?

MohamadNateqi:
Maybe we can have separate logs for each document number, and also, update the current log to collect more data and maybe registering new logs?

alexmigf:
> have separate logs for each document number

You will end up with a large number of log files, which doesn't seem practical.

> update the current log to collect more data and maybe registering new logs

What data do you think is missing or should be recorded that isn't currently?

MohamadNateqi:
> You will end up with a large number of log files, which doesn't seem practical.

I think I didn't express myself well. I meant one log file for each document type, just for numbers, like `...invoice-numbers`.

> What data do you think is missing or should be recorded that isn't currently?

In the recent issue we had in the `initiate_number()`, we registered several more logs just to track the flow. Maybe we can register some of the logs we did that time for the test? However, I don't recall it right now, and I need to check the code.

But besides this, what's your idea?

alexmigf:
I think we have a good amount of information right now:

* Number tools
* Order notes
* Order meta
* Semaphore logs

If any error occurs when getting the number it should be displayed in the general log file I think.

MohamadNateqi:
You're right. I also agree. 
So this PR and the issue need to be closed?

alexmigf:
This is just my opinion, but maybe we could improve the numbers DB table instead of removing numbers from it. For example, we could add an `active` column. I think you also suggested something similar in the past. We can discuss this further in a new issue.

alexmigf:
Yes, let's open a new issue to discuss that, but it's not urgent.

MohamadNateqi:
Should I open the new issue?

alexmigf:
If we don't have anything related already opened yes.

BrunoPavlinic98:
@alexmigf @MohamadNateqi Agreed. I will close this issue and PR.

PR #842, opened 2024-07-29
New: tool to remove expired Semaphore locks

### LINKED ISSUE #841: Semaphore expired locks are not being removed from the options table
A [user](https://wordpress.org/support/topic/database-options-table-gets-flooded/) reported that the Semaphore expired locks are not getting removed from the options page.

alexmigf:
closes #841

### DIFF
--- a/includes/class-wcpdf-updraft-semaphore.php
+++ b/includes/class-wcpdf-updraft-semaphore.php
@@ -8,70 +8,64 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 if ( ! class_exists( '\\WPO\\WC\\PDF_Invoices\\Updraft_Semaphore_3_0' ) ) :
 
-/**
- * Class Updraft_Semaphore_3_0
- *
- * This class is much simpler to use than the the previous series, as it has dropped support for complicated cases that were not being used. It also now only uses a single row in the options database, and takes care of creating it itself internally.
- *
- * Logging, though, may be noisier, unless your loggers are taking note of the log level and only registering what is required.
- *
- * Example of use (a lock that will expire if not released within 300 seconds)
- *
- * See test.php for a longer example (including logging).
- *
- * $my_lock = new Updraft_Semaphore_3_0('my_lock_name', 300);
- * // If getting the lock does not succeed first time, try again up to twice
- * if ($my_lock->lock(2)) {
- *   try {
- *     // do stuff ...
- *   } catch (Exception $e) {
- *     // We are making sure we release the lock in case of an error
- *   } catch (Error $e) {
- *     // We are making sure we release the lock in case of an error
- *   }
- *   $my_lock->release();
- * } else {
- *   error_log("Sorry, could not get the lock");
- * }
- */
 class Updraft_Semaphore_3_0 {
-
-	// Time after which the lock will expire (in seconds)
-	protected $locked_for;
-
-	// Name for the lock in the WP options table
+	
+	/**
+	 * Name for the lock in the WP options table
+	 *
+	 * @var string
+	 */
 	protected $option_name;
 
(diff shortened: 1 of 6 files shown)


includes/views/advanced-tools.php
```diff
@@ -57,6 +57,33 @@
 			</form>
 		</div>
 		<!-- /clear_tmp -->
+		<!-- clear_semaphore_expired_locks -->
+		<div class="tool">
```
MohamadNateqi, line 61: It's not only related to this PR, but maybe showing a spinner icon after clicking on the tool buttons would be good. This way, customers can be certain that the button is functioning and that a process is underway. 
What do you think?

includes/views/advanced-tools.php
```diff
@@ -57,6 +57,33 @@
...
+			<p><?php _e( 'Clean up the expired locks from the database.', 'woocommerce-pdf-invoices-packing-slips' ); ?></p>
+			<form method="post" id="clear_semaphore_expired_locks">
+				<input type="hidden" name="debug_tool" value="clear_semaphore_expired_locks">
+				<input type="submit" class="button button-secondary submit" value="<?php _e( 'Remove expired locks', 'woocommerce-pdf-invoices-packing-slips' ); ?>">
+				<fieldset>
+					<div class="notice inline" style="display:none;"><p></p></div>
+				</fieldset>
+			</form>
+			<?php $semaphore_expired_locks = \WPO\WC\PDF_Invoices\Updraft_Semaphore_3_0::count_expired_locks(); ?>
+			<?php if ( $semaphore_expired_locks > 0 ) : ?>
```
MohamadNateqi, line 72: Would it be good to inform there are no expired locks?

MohamadNateqi (changes requested):

BrunoPavlinic98:
@alexmigf I think the code is OK, I just have 2 suggestions for improvement, but I see that @MohamadNateqi already mentioned that:

1. I think that we should always show the number of expired locks. That way, the customer knows that he cleaned them all and that he does not need to do that again.
2. First time I tried it, I had 151 expired locks, and it took some time (2 or 3 seconds) for them to be removed. In that time, I did not receive any information about the status of that job. I think that we should put a spinner in order to show the customer that the job is still in progress.

What do you think?

includes/class-wcpdf-updraft-semaphore.php
```diff
@@ -153,48 +146,54 @@ public function lock($retries = 0) {
...
+	 * @param string $level   - the message level (debug, notice, info, warning, error)
+	 * @param array  $context - Optional. Additional information for log handlers.
+	 * 
+	 * @return void
 	 */
-	public function log( $message, $level = 'info', $context = array() ) {
-		$context = ! empty( $context ) ? $context : $this->context;
+	public function log( string $message, string $level = 'info', array $context = array() ): void {
+		$context      = ! empty( $context ) ? $context : $this->context;
+		$logs_enabled = isset( WPO_WCPDF()->settings->debug_settings['semaphore_logs'] );
```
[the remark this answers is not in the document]
alexmigf: I believe it depends on our settings, as we have an option that enables this type of logging. I prefer having a single check here rather than multiple checks throughout the plugin.

includes/views/advanced-tools.php
alexmigf: I think we could do that in a new issue because we could add to other tools as well. Do you want to do that?

MohamadNateqi: Sure.

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/843

MohamadNateqi (approved):
Great improvement.

BrunoPavlinic98 (approved):
LGTM!

PR #846, opened 2024-08-01
Fix unblocking UI when a tool button is clicked

### LINKED ISSUE #843: Add a spinner for buttons of Tools page
It would be good to show a spinner icon after clicking on the tool buttons. This way, customers can be sure that the button is functioning and that a process is underway.

MohamadNateqi:
close #843 

The unblocking action was immediately called after the AJAX request. Since the AJAX request is an async process, the unblock action next to it immediately ran. Moving the unblocking code inside the AJAX response fixed the issue.


alexmigf (approved):
LGTM!

PR #849, opened 2024-08-08
Add sections to settings

### LINKED ISSUE #668: Create settings sections
As our plugin has evolved over the years, the number of settings has grown significantly. This expansion has led to a cluttered and overwhelming settings page, making it challenging for users to locate and manage specific configurations.

These issues will be addressed when we create a whole new UI for the plugin early next year. But for now it might be nice to already create distinct sections for the available settings.

MohamadNateqi:
close #668

This PR aims to add sections to settings.

image

Each section is an accordion, as shown in the image. When you first visit the settings page, only the General section is open, while the rest are closed. Any changes made to the state of the accordions will be saved in the localStorage. The localStorage has been chosen to prevent the unnecessary overhead of storing the states in the database and making queries.

For settings that are added by the Professional plugin, two approaches exist:
1. Disable settings sections if the Pro extension doesn't support it
2. Group uncategorized settings in a new section.

I implemented the second approach and created a "More" section to keep the rest of the uncategorized settings.

image

Also, the helpers method has been created to modify sections more easily. To see how sections can be added for the Pro extension, you can switch to this PR's branch: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/443.

Since we used the word "section" for [another purpose](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/main/includes/settings/class-wcpdf-settings-debug.php#L927), I used the word "category" in my codes.


BrunoPavlinic98 (approved):
LGTM!
Categorized sections look OK to me, and I have no negative remarks for the code.

Terminator-Barbapapa:
@MohamadNateqi @alexmigf @BrunoPavlinic98 I made a few style tweaks. Let me know what you think.

MohamadNateqi:
It looks nice.

I find it a little hard to distinguish between the title and content here:

image

It would be good if you could somehow improve that.

BrunoPavlinic98:
> @Terminator-Barbapapa It looks nice.
> 
> I find it a little hard to distinguish between the title and content here:
> 
> image
> 
> It would be good if you could somehow improve that.

@Terminator-Barbapapa I think it looks and works great.

Regarding the hard distinguish issue @MohamadNateqi mentioned, maybe we could put the Enable option (and all other "standalone" options, if there are any more of them) in their standalone div (the lighter div with gray border), the same one that is being shown when we open a particular section. If not that, we can provide tabulation to it so that standalone options are in the same level as the section grouped options.

Terminator-Barbapapa:
I agree that this option seems out of place. I think apart from some changes in the styling we should also look at where the options are placed, as I think we can create a more logical layout. In this case I think the 'Enable' option would fit well in the 'General' section. Like this:

Scherm­afbeelding 2024-09-26 om 10 56 24

<br><br>
In a separate issue I think we can make some additional styling improvements, where we update the primary color to `#6e1edc` and introduce sliders to replace checkboxes. Like this:

Scherm­afbeelding 2024-09-26 om 10 46 51

BrunoPavlinic98:
> I agree that this option seems out of place. I think apart from some changes in the styling we should also look at where the options are placed, as I think we can create a more logical layout. In this case I think the 'Enable' option would fit well in the 'General' section. Like this:
> 
> Scherm­afbeelding 2024-09-26 om 10 56 24
> In a separate issue I think we can make some additional styling improvements, where we update the primary color to `#6e1edc` and introduce sliders to replace checkboxes. Like this:
> 
> Scherm­afbeelding 2024-09-26 om 10 46 51

@Terminator-Barbapapa I agree to put the Enable option to general setting, and I agree with sliders.

includes/class-wcpdf-settings.php
```diff
@@ -1005,6 +1009,182 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+		// Check for any unprocessed settings fields.
+		$unprocessed_settings_fields = array_diff_key( $settings_fields, array_flip( $processed_keys ) );
+
+		// Create a "More" section for uncategorized settings fields.
+		if ( ! empty( $unprocessed_settings_fields ) ) {
+			$category_name              = 'more';
+			$modified_settings_fields[] = array(
+				'type'     => 'section',
+				'id'       => $category_name,
+				'title'    => __( 'More', 'woocommerce-pdf-invoices-packing-slips' ),
```
Terminator-Barbapapa, line 1107: I think we should rename this to 'Additional settings'.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -585,6 +585,63 @@ public function get_ubl_settings_fields( $option_name ) {
...
+				)
+			),
+			'due_date'         => array(
+				'title'   => __( 'Due date', 'woocommerce-pdf-invoices-packing-slips' ),
+				'members' => array(
+					'due_date',
+				)
+			),
+			'admin_display'    => array(
+				'title'   => __( 'Admin display options', 'woocommerce-pdf-invoices-packing-slips' ),
```
Terminator-Barbapapa, line 622: Perhaps we can abbreviate this to just 'Admin', as there might be other options in the future that might fall into this section but are not strictly 'Admin display options'.

includes/documents/class-wcpdf-invoice.php
```diff
@@ -585,6 +585,63 @@ public function get_ubl_settings_fields( $option_name ) {
...
+					'display_date',
+					'display_number',
+					'next_invoice_number' // this should follow 'display_number'
+				)
+			),
+			'due_date'         => array(
+				'title'   => __( 'Due date', 'woocommerce-pdf-invoices-packing-slips' ),
+				'members' => array(
+					'due_date',
+				)
```
Terminator-Barbapapa, line 619: I think the due date option should be moved to the end of the 'Document details' section.

includes/documents/class-wcpdf-packing-slip.php
```diff
@@ -177,6 +177,32 @@ public function init_settings() {
...
+				'members' => array( 'enabled' ),
+			),
+			'general'          => array(
+				'title'   => __( 'General', 'woocommerce-pdf-invoices-packing-slips' ),
+				'members' => array(
+					'attach_to_email_ids',
+				),
+			),
+			'document_display' => array(
+				'title'   => __( 'Document display options', 'woocommerce-pdf-invoices-packing-slips' ),
```
Terminator-Barbapapa, line 193: Can be renamed to 'Document details'.

Terminator-Barbapapa (changes requested):

alexmigf:
I noticed that the document title and the related settings, which used to be at the top, are now placed below everything. This change feels a bit less intuitive.

I have a few suggestions to improve the user experience:

* Rename "Document details" to "Display," as it primarily contains display settings.
* Move the "Additional settings" section to be directly below the "General" section for a more logical flow.
* Rename "Additional settings" to "Document details" for better clarity.
* Move the "Next invoice number (without prefix/suffix, etc.)" setting to the "General" section.

assets/js/admin-script.js
```diff
@@ -585,4 +585,34 @@ jQuery( function( $ ) {
...
+
+		// Show the first accordion by default if there's no state saved
+		if ( ! state_exists ) {
+			$( '.settings_category h2:first' ).addClass( 'active' ).next( '.form-table' ).show();
+		}
+
+		$('.settings_category h2' ).click( function() {
+			const index = $( '.settings_category h2' ).index( $( this ) );
+
+			$( this ).toggleClass( 'active' ).next( '.form-table' ).slideToggle( 'fast', function() {
```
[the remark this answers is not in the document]
MohamadNateqi: @Terminator-Barbapapa For the full-screen mode, where the `display: table;`, I think there is no easy way to fix it.

One way would be to make the `display` property `block`, then make the transition, and then revert it to `table`, which has other side effects and issues, such as the bad display of elements during the transition.

Another option would be to switch to height transition instead of toggling the display.

Another option would be wrapping the `form-table` around a DIV element and working with that, but this is not a good way as it requires many modifications.

I tried to fix it using the first option, but several issues appeared that made me think this was not a good approach.

If we're going to fix this issue for full-screen mode, then switching to height transition might be the better option here.

Would you happen to have any suggestions or opinions?

alexmigf:

assets/js/admin-script.js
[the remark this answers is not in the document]
Terminator-Barbapapa: I think we can achieve the same animation without `slideToggle()` by manipulating the `max-height` property and set a `transition` in CSS.

MohamadNateqi:
@Terminator-Barbapapa, I updated the style and did several tests, and it seems to be fine. Could you please review it?

Terminator-Barbapapa:
I think we need some adjusted values for `grid-template-columns` and `grid-column-gap` on small and medium screen sizes.

Scherm­afbeelding 2024-10-17 om 16 09 55

_Current medium screen size_

Scherm­afbeelding 2024-10-17 om 16 09 34

_Current small screen size_

MohamadNateqi:
@Terminator-Barbapapa Can you please test it now?

Terminator-Barbapapa (approved):
Looking good. Great work! 🚀

alexmigf:
I'm looking at this from the customer's perspective and noticed a few potential UX issues that could be improved:

* The separation between the number format and the next number feels a bit odd, especially since they both relate to the document number.
* The 'Allow My Account invoice download' setting might not need to be placed at the top of the list—it doesn’t seem as important.
* While we’ve updated the PDF output styling, the UBL format is still using the old design.

assets/js/admin-script.js
```diff
@@ -596,4 +596,30 @@ jQuery( function( $ ) {
...
+
+		// Retrieve the state from localStorage
+		$( '.settings_category h2' ).each( function( index ) {
+			const state = localStorage.getItem( 'wcpdf_accordion_state_' + index );
+			if ( 'true' === state ) {
+				$( this ).addClass( 'active' ).next( '.form-table' ).show();
+			}
+		} );
+
+		$('.settings_category h2' ).click( function() {
```
alexmigf, line 612: Maybe use `on`?

includes/Documents/Invoice.php
```diff
@@ -619,6 +619,60 @@ public function get_ubl_settings_fields( $option_name ) {
...
+					'reset_number_yearly',
+					'mark_printed',
+					'unmark_printed',
+					'disable_free',
+					'use_latest_settings',
+				)
+			)
+		);
+
+		return apply_filters( "wpo_wcpdf_{$this->type}_settings_categories", $settings_categories, $this );
```
alexmigf, line 673: Maybe it's better to avoid the type in the filter name, you can use the `$this`.

includes/Documents/PackingSlip.php
```diff
@@ -208,6 +208,34 @@ public function init_settings() {
...
+				'members' => array(
+					'display_billing_address',
+					'display_email',
+					'display_phone',
+					'display_customer_notes',
+				)
+			),
+		);
+
+		return apply_filters( "wpo_wcpdf_{$this->type}_settings_categories", $settings_categories, $this );
```
alexmigf, line 236: Same.

includes/Settings.php
```diff
@@ -1008,6 +1012,182 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
 		return $new_settings;
 	}
 
+	public function update_documents_settings(): void {
```
alexmigf, line 1015: Consider including `sections` in the function name to make it clearer that you're modifying specific sections rather than the overall document settings.

includes/Settings.php
```diff
@@ -1008,6 +1012,182 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+		$documents = WPO_WCPDF()->documents->get_documents( 'all' );
+
+		foreach ( $documents as $document ) {
+			$document = str_replace( '-', '_', $document->get_type() );
+			add_filter( 'wpo_wcpdf_settings_fields_documents_' . $document, array( $this, 'apply_settings_categories' ), 999, 4 );
```
alexmigf, line 1020: This is not the correct filter, you should use this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/a4e63addbf0abcbe25d5655287d4b8e4ad4bf005/includes/Documents/Invoice.php#L194

Maybe this filter needs to be updated for the Packing Slip.

Terminator-Barbapapa:
> The separation between the number format and the next number feels a bit odd, especially since they both relate to the document number.

I'm not against moving it to the 'Document details' section, underneath 'Next invoice number'. And we could move the 'Display due date' underneath 'Display invoice date'. Also when thinking about this it might make more sense to move invoice date and due date underneath the invoice number options. So it would become:

- Display email address
- Display phone number
- Display customer notes
- Display shipping address
- Display invoice number
- Next invoice number
- Number format
- Display invoice date
- Display due date

>The 'Allow My Account invoice download' setting might not need to be placed at the top of the list—it doesn’t seem as important.
Do you mean that on your local environment this option is placed at the top of the General section? For me it is at the bottom of the General section, which seems reasonable to me. Or do you mean this option does not belong in the General section?

>While we’ve updated the PDF output styling, the UBL format is still using the old design.
I agree we can also place these settings in a General section.

includes/Settings.php
```diff
@@ -1008,6 +1012,182 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+		}
+
+		return $modified_settings_fields;
+	}
+
+	/**
+	 * Helper method to add a setting field to a category.
+	 *
+	 * @param array           $settings_categories
+	 * @param string|array    $new_setting_id
```
alexmigf, line 1132: If this is a helper method, why not enforce strict variable types to ensure more reliable and predictable behavior?

alexmigf (changes requested):

alexmigf:
> I'm not against moving it to the 'Document details' section, underneath 'Next invoice number'. And we could move the 'Display due date' underneath 'Display invoice date'. Also when thinking about this it might make more sense to move invoice date and due date underneath the invoice number options. So it would become:

I agree.

> Do you mean that on your local environment this option is placed at the top of the General section? For me it is at the bottom of the General section, which seems reasonable to me. Or do you mean this option does not belong in the General section?

In my opinion, this doesn't belong in the 'General' tab since it doesn't carry that level of importance. I think it would be more appropriate to place it under the 'Advanced' tab.

includes/Settings.php
MohamadNateqi: I found it useful to allow bulk fields addition to categories, like [here](https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/443/files#diff-175a10c63e9e3b23d5635a9bcd094f3d0285a0688e248fe064fb86689b952218R245-R251).

Do you suggest to make the `$new_setting_id` only accept `string`?

alexmigf: Maybe in a new function with that explained?

alexmigf:
> I don't think I agree with that. I think configuring access on the My Account page for the invoice (or other documents) is quite a basic feature of our plugin, and one that should be in sight when setting it up. So I don't think we should hide it in the 'Advanced' section.

I'm not suggesting this setting is advanced, but it feels more in line with options like resetting numbers yearly. It's a bit odd that something previously at the bottom is now at the top, especially since we've never had any complaints about its position before. That said, I'm fine with keeping it at the top.

includes/Settings.php
MohamadNateqi: Maybe in the future, we can update the `init_settings()` method in the `OrderDocument` class to something like this:

```php
public function init_settings() {
	do_action( "wpo_wcpdf_before_{$this->type}_init_settings", $this );

	foreach ( $this->output_formats as $output_format ) {
		$page            = $option_group = $option_name = "wpo_wcpdf_documents_settings_{$this->get_type()}";
		$settings_fields = is_callable( array( $this, "get_{$output_format}_settings_fields" ) ) ?
			$this->{"get_{$output_format}_settings_fields"}( $option_name ) :
			array();

		// Apply legacy filter.
		$settings_fields = apply_filters( "wpo_wcpdf_settings_fields_documents_{$this->get_type()}", $settings_fields, $page, $option_group, $option_name );

		// Allow plugins to alter settings fields.
		$settings_fields = apply_filters( "wpo_wcpdf_settings_fields_documents_{$this->type}_{$output_format}", $settings_fields, $page, $option_group, $option_name, $this );

		if ( ! empty( $settings_fields ) ) {
			WPO_WCPDF()->settings->add_settings_fields( $settings_fields, $page, $option_group, $option_name );
		}
	}

	do_action( "wpo_wcpdf_after_{$this->type}_init_settings", $this );
}
```

This is similar to the function implemented for the `Invoice` class. In this way, the child class only implements a getter for the settings field, and this function also gets called in the constructor of the `OrderDocument`.

MohamadNateqi:
@Terminator-Barbapapa @alexmigf,
Style issues have been fixed, settings have been reordered, and UBL is supported now.

Terminator-Barbapapa (approved):

includes/Documents/PackingSlip.php
```diff
@@ -201,11 +201,50 @@ public function init_settings() {
 			$settings_fields = WPO_WCPDF()->settings->move_setting_after_id( $settings_fields, $pro_notice, 'enabled' );
 		}
 
-		// allow plugins to alter settings fields
+		// Legacy filter to allow plugins to alter settings fields.
 		$settings_fields = apply_filters( 'wpo_wcpdf_settings_fields_documents_packing_slip', $settings_fields, $page, $option_group, $option_name );
```
alexmigf, line 205: Maybe this filter has to be changed?

includes/Documents/PackingSlip.php
```diff
@@ -201,11 +201,50 @@ public function init_settings() {
...
+						'display_email',
+						'display_phone',
+						'display_customer_notes',
+						'display_billing_address',
+					),
+				),
+			),
+		);
+
+		return apply_filters( "wpo_wcpdf_document_settings_categories", $settings_categories[ $output_format ], $output_format, $this );
```
alexmigf, line 247: Single quote.

includes/Settings.php
```diff
@@ -1008,6 +1012,213 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+	 * Applies categories to document settings.
+	 *
+	 * @return void
+	 */
+	public function update_documents_settings_sections(): void {
+		$documents = WPO_WCPDF()->documents->documents;
+
+		foreach ( $documents as $document ) {
+			foreach ( $document->output_formats as $output_format ) {
+				add_filter( "wpo_wcpdf_settings_fields_documents_{$document->type}_{$output_format}", array( $this, 'apply_settings_categories' ), 999 );
```
alexmigf, line 1025: Use `get_type()` instead.

includes/Settings.php
```diff
@@ -1008,6 +1012,213 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+	 * Internal method to handle adding setting fields to a category.
+	 *
+	 * @param array           $settings_categories
+	 * @param array           $new_setting_ids
+	 * @param string          $category_name
+	 * @param string|int|null $position
+	 *
+	 * @return array
+	 */
+	private function add_setting_field_to_category_impl( array $settings_categories, array $new_setting_ids, string $category_name, $position = null ): array {
```
alexmigf, line 1170: Maybe `impl` is not needed?

includes/Settings.php
```diff
@@ -1008,6 +1012,213 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+		return $modified_settings_fields;
+	}
+
+	/**
+	 * Helper method to add a single setting field to a category.
+	 *
+	 * @param array           $settings_categories
+	 * @param string          $new_setting_id
+	 * @param string          $category_name
+	 * @param string|int|null $position
```
alexmigf, line 1138: Can we strict this to `int`only?

includes/Settings.php
```diff
@@ -1008,6 +1012,213 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+		return $this->add_setting_field_to_category_impl( $settings_categories, [ $new_setting_id ], $category_name, $position );
+	}
+
+	/**
+	 * Helper method to add multiple setting fields to a category.
+	 *
+	 * @param array           $settings_categories
+	 * @param array           $new_setting_ids
+	 * @param string          $category_name
+	 * @param string|int|null $position
```
alexmigf, line 1152: Same.

includes/Settings.php
```diff
@@ -1008,6 +1012,213 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+		return $this->add_setting_field_to_category_impl( $settings_categories, $new_setting_ids, $category_name, $position );
+	}
+
+	/**
+	 * Internal method to handle adding setting fields to a category.
+	 *
+	 * @param array           $settings_categories
+	 * @param array           $new_setting_ids
+	 * @param string          $category_name
+	 * @param string|int|null $position
```
alexmigf, line 1166: Same.

includes/Documents/Invoice.php
```diff
@@ -619,6 +619,78 @@ public function get_ubl_settings_fields( $option_name ) {
...
+					'members' => array(
+						'enabled',
+						'attach_to_email_ids',
+						'include_encrypted_pdf',
+					),
+				),
+			)
+		);
+
+		return apply_filters( "wpo_wcpdf_document_settings_categories", $settings_categories[ $output_format ], $output_format, $this );
```
alexmigf, line 691: Single quote.

alexmigf (changes requested):

includes/Settings.php
MohamadNateqi: This was actually a mistake when I was merging the main; thank you.

includes/Documents/PackingSlip.php
MohamadNateqi: Do you mean using `apply_filters_deprecated()`?

alexmigf: No, just append the output format.

includes/Settings.php
MohamadNateqi: I found using the setting name for the position so helpful. For example, if we only allow `int`, then we need to manually find the setting's position and add the setting to the correct position, like [here](https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/443/files#diff-175a10c63e9e3b23d5635a9bcd094f3d0285a0688e248fe064fb86689b952218R240-R241).

includes/Documents/PackingSlip.php
MohamadNateqi: Since this is the legacy filter, should we append the output format?
If we're going to change the filter name, then we can totally remove it, I think?

includes/Settings.php
MohamadNateqi: Maybe we can introduce a method to find the position of a setting, and then send it to the `add_setting_field_to_category` method?

includes/Documents/PackingSlip.php
alexmigf: I thought it was already deprecated. Perhaps we should address it in a separate issue.

includes/Settings.php
alexmigf: Good idea.

includes/Settings.php
```diff
@@ -1008,6 +1012,220 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
 		return $new_settings;
 	}
 
+	/**
+	 * Applies categories to document settings.
+	 *
+	 * @return void
+	 */
+	public function update_documents_settings_sections(): void {
+		$documents = WPO_WCPDF()->documents->documents;
```
alexmigf, line 1021: Maybe `get_documents()` instead?

includes/Settings.php
```diff
@@ -1008,6 +1012,220 @@ public function move_setting_after_id( $settings, $insert_settings, $after_setti
...
+			}
+		}
+
+		// Check for any unprocessed settings fields.
+		$unprocessed_settings_fields = array_diff_key( $settings_fields, array_flip( $processed_keys ) );
+
+		// Create an "Additional settings" section for uncategorized settings fields.
+		if ( ! empty( $unprocessed_settings_fields ) ) {
+			$category_name              = 'additional';
+			$modified_settings_fields[] = array(
```
alexmigf, line 1111: You’re repeating this array; consider extracting it into a new function.

alexmigf (changes requested):

alexmigf (approved):
LGTM!

PR #850, opened 2024-08-09
Add 'You might also like' section to the Upgrade tab.

### LINKED ISSUE #847: Update the Upgrade tab
As we will stop selling Premium Templates as a single plugin (but only as part of the PDF Invoice Bundle) we will need to change the feature table on the 'Upgrade' tab. The Premium Templates column can be removed.

### LINKED ISSUE #848: Add 'You might also like' section to the Upgrade tab.
Underneath the features table on the Upgrade tab I would like a 'You might also like...' section that contains three additional choices. I would say: Smart Reminder Emails, Address Labels and Automatic Order Printing.

If one of these plugins is currently installed, we can fade them slightly and add a 'Currently installed' label.

BrunoPavlinic98:
closes #848 
closes #847


includes/views/upgrade-table.php
```diff
@@ -95,3 +96,48 @@
...
+				<a href="https://wpovernight.com/downloads/woocommerce-reminder-emails-payment-reminders/" target="_blank">
+					<?php esc_html_e( 'Visit plugin site', 'woocommerce-pdf-invoices-packing-slips' ); ?>
+				</a>
+			</span>
+		</td>
+	</tr>
+	<tr>
+		<td class="feature-label">
+			<?php esc_html_e( 'WooCommerce Print Address Labels', 'woocommerce-pdf-invoices-packing-slips' ); ?>
+			<?php echo isset( $all_plugins['woocommerce-address-labels/woocommerce-address-labels.php'] ) ? '<sup class="wcpdf_beta">' . __( 'Currently installed', 'woocommerce-pdf-invoices-packing-slips' ) . '</sup>' : ''; ?>
```
MohamadNateqi, line 120: Would it be better to change this text to "Activate it" if the plugin is installed but not activated?

MohamadNateqi (changes requested):
Maybe it's better to have the same style we have on the website for it; I mean card styles:

image

@Terminator-Barbapapa What do you think?

includes/views/upgrade-table.php
BrunoPavlinic98: @MohamadNateqi I am not sure. Both ways make sense to me. Let's wait for @alexmigf and @Terminator-Barbapapa and their comments.

alexmigf: I don't have a strong opinion, maybe @Terminator-Barbapapa has a preference.

BrunoPavlinic98:
I have also added changes for https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/847 because I think those 2 can be merged togerher.

Terminator-Barbapapa:
It might be a nice touch if plugins that are already installed move to the right of the row.

BrunoPavlinic98:
@Terminator-Barbapapa I think I managed to create a grid system that will allow us to easily add new plugins customers might like.

assets/css/settings-styles.css
```diff
@@ -1152,6 +1152,81 @@ div.upgrade-table-description a,
...
+	padding: 3%;
+    background: white;
+    border-radius: 5px;
+    border: 1px solid #6e1edc;
+    color: #6e1edc;
+    cursor: pointer;
+	float: left;
+	font-weight: bold;
+    font-size: 1.2em;
+}
```
Terminator-Barbapapa, line 1212: There is already a button styled that is used in the features table: `a.upgrade_button`. You can use those styles including the styles for the `:after`, `:hover` and `:focus`.

assets/css/settings-styles.css
```diff
@@ -1152,6 +1152,81 @@ div.upgrade-table-description a,
...
 }
 
+.card-container {
+    border-radius: 5px;
+    width: calc(100% - 30px);
+    display: flex;
+    flex-wrap: wrap;
+    gap: 30px;
+    padding: 15px;
+    background-color: #f0f0f0;
```
Terminator-Barbapapa, line 1162: You can use `#f1e9fc` for the background color of the section. Which is a pale tint variation of the new primary brand color `#6e1edc`.

assets/css/settings-styles.css
```diff
@@ -1152,6 +1152,81 @@ div.upgrade-table-description a,
...
+	justify-content: center;
```
Terminator-Barbapapa, line 1163: I would let the cards left align.

assets/css/settings-styles.css
```diff
@@ -1152,6 +1152,81 @@ div.upgrade-table-description a,
...
+	padding-bottom: 50px;
+}
+
+.card {
+	border-radius: 5px;
+	margin-top: 0;
+	width: 25%;
```
Terminator-Barbapapa, line 1170: The cards need a max width to prevent them from becoming huge on larger screens. Perhaps a wrapper with a max width might be useful. And use either `flex` or `grid` to make three columns within this wrapper. The max width of the wrapper should more or less correspond with the width of the features table above it. So i would suggest a max width of something between 900px and 1200px.

It would be good to let the cards all have the same height no matter the content. I believe `grid` would let you more easily achieve this. But I'll leave it up to you how to achieve this, as I have no preference. 

Also the grid should be responsive. So on smaller screens the cards should stack underneath each other.

includes/views/upgrade-table.php
```diff
@@ -95,3 +95,39 @@
 		<td align="left" class="last">&nbsp;</td>
```
Terminator-Barbapapa, line 95: Currently the bottom row of the features table can display the wrong amount of cells. This is because we present information here that is using the license data from the installed plugins. To prevent a call to our license server each time the Upgrade tab is opened we cache this data. So in my case there still was a cell displayed for the Premium Templates extension. We need to make sure this is skipped.

Terminator-Barbapapa (changes requested):
Overall good work, but there are improvements to be made. 

Also I think if we are going to add plugin thumbnails in the plugin files we should make them as small as possible. I think a good start would be to use JPEG instead of PNG and make sure they are all the same resolution.

alexmigf:
@Terminator-Barbapapa  don't you think placing this section below the table might go unnoticed? It might be a good idea to add some UTM tags to track clicks and leads.

Terminator-Barbapapa:
@alexmigf I think in this case the upgrade table with the features of the extensions is more important, so I would like that to be at the top. Also I think if scrolling is required to get to the bottom of the features list, the recommendations will get into view automatically. But in the near future we could experiment with some A-B testing to see what is the most effective.

Also good point on the UTM tags. I forgot to mention that. The integration for this on the site is up an running so we should be able to just add the tags to all links and see the data in our dashboard. @BrunoPavlinic98 can you add these tags to the links?

Terminator-Barbapapa:
@BrunoPavlinic98 One additional thing: Currently the Upgrade tab is hidden when users have both extensions installed for the PDF Invoices plugin. We check for that with the `bundle_is_active()` function. But since we are now also offering recommendations I think we can remove that functionality, so the Upgrade tab is always visible.

assets/css/settings-styles.css
BrunoPavlinic98: @Terminator-Barbapapa I already use grid system, and I set it up in a way that all elements will have the height of the most higher element, so that they all look the same. That also goes for the installed plugin styles as well. Grid is responsible so on smaller screens, they will be stacked.

assets/css/settings-styles.css
BrunoPavlinic98: @Terminator-Barbapapa I agree. Just FYI, I have placed the justify-content: center on mobile phones media query because there, plugins will be stacked one under another, so it looks weird if they are aligned left.

BrunoPavlinic98:
> @BrunoPavlinic98 One additional thing: Currently the Upgrade tab is hidden when users have both extensions installed for the PDF Invoices plugin. We check for that with the `bundle_is_active()` function. But since we are now also offering recommendations I think we can remove that functionality, so the Upgrade tab is always visible.

@Terminator-Barbapapa Makes sense. I will remove it, but I am not sure if it is safe to remove the `bundle_is_active()`. It is `Settings_Upgrade`'s class public function, and even tough it is not being used in this plugin, I am not sure about some other plugins. What do you think?

assets/css/settings-styles.css
Terminator-Barbapapa: @BrunoPavlinic98 Currently cards with more content are longer than cards with less content:

Scherm­afbeelding 2024-08-23 om 13 11 39

Also the current grid stacks in a weird way because the content is centered. But it also allows for one card to jump to the next row. Because we are providing three options here I think it would work best if we keep the row of three in tact for as long as possible and then switch to a stacked layout where all cards are underneath each other on small screens.

Scherm­afbeelding 2024-08-23 om 13 12 34

includes/views/upgrade-table.php
BrunoPavlinic98: @Terminator-Barbapapa I have added `true` parameter for ignore_cache parameter, I think it helps:
`$extension_license_infos = $this->get_extension_license_infos( true );`

BrunoPavlinic98:
> @alexmigf I think in this case the upgrade table with the features of the extensions is more important, so I would like that to be at the top. Also I think if scrolling is required to get to the bottom of the features list, the recommendations will get into view automatically. But in the near future we could experiment with some A-B testing to see what is the most effective.
> 
> Also good point on the UTM tags. I forgot to mention that. The integration for this on the site is up an running so we should be able to just add the tags to all links and see the data in our dashboard. @BrunoPavlinic98 can you add these tags to the links?

@Terminator-Barbapapa @alexmigf Since I never worked with UTM tags, can you point to a place in code where they are located? On on which links do you want them to be placed? Only on plugin "Buy now" links?

includes/views/upgrade-table.php
Terminator-Barbapapa: We still want to cache the license data. So that solution won't work. This will need to be handled in `upgrade-table.php` file, in the table row where the upgrade links are generated.

Terminator-Barbapapa:
> I will remove it, but I am not sure if it is safe to remove the bundle_is_active() function. It is Settings_Upgrade's class public function, and even tough it is not being used in this plugin, I am not sure about some other plugins.

I don't believe this is used anywhere else. @alexmigf Can you confirm this?

alexmigf:
> I don't believe this is used anywhere else. @alexmigf Can you confirm this?

Yes, I confirm that.

includes/views/upgrade-table.php
BrunoPavlinic98: @Terminator-Barbapapa I was able to reproduce your issue and I managed to solve it.

BrunoPavlinic98:
@Terminator-Barbapapa This is ready for review, I think I addressed all review observations.

Terminator-Barbapapa (changes requested):

Terminator-Barbapapa:
@BrunoPavlinic98 @alexmigf @MohamadNateqi I added some small styling tweaks to the 'You might also like' section. Could you check if these work for you?

MohamadNateqi:
@Terminator-Barbapapa My first impression of this section was this:

_Why is everything too large? Why does it feel paled?_

image

Maybe it's just me, I don't know, but maybe making it smaller and changing the opacity for installed plugins would be better?

alexmigf:
> So perhaps we should not load the 'You might also like' section when a user already has all recommended plugins installed.

Maybe we could rotate plugins? Suggesting other WooCommerce plugins from our portfolio?

> Can you comment on the size? How does it look to you?

I don't have any issues with it, and it looks fine to me. Since my screen is large, everything is displayed clearly. I also noticed that the elements resize appropriately when adjusting the screen size, so I believe it should work well on lower screen resolutions too.

BrunoPavlinic98:
@Terminator-Barbapapa I think the size looks great. I tend to have larger elements always in case I have room because think they are easier to be spotted. Since we have room here, I would keep it as is.

I would not remove this section in case all plugins are installed. If they are here, it tells the user that we have a lot more to offer. And I don't have the feeling that we will achieve something if we remove them.

My opinion is that we leave everything as is, only change the heading from **You might also like...** to **You might also like this WP Overnight plugins...** - in that way, even if we have everything installed, the title could provide value to us by leading customers to our plugin shop.

@alexmigf @MohamadNateqi What do you think?

BrunoPavlinic98:
> @BrunoPavlinic98 maybe just **"You might also like these plugins..."** would be enough, makes the title shorter.

@alexmigf Makes sense, I agree with it!

Terminator-Barbapapa:
@BrunoPavlinic98 @alexmigf @MohamadNateqi I've added the additional recommendation and a link to the shop in the rare case that a user has all recommendations installed. Let me know your thoughts.

includes/views/upgrade-table.php
```diff
@@ -95,3 +93,35 @@
...
 	</tr>
 </table>
+
+<div id="plugin-recommendations" class="upgrade-table-description">
+	<h1><?php esc_html_e( 'You might also like these plugins...', 'woocommerce-pdf-invoices-packing-slips' ); ?></h1>
+	<?php
+	if ( count( array_column( $sorted_plugin_recommendations, 'installed' ) ) == count( $sorted_plugin_recommendations ) ) {
```
alexmigf, line 100: Maybe `===` since `count` returns `int`.

alexmigf (approved):
Seems fine to me :rocket:

assets/images/wc-reminder-emails-thumbnail-400x400.jpg
MohamadNateqi, line 1: I think all images could be compressed to reduce their size.

includes/views/upgrade-table.php
Terminator-Barbapapa: Fixed: [5f18511](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/850/commits/5f1851173cea7425cb4a29f3f81d8ab25c418f43)

BrunoPavlinic98:
> @BrunoPavlinic98 @alexmigf @MohamadNateqi I've added the additional recommendations and a link to the shop in the rare case that a user has all recommendations installed. Let me know your thoughts.

@Terminator-Barbapapa I like this!
I have also compressed the images as @MohamadNateqi suggested.

BrunoPavlinic98:
@Terminator-Barbapapa It looks OK to me now!

Terminator-Barbapapa:
@MohamadNateqi Images are compressed now.

MohamadNateqi:
@Terminator-Barbapapa I think it would be good to also convert them to WEBP. It provides smaller file sizes than JPG and keeps the quality. In my test, it was reduced by almost 5 to 10 KB for each image.

BrunoPavlinic98:
> @Terminator-Barbapapa I think it would be good to also convert them to WEBP. It provides smaller file sizes than JPG and keeps the quality. In my test, it was reduced by almost 5 to 10 KB for each image.

@MohamadNateqi if the quality remains the same then I think this is a way to go. Can you please push those files since you already have them so I can check them out?

Terminator-Barbapapa:
@MohamadNateqi Yes, could you push the WEBP files? I also did some test with WEBP but kept getting artifacts and noise in the images when they went below the file size of the compressed JPG files.

MohamadNateqi:
@Terminator-Barbapapa @BrunoPavlinic98 Updated here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/850/commits/1865c750f9ed9d8ba9dc134432e6d58d8943899f.

All image sizes are 155 KB with JPG and 125 KB with WEBP.

MohamadNateqi (approved):

PR #852, opened 2024-08-16
Add a note to inform about more documents

### LINKED ISSUE #851: Adding a note to inform about other documents
It might be good to have a text below the document dropdown to inform customers about other kinds of documents we offer. It can be such text:
_Need more documents? Please visit this link._

This would direct customers to an article on our documentation and provide information about the documents available in the Pro plugin, and more.

MohamadNateqi:
close #851 

image

We need to create the article in our documentation before merging this PR.


includes/settings/class-wcpdf-settings-documents.php
```diff
@@ -69,6 +69,16 @@ public function output( $section ) {
 				}
 				?>
 			</ul>
+			<p><i>
+				<?php
+					printf(
+						/* translators: 1. open anchor tag, 2. close anchor tag */
+						__( 'Need more document? Please visit %1$shere%2$s.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 76: I think this should be behind a Pro check. We don't want to promote this for the free plugin users.

alexmigf (changes requested):

alexmigf:
@dpeyou can you create a documentation for this? I don't think we should promote this for the free users, or even provide zip files for the custom documents.

includes/settings/class-wcpdf-settings-documents.php
MohamadNateqi: What if we inform free users about the Pro plugin documents in the documentation article?
Do you think it would be good to keep this for free users then?

dpeyou:
@alexmigf @MohamadNateqi 
Draft 1. 
Feedback please!

---

# More Document Types

The [Professional extension](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) of our Invoicing plugin offers access to the following new PDF types: 
- Proforma Invoice
- Receipt
- Credit Note


An image showing the document types<br>

## Proforma Invoice
This could be used as a secondary or alternate Invoice.
Remember that all document types can be renamed via the plugin settings.

## Receipt
Some places require to deliver separate Invoice and Receipt documents. Our plugin is therefore suitable for those cases, as well.

## Credit Note 
This serves as the Refund document. This could be thought of as an Invoice for refunds, or a Credit Invoice. The reason for the existence of this document, is because we follow **GAAP** (Generally Accepted Accounting Principles) - which states that: 

***Administrative changes should be documented separately. This means that an original invoice should never be modified by a refund. In most countries (at least in the EU) this is even considered fraudulent (and illegal).***

includes/settings/class-wcpdf-settings-documents.php
alexmigf: Yes, I agree with that.

alexmigf:
@dpeyou 

> The [Professional extension](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) of our Invoicing plugin offers access to the following new PDF types:

Maybe "document types" instead of "PDF types".

includes/settings/class-wcpdf-settings-documents.php
MohamadNateqi: So, is the current text fine?

MohamadNateqi:
@alexmigf @dpeyou What do you think about mentioning the documents the Order Proposal plugin adds?

includes/settings/class-wcpdf-settings-documents.php
alexmigf: Seems fine to me.

alexmigf:
I agree.

dpeyou:
@alexmigf @MohamadNateqi 
## Draft 2
### Items Addressed :heavy_check_mark: 
> - Maybe "document types" instead of "PDF types". 
> - What do you think about mentioning the documents the Order Proposal plugin adds? 

--

# More Document Types

## 1. Professional extension
[PDF Invoices for WooCommerce - Professional extension](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) offers access to the following new document types: 
- Proforma Invoice
- Receipt
- Credit Note

An image showing additional document types<br>

### Proforma Invoice
This could be used as a secondary or alternate Invoice.
Remember that all document types can be renamed via the plugin settings.

### Receipt
Some places require to deliver separate Invoice and Receipt documents. Our plugin is therefore suitable for those cases, as well.

### Credit Note 
This serves as the Refund document. This could be thought of as an Invoice for refunds, or a Credit Invoice. The reason for the existence of this document, is because we follow **GAAP** (Generally Accepted Accounting Principles) - which states that: 

***Administrative changes should be documented separately. This means that an original invoice should never be modified by a refund. In most countries (at least in the EU) this is even considered fraudulent (and illegal).***
<br>

## 2. Order Proposal plugin
[WooCommerce Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/) plugin is a standalone plugin of its own. In combination with [PDF Invoices for WooCommerce - Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/), it offers access to the following additional document types:
- Proposal
- Order Confirmation

An image showing the document types<br>

### Proposal
The document created when an order proposal/quote has been created by either the customer or the site admin.

### Order Confirmation 
This document could optionally be used to confirm a proposal, as that would mean confirming an order.

MohamadNateqi:
@dpeyou Would it be better to put this "Remember that all document types can be renamed via the plugin settings." out of the Proforma section? As it seems to be general for all documents.

dpeyou:
@MohamadNateqi yeah, I think that makes sense.

Draft 2.1
Addressed:
- Text moved, image added, text added, based on the comment below:
> @dpeyou Would it be better to put this "Remember that all document types can be renamed via the plugin settings." out of the Proforma section? As it seems to be general for all documents.
<br>

--

# More Document Types

## 1. Professional extension
[PDF Invoices for WooCommerce - Professional extension](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) offers access to the following new document types: 
- Proforma Invoice
- Receipt
- Credit Note

An image showing additional document types
<br>

Note: Remember that all document types can be renamed via the plugin settings, thanks to this extension 
Path: **WooCommerce > PDF Invoices > Documents > Document title**.

An image of the document title setting - ips pro

### Proforma Invoice
This could be used as a secondary or alternate Invoice.

### Receipt
Some places require to deliver separate Invoice and Receipt documents. Our plugin is therefore suitable for those cases, as well.

### Credit Note 
This serves as the Refund document. This could be thought of as an Invoice for refunds, or a Credit Invoice. The reason for the existence of this document, is because we follow **GAAP** (Generally Accepted Accounting Principles) - which states that: 

***Administrative changes should be documented separately. This means that an original invoice should never be modified by a refund. In most countries (at least in the EU) this is even considered fraudulent (and illegal).***
<br>

## 2. Order Proposal plugin
[WooCommerce Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/) plugin is a standalone plugin of its own. In combination with [PDF Invoices for WooCommerce - Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/), it offers access to the following additional document types:
- Proposal
- Order Confirmation

An image showing the document types<br>

### Proposal
The document created when an order proposal/quote has been created by either the customer or the site admin.

### Order Confirmation 
This document could optionally be used to confirm a proposal, as that would mean confirming an order.

dpeyou:
Here is Draft 2.1, published:
https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/more-document-types/

alexmigf:

alexmigf:
@dpeyou Could you please publish the page so @MohamadNateqi can update the link?

alexmigf (commented):
Perhaps this should only be displayed when Pro is not enabled?  

Also, the text feels a bit off. Would this be better?

> Looking for more documents? Learn more here.

MohamadNateqi:
> Perhaps this should only be displayed when Pro is not enabled?

I think we discussed adding a section at the end of the documentation page to inform customers they can email us for additional documents not included even in the Pro version. So this is why we wanted to have it always.

Do you think it would be better to remove it when Pro is installed?
<hr>

> Also, the text feels a bit off. Would this be better?

Both seem good to me, I have no preference. I'll update the text.

alexmigf:
I believe it would be better to display the message only when Pro is not enabled. Additionally, we should aim to minimize requests for custom documents, as they are not our primary focus.

dpeyou:
@alexmigf this has already been published: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/852#issuecomment-2316187127

includes/Settings/SettingsDocuments.php
```diff
@@ -75,7 +75,7 @@ public function output( $section ) {
 					printf(
 						/* translators: 1. open anchor tag, 2. close anchor tag */
 						__( 'Looking for more documents? Learn more %1$shere%2$s.', 'woocommerce-pdf-invoices-packing-slips' ),
-						'<a href="https://docs.wpovernight.com/">', // ToDo: Update URL
+						'<a href="https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/more-document-types/">', // ToDo: Update URL
```
alexmigf, line 78: You could remove the comment now.

alexmigf (changes requested):

PR #855, opened 2024-09-02
Simplify due date setting

### LINKED ISSUE #853: Simplify the due date option
I think the due date option can be simplified in the free version, by not providing preset options. I think this setting should always be an input field, that defaults to 30 days when enabled. 

So it could look something like this:

Scherm­afbeelding 2024-08-30 om 17 00 52

The input can be disabled when the checkbox is unchecked. Like this:

Scherm­afbeelding 2024-08-30 om 17 03 54

I think we can remove the placeholder option we provide in the Professional extension. Users can simply uncheck the checkbox and use the placeholder in the Customizer if they need to place the due date in a custom position.

MohamadNateqi:
close #853 

The setting now looks like this:

image


assets/js/admin-script.js
```diff
@@ -105,6 +105,17 @@ jQuery( function( $ ) {
...
+		const $due_date_checkbox   = $( '#wpo-wcpdf-preview-wrapper #due_date' );
+		const $due_date_days_input = $( '#wpo-wcpdf-preview-wrapper #due_date_days' );
+
+		if ( $due_date_checkbox.is( ':checked' ) ) {
+			$due_date_days_input.prop( 'disabled', false );
+		} else {
+			$due_date_days_input.prop( 'disabled', true );
+		}
+	} ).trigger( 'change' );
+
```
Terminator-Barbapapa, line 118: I think we also should provide some visual feedback, showing the field is disabled. Also the text in the field should align right so it looks better in combination with the word 'days' next to it.

Scherm­afbeelding 2024-09-13 om 16 00 58

You should be able to achieve that with the following styles:

```
#wpo-wcpdf-preview-wrapper input#due_date_days {
	text-align: right;
}

#wpo-wcpdf-preview-wrapper input#due_date_days:disabled {
	background-color: #eaeaea;
	color: #999;
}
```

includes/documents/class-wcpdf-invoice.php
```diff
@@ -286,18 +286,17 @@ public function get_pdf_settings_fields( $option_name ) {
...
-					), $this->type ),
-					'description'	=> __( 'Displays a due date below the order data.', 'woocommerce-pdf-invoices-packing-slips' ),
+					'option_name'        => $option_name,
+					'id'                 => 'due_date',
+					/* translators: number of days */
+					'text_input_wrap'    => __( '%s days', 'woocommerce-pdf-invoices-packing-slips' ),
+					'text_input_size'    => 1,
+					'text_input_id'      => 'due_date_days',
+					'text_input_default' => 30,
+					'description'        => __( 'Displays a due date below the order data.', 'woocommerce-pdf-invoices-packing-slips' ),
```
Terminator-Barbapapa, line 299: This description can be removed, as I think the due date should always be presented underneath the invoice date.

Terminator-Barbapapa (changes requested):
I see the default base date for the due date is currently the order date. It should always be the invoice date. Therefore I think we can also remove the base date option from Pro. The due date should always reflect the timeframe between when the invoice was issued and when it is due.

Also I see the due date is currently added to the bottom of the order data with an action hook. Because it is so intertwined with the invoice date I think it makes more sense to always display it underneath the invoice date. Which means it should be added into the templates.

MohamadNateqi:
@Terminator-Barbapapa, I applied the changes and moved the due date to the template.

I also need to update templates for the Premium Templates plugin. 
I had to keep the action hook but added only a few conditions, so no issues will happen for customers using an older PT version.

I wondered if adding more actions to templates for the `order-data` section would be good so these cases could be handled without modifying the templates.

Of course, I think a better approach might be creating functions to output the `order-data` section dynamically, but we need to discuss this at another time since then each section of templates could be modular and have a function to output data.

For example, something like this:

```php
function output_order_data( $document ): void {
	$fields = array(
		'display_number' => array(
			'class'     => 'invoice-number',
			'title'     => $document->number_title(),
			'value'     => $document->number( $document->get_type() ),
			'condition' => isset( $document->settings['display_number'] )
		),
		'display_date'   => array(
			'class'     => 'invoice-date',
			'title'     => $document->date_title(),
			'value'     => $document->date( $document->get_type() ),
			'condition' => isset( $document->settings['display_date'] )
		),
		'due_date'       => array(
			'class'     => 'due-date',
			'title'     => $document->due_date_title(),
			'value'     => $document->due_date(),
			'condition' => $document->get_due_date() > 0
		),
		// rest of rows
	];

	$fields = apply_filters( 'wpo_wcpdf_templates_order_data_fields', $fields, $document );

	// Loop through each field
	foreach ( $fields as $field ) {
		if ( ! empty( $field['condition'] ) ) {
			echo '<tr class="', esc_attr__( $field['class'] ), '">
			<th>', esc_html__( $field['title'] ), '</th>
			<td>', esc_html__( $field['value'] ), '</td>
		</tr>';
		}
	}
}
```

Terminator-Barbapapa:
I do think it could have added value to output certain sections in a template via a function like you described. This has to be done with caution though. Everything should be backwards compatible and custom setups should not break.

Speaking of custom setups, I think the current implementation of showing the due date for legacy templates does not take custom templates into account. When running a custom template the due date is now only displayed if the user has version `2.21.9` or lower of the Premium Templates installed. For other users the due date simply is removed from their templates where before it was displayed.

Also I'm not sure if we should call `display_due_date()` from within `class-wcpdf-main.php`. This seems to cause some issues in the PDF preview when switching between templates. For instance when I have the 'Simple' template set and then select the 'Simple Premium' template, the due date disappears from the preview. When I have the 'Simple Premium' template set and then select the 'Simple' template, the preview shows the due date twice.

I would propose adding the due date through the `wpo_wcpdf_after_order_data` hook in `wcpdf-functions.php`, and performing the required checks within that function like this:

```
// Add due date via action hook for legacy templates
add_action( 'wpo_wcpdf_after_order_data', 'wpo_wcpdf_due_date_table_row', 10, 2 );
function wpo_wcpdf_due_date_table_row( string $document_type = null, $order = null ): void {

	$current_template_path = explode( '/', WPO_WCPDF()->settings->get_template_path() );
	$current_template	   = end( $current_template_path );
	$premium_templates     = array( 'Simple Premium', 'Modern', 'Business' );
	$document              = wcpdf_get_document( $document_type, $order );

	// Simple template is selected. Due date is displayed through template
	if ( 'Simple' == $current_template ) {
		return;
	}
	
	// Updated Premium Template is selected. Due date is displayed through template
	if ( function_exists( 'WPO_WCPDF_Templates' ) && version_compare( WPO_WCPDF_Templates()->version, '2.21.9', '>' ) && in_array( $current_template, $premium_templates ) ) {
		return;
	}

	if ( empty( $order ) || empty( $document_type ) || 'invoice' !== $document_type ) {
		return;
	}

	if ( ! $document ) {
		return;
	}

	$due_date_timestamp = is_callable( array( $document, 'get_due_date' ) ) ? $document->get_due_date() : 0;

	if ( 0 >= $due_date_timestamp ) {
		return;
	}

	$due_date       = apply_filters( 'wpo_wcpdf_due_date_display', date_i18n( wcpdf_date_format( $document, 'due_date' ), $due_date_timestamp ), $due_date_timestamp, $document_type, $document );
	$due_date_title = is_callable( array( $document, 'get_due_date_title' ) ) ? $document->get_due_date_title() : __( 'Due Date:', 'woocommerce-pdf-invoices-packing-slips' );

	if ( ! empty( $due_date ) ) {
	?>
		<tr class="due-date">
			<th><?= $due_date_title; ?></th>
			<td><?= $due_date; ?></td>
		</tr>
	<?php
	}
}
```

This way the due date functionality keeps working for custom templates as well as for users of legacy Premium Templates. And if a user wants to place the due date somewhere else in their custom template they can easily remove our placement by using:

```
remove_action( 'wpo_wcpdf_after_order_data', 'wpo_wcpdf_due_date_table_row', 10, 2 );
```

It also solves the issue with the preview. 

@MohamadNateqi @alexmigf Let me know what you think.

MohamadNateqi:
@Terminator-Barbapapa I actually forgot about custom templates to consider them; you're right. It needs modification to consider them.

However, regarding moving this function from the `class-wcpdf-main.php` to the `wcpdf-functions.php`, it might be better to keep it in the main file since functions defined in the `wcpdf-functions.php` are in the global scope, and I think we can consider them as a developer API to our plugin functionality. Because of this, maybe it's better to keep it in the main file. What do you think?

Terminator-Barbapapa:
I think calling the hook from within main causes some issues. For instance the issue with the preview as mentioned.

MohamadNateqi:
@Terminator-Barbapapa 

After applying your suggested improvement to the function, I couldn't see any issues with the preview. It seems the update also fixed the preview.
I’ve tested it with PT templates, custom templates, and the Simple templates, and everything appears to work fine with the preview. Could you please test it using my latest commit as well?

Please let me know if moving the code to the functions file is still preferred and better.

Terminator-Barbapapa (approved):
It seems the issue with the preview is indeed fixed, so I agree with keeping the `display_due_date_table_row()` method in main. 

And I see now that my concern about the fact that it would be hard to hide the default due date in a custom template is invalid as it can be removed with:

```
add_filter( 'wpo_wcpdf_due_date_display', '__return_false' );
```

And then added back in the custom template on a different position with the same code that is now used in the Simple template. Which will also keep the functionality of the 'Display due date' checkbox intact. 

I have two small issues though, but those can be fixed in separate issues later on:

1. The due date days input is currently a text input. Ideally this would be a number input.
2. Using the `disabled` property on the due date days input clears any input in the field when saving the options. This means that if you've set the due date to a custom number of days and uncheck the checkbox and save the options, the field is cleared and it will fall back to the default 30 days. Although this is a bit of an edge case and not a huge problem.

Having said that, I approve this PR. Good work! 🚀

MohamadNateqi:
@Terminator-Barbapapa I created a new issue to address those issues: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/865. Thank you for informing me about them.

alexmigf:
@MohamadNateqi can you fix the conflicts with `main` please?

alexmigf (changes requested):

includes/documents/abstract-wcpdf-order-document.php
```diff
@@ -1463,23 +1473,34 @@ public function get_number_store_year( $table_name ) {
...
-		$base_date         = apply_filters( 'wpo_wcpdf_due_date_base_date', $this->order->get_date_created(), $this->type, $this );
-		$due_date_datetime = clone $base_date;
-		$due_date_datetime = $due_date_datetime->modify( "+$due_date_days days" );
+		$document_creation_date = $this->get_date( $this->get_type(), $this->order ) ?? new \WC_DateTime( 'now', new \DateTimeZone( 'UTC' ) );
+		$base_date              = apply_filters( 'wpo_wcpdf_due_date_base_date', $document_creation_date, $this->get_type(), $this );
+		$due_date_datetime      = clone $base_date;
+		$due_date_datetime      = $due_date_datetime->modify( "+$due_date_days days" );
 
-		return apply_filters( 'wpo_wcpdf_due_date', $due_date_datetime->getTimestamp() ?? 0, $this->type, $this );
+		return apply_filters( 'wpo_wcpdf_due_date', $due_date_datetime->getTimestamp() ?? 0, $this->get_type(), $this );
```
alexmigf, line 1494: I think we don't need the type here too?

alexmigf (approved):
:rocket:

PR #856, opened 2024-09-03
New: automatic cleanup of unlocked Semaphore locks. Other improvements.

### LINKED ISSUE #854: Clean up locks
@alexmigf The current code is using per-order lock names like `updraft_lock_wpo_wcpdf_main_semaphore_lock_order_2177401_email_customer_completed_order`.

On the site I'm looking at currently, there are up to 5 separate locks per order (one lock per order status). The locks aren't being cleaned up, and consequently, these locks account for over 40% of rows in the `options` table (or over 90% once I eliminate a large number of rows form a long-running WooCommerce bug which also creates zillions of rarely-used rows).

Solution:

1) To clean-up already-existing ones: delete unused locks (where the `value` is `0`)

2) To stop new ones accumulating: for rarely-used locks (e.g. ones with per-order names), call the `delete()` method in the `Updraft_Semaphore_3_0` class after releasing the lock.

Step 1) could also be run as a scheduled task periodically to catch any that aren't properly released/deleted.

alexmigf:
closes #854

### DIFF
--- a/includes/class-wcpdf-updraft-semaphore.php
+++ b/includes/class-wcpdf-updraft-semaphore.php
@@ -9,16 +9,16 @@ if ( ! defined( 'ABSPATH' ) ) {
 if ( ! class_exists( '\\WPO\\WC\\PDF_Invoices\\Updraft_Semaphore_3_0' ) ) :
 
 class Updraft_Semaphore_3_0 {
-	
+
 	/**
 	 * Prefix for the lock in the WP options table
 	 *
 	 * @var string
 	 */
 	protected static $option_prefix = 'wpo_ips_semaphore_lock_';
-	
+
 	/**
-	 * Hook name suffix for the cleanup of unlocked locks
+	 * Hook name suffix for the cleanup of released locks
 	 *
 	 * @var string
 	 */
@@ -261,13 +261,13 @@ class Updraft_Semaphore_3_0 {
 	}
 
 	/**
-	 * Cleanup unlocked locks from the database
+	 * Cleanup released locks from the database
 	 *
 	 * @return void
 	 */
-	public static function cleanup_unlocked_locks(): void {
+	public static function cleanup_released_locks(): void {
 		global $wpdb;
-		
+
 		$wpdb->query(
 			$wpdb->prepare(
 				"DELETE FROM {$wpdb->options} WHERE option_name LIKE %s AND option_value = '0'",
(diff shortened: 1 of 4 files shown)


MohamadNateqi:
I wonder if the "unlocked locks" combination could be a little confusing. What do you think about "released locks"?

includes/class-wcpdf-updraft-semaphore.php
```diff
@@ -247,43 +261,109 @@ public function get_loggers(): array {
...
 
 		return $count;
 	}
 	
+	/**
+	 * Get the hook name for the cleanup of unlocked locks
+	 *
+	 * @return string - the hook name
+	 */
+	private static function get_hook_name(): string {
```
MohamadNateqi, line 302: What do you think about `get_cleanup_hook_name()`?

includes/class-wcpdf-updraft-semaphore.php
```diff
@@ -247,43 +261,109 @@ public function get_loggers(): array {
...
+	public static function is_cleanup_scheduled(): bool {
+		return function_exists( 'as_next_scheduled_action' ) && as_next_scheduled_action( self::get_hook_name() );
+	}
+	
+	/**
+	 * Get the next scheduled cleanup of unlocked locks
+	 *
+	 * @return object|null - the next scheduled cleanup action or null
+	 */
+	public static function get_cleanup_action() {
```
MohamadNateqi, line 320: Return type declaration.

includes/class-wcpdf-updraft-semaphore.php
```diff
@@ -247,43 +261,109 @@ public function get_loggers(): array {
...
+	}
+	
+	/**
+	 * Schedule the cleanup of unlocked locks
+	 *
+	 * @return void
+	 */
+	public static function schedule_semaphore_cleanup(): void {
+		if ( ! self::is_cleanup_scheduled() ) {
+			$interval = apply_filters( self::get_hook_name() . '_interval', 30 * DAY_IN_SECONDS ); // default: every 30 days
```
MohamadNateqi, line 349: Would it be useful to provide a filter for the days?

MohamadNateqi (changes requested):

includes/class-wcpdf-updraft-semaphore.php
MohamadNateqi: Sorry, my mistake. No problem, all good.

MohamadNateqi (approved):
LGTM!

DavidAnderson684 (approved):

PR #866, opened 2024-09-27
Tweak: lower hook priority for Invoice columns in WC orders list

alexmigf:
Adjusted the priority from `999` to `200` to ensure compatibility with Admin Columns plugin: https://www.admincolumns.com/


MohamadNateqi (approved):
LGTM!

PR #869, opened 2024-10-02
Remove path to generic_document.png

MohamadNateqi:
There is no `generic_document.png` in the images, while it has been used for the documents icon.


alexmigf (commented):
Maybe having that file back would be better? I made the new ones, I can do something simple.

MohamadNateqi:
That would be great.
I'm no expert in creating these SVGs, so I just removed it.

alexmigf:
What about this?

generic

MohamadNateqi:
This is great, I'll add it. Thanks

alexmigf:
@MohamadNateqi consider reverting the last commit, merging `main`, and then making the changes afterward.

MohamadNateqi:
@alexmigf Done.

alexmigf:
The parent constructor must be called before setting properties in the derived classes.

Screenshot from 2024-10-04 08-14-09

alexmigf:
Or maybe just move the icon after it.

alexmigf:
After reconsidering, it might be best to hold off on this change for now, as it will affect the Pro extension documents.

MohamadNateqi:
As you said, I think the parent constructor should be called first, but for now, I can revert the last commit, and we can fix it later in another issue.

alexmigf (approved):

PR #871, opened 2024-10-02
Fix: load plugin translations later on `init`

### LINKED ISSUE #870: Load plugin translations on `init`
In WordPress 6.7, new `_doing_it_wrong` messages were [introduced](https://core.trac.wordpress.org/changeset/59127), resulting in these messages appearing in the error log, for example:

> [02-Oct-2024 07:28:57 UTC] PHP Notice:  Function load_plugin_textdomain was called <strong>incorrectly</strong>. Attempted to load translations for the <code>woocommerce-pdf-invoices-packing-slips</code> domain too early. Translations should be loaded after the <code>after_setup_theme</code> action has fired, to ensure that the current user is already set up. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.) in /home/alex/www/wpovernight/dev/wp-includes/functions.php on line 6099

We need to ensure they are loaded after the `after_setup_theme` hook, so I recommend using `init` or `wp_loaded`.

alexmigf:
closes #870

### Remarks:

- Translations must be loaded before the classes to ensure that all translatable strings are available when needed.
- The classes should be initialized with a lower priority compared to the Pro extension to prevent potential conflicts or issues.

### DIFF
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -70,7 +70,7 @@ class WPO_WCPDF {
 		require $this->plugin_path() . '/vendor/autoload.php';
 
 		// load the localisation & classes
-		add_action( 'plugins_loaded', array( $this, 'translations' ) );
+		add_action( 'init', array( $this, 'translations' ) );
 		add_action( 'plugins_loaded', array( $this, 'load_classes' ), 9 );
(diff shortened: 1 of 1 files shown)


MohamadNateqi (approved):
LGTM!

MohamadNateqi (approved):
LGTM!

PR #872, opened 2024-10-03
Check current user role in the `enable_debug` function

MohamadNateqi:
Prevent unauthenticated users to enable debug and error reporting.

### DIFF
--- a/includes/class-wcpdf-main.php
+++ b/includes/class-wcpdf-main.php
@@ -1561,11 +1561,15 @@ class Main {
 	}
 
 	/**
-	 * Enable PHP error output
+	 * Enable PHP error output for administrators.
 	 */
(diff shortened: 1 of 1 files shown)


alexmigf (commented):
Maybe we could use `user_can_manage_settings()` instead?

MohamadNateqi:
@alexmigf, The `user_can_manage_settings()` function checks for `manage_woocommerce` permission and also allows to change it using a filter.

Do you think it's OK to allow other roles to capabilities to enable the debug?

alexmigf:
Yes, if they are in management roles. I believe that question would also be appropriate for the settings, wouldn't it?

alexmigf (approved):
LGTM!

PR #874, opened 2024-10-04
Introduce `calculate_due_date()` function

MohamadNateqi:
This PR introduces the `calculate_due_date()` function to allow third parties to calculate the due date despite the current document settings.


includes/Documents/OrderDocument.php
```diff
@@ -1795,11 +1807,11 @@ public function get_due_date(): int {
 		);
 		$due_date_days = apply_filters( 'wpo_wcpdf_document_due_date_days', $due_date_days, $this );
 
-		if ( 0 >= intval( $due_date_days ) ) {
+		if ( intval( $due_date_days ) <= 0 ) {
 			return 0;
 		}
 
-		$document_creation_date = $this->get_date( $this->get_type(), $this->order ) ?? new \WC_DateTime( 'now', new \DateTimeZone( 'UTC' ) );
+		$document_creation_date = $this->get_date( $this->get_type(), $order ) ?? new \WC_DateTime( 'now', new \DateTimeZone( 'UTC' ) );
```
alexmigf, line 1814: It might be better to use `wc_timezone_string()` instead of hardcoding `UTC`.

alexmigf (changes requested):

includes/Documents/OrderDocument.php
```diff
@@ -1808,6 +1808,18 @@ public function get_due_date(): int {
 			return 0;
 		}
 
+		return $this->calculate_due_date( $due_date_days, $this->order );
```
alexmigf, line 1811: Is it necessary to pass the order as a parameter, considering it’s always accessible via the document object?

MohamadNateqi: Good point. Fixed.

alexmigf: Remember to update here: https://github.com/wpovernight/woocommerce-pdf-ips-templates/pull/340

alexmigf (approved):

PR #880, opened 2024-10-08
Fix: adds missing deprecated filters to document title methods

### LINKED ISSUE #879: Legacy filter is not applying on template titles methods
The legacy filters are ignored in the current template titles function like here:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/main/includes/documents/class-wcpdf-invoice.php#L54-L57

alexmigf:
closes #879

### DIFF
--- a/includes/documents/class-wcpdf-invoice.php
+++ b/includes/documents/class-wcpdf-invoice.php
@@ -54,7 +54,8 @@ class Invoice extends Order_Document_Methods {
 	public function get_title() {
 		// override/not using $this->title to allow for language switching!
 		$title = __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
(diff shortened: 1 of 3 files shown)


includes/documents/class-wcpdf-invoice.php
```diff
@@ -53,29 +53,32 @@ public function storing_settings_enabled() {
 	 */
 	public function get_title() {
 		// override/not using $this->title to allow for language switching!
-		return apply_filters( 'wpo_wcpdf_document_title', __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' ), $this );
+		$title = __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' );
+		return apply_filters_deprecated( "wpo_wcpdf_{$this->slug}_title", array( $this->title, $this ), '3.8.7', 'wpo_wcpdf_document_title' );
```
MohamadNateqi, line 57: Is there any difference which one comes first?

alexmigf: Fixed.

PR #885, opened 2024-10-10
Modify the product bundle item in the order items

MohamadNateqi:
This is related to this issue: https://github.com/wpovernight/woocommerce-pdf-ips-templates/issues/306.


MohamadNateqi:
This approach is not preferred.

PR #886, opened 2024-10-10
only error out on "header_logo" if both "src" and "path" are empty.

commercial-hippie:
hi

I was debugging an issue where the logo wasn't showing on a site that's using wp-stateless:
https://wordpress.org/plugins/wp-stateless/

I think the `wp_normalize_path` function was failing here, but that shouldn't matter if we only need the URL, so I just moved the check down a few lines and added `\add_filter('wpo_wcpdf_use_path', '__return_false');` to our site, and now it seems to work.

thanks!


alexmigf (approved):
Looks good to me.

Thank you so much, for your contribution!

MohamadNateqi (approved):
LGTM!

PR #888, opened 2024-10-16
Tweak: updates `sabre/xml` library to version 4

alexmigf:
### PR Notes:

- Upgraded `sabre/xml` to version `4.0.6`
- Added proper charset handling to the `wcpdf_ubl_headers()` function
- Introduced new filter: `wcpdf_ubl_headers_charset` for customizing the charset
- Fixed encoding issues in the UBL implementation
- Implemented a PHP version check for UBL compatibility and display a notice if requirements are not met

The `sabre/xml` library already handles escaping effectively without causing any issues. Therefore, we should avoid implementing it ourselves.


MohamadNateqi (approved):
LGTM!

alexmigf:
The `sabre/xml` library already handles escaping effectively without causing any issues. Therefore, we should avoid implementing it ourselves.

MohamadNateqi (approved):
Nice improvements!

PR #895, opened 2024-10-25
Updating the setup Wizard

### LINKED ISSUE #857: Updating the setup Wizard
The goal of this task is to improve the current (wp-admin/admin.php?page=wpo-wcpdf-setup) PDF Invoices onboarding process.

We want to update the styles and see if we can shorten the amount of steps, and highlight the features of our paid extensions with a link to the upgrade tab.

BrunoPavlinic98:
closes #857


BrunoPavlinic98:
@Terminator-Barbapapa Here is the initial version. I think this is quite of a progress in terms of design. On the final step, the user gets redirected to the upgrade page after he clicks the "Finish" button.

BrunoPavlinic98:
@Terminator-Barbapapa I have added enhanced select.

I have deliberately added the following lines:
```
<link rel="stylesheet" href="<?php echo WC()->plugin_url() . '/assets/css/admin.css'; ?>">
<script src="<?php echo WC()->plugin_url() . '/assets/js/select2/select2.full.min.js'; ?>"></script>
```
because for some reason, `wp_enqueue_style`, `wp_register_script`, and `wp_enqueue_style` did not work for me.

With that, I am not sure why JS minifying is not working, I will try to reboot it.

includes/SetupWizard.php
```diff
@@ -145,6 +151,8 @@ public function setup_wizard_header() {
 			<?php wp_print_scripts( 'wpo-wcpdf-setup-confetti' ); ?>
 			<?php do_action( 'admin_print_styles' ); ?>
 			<?php do_action( 'admin_head' ); ?>
+			<link rel="stylesheet" href="<?php echo WC()->plugin_url() . '/assets/css/admin.css'; ?>">
```
alexmigf, line 154: You need to enqueue these two assets and then print them here using `wp_print_scripts`. Refer to the example on lines `150` and `151`.

includes/SetupWizard.php
```diff
@@ -91,6 +91,12 @@ public function setup_wizard() {
 			array( 'dashicons', 'install' ),
 			WPO_WCPDF_VERSION
 		);
+		wp_enqueue_style(
+			'wpo-wcpdf-toggle-switch',
+			WPO_WCPDF()->plugin_url() . '/assets/css/toggle-switch'.$suffix.'.css',
+			array( 'dashicons', 'install' ),
```
alexmigf, line 97: Are these dependencies in use?

views/setup-wizard/attach-to.php
```diff
@@ -10,28 +10,18 @@
...
-		);
 
-	}
+	$attach_to_setting = array(
+		'option_name'     => 'wcpdf_settings[wpo_wcpdf_documents_settings_invoice]',
+		'id'              => 'attach_to_email_ids',
+		'options'         => $wc_emails,
+		'multiple'        => true,
+		'enhanced_select' => true,
+		'default'         => isset( $current_settings['attach_to_email_ids'] ) ? array_keys( array_filter( $current_settings['attach_to_email_ids'], function( $value ) { return $value === '1'; } ) ) : array(),
```
alexmigf, line 20: I'm unsure if we should provide a default here, as it's not something we do in the settings.

includes/SetupWizard.php
alexmigf, line 154: Can you clarify the purpose of the `admin.css` file?

alexmigf (changes requested):

includes/SetupWizard.php
BrunoPavlinic98: @alexmigf With the new implementation, I don't even need to print the styles.

BrunoPavlinic98: We need `admin.css` for enhanced select styles.

includes/SetupWizard.php
```diff
@@ -94,7 +94,13 @@ public function setup_wizard() {
...
 			WPO_WCPDF()->plugin_url() . '/assets/css/toggle-switch'.$suffix.'.css',
-			array( 'dashicons', 'install' ),
+			array(),
+			WPO_WCPDF_VERSION
+		);
+		wp_enqueue_style(
+			'woocommerce_admin_styles',
+			WC()->plugin_url() . '/assets/css/admin.css',
+			array(),
 			WPO_WCPDF_VERSION
```
alexmigf, line 104: Maybe this could be the Woo version.

includes/SetupWizard.php
```diff
@@ -151,7 +157,6 @@ public function setup_wizard_header() {
 			<?php wp_print_scripts( 'wpo-wcpdf-setup-confetti' ); ?>
 			<?php do_action( 'admin_print_styles' ); ?>
 			<?php do_action( 'admin_head' ); ?>
-			<link rel="stylesheet" href="<?php echo WC()->plugin_url() . '/assets/css/admin.css'; ?>">
 			<script src="<?php echo WC()->plugin_url() . '/assets/js/select2/select2.full.min.js'; ?>"></script>
```
alexmigf, line 160: Could you consider enqueuing this properly as well?

alexmigf (changes requested):

includes/SetupWizard.php
BrunoPavlinic98: Done!

alexmigf (changes requested):

alexmigf (changes requested):
The "Set image" doesn't seem to be working as expected:

> Uncaught TypeError: $settings_wrapper.parent().block is not a function

BrunoPavlinic98:
> The "Set image" doesn't seem to be working as expected:
> 
> > Uncaught TypeError: $settings_wrapper.parent().block is not a function

@alexmigf Fixed! jQuery blockui was not added, that is the reason for the error.

alexmigf (changes requested):
I already have an image set, but it doesn't display when I run the wizard.

Screenshot from 2025-01-29 08-05-24

alexmigf:
@BrunoPavlinic98 I tried again, and it seems to be working now. I'm not sure if caching was the issue, even though I cleared it before running the test.

alexmigf (changes requested):
I installed this branch on a different local setup and found a conflict with the [Discount Rules for WooCommerce](https://wordpress.org/plugins/woo-discount-rules/) plugin. We should find a way to prevent this, as other plugins might use a similar approach.

Screenshot from 2025-01-29 08-14-58

BrunoPavlinic98:
> I installed this branch on a different local setup and found a conflict with the [Discount Rules for WooCommerce](https://wordpress.org/plugins/woo-discount-rules/) plugin. We should find a way to prevent this, as other plugins might use a similar approach.
> 
> Screenshot from 2025-01-29 08-14-58

@alexmigf How to I reproduce this behaviour? I installed that plugin, but I don't see the extra content.

alexmigf:
@BrunoPavlinic98 I have implemented a permanent fix for the issue.

BrunoPavlinic98:
> @BrunoPavlinic98 I have implemented a permanent fix for the issue.

@alexmigf Nice, I saw it. I think it is ready for review again.

alexmigf:
@BrunoPavlinic98 I believe [this](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/895#issuecomment-2620947326) issue warrants a more thorough review, especially since @Terminator-Barbapapa  has also [observed](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/895#issuecomment-2620962676) the same behavior.

BrunoPavlinic98:
> @BrunoPavlinic98 I believe [this](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/895#issuecomment-2620947326) issue warrants a more thorough review, especially since @Terminator-Barbapapa has also [observed](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/895#issuecomment-2620962676) the same behavior.

@alexmigf @Terminator-Barbapapa I tried to replicate this multiple times with cache cleaning, without in, in incognito windows and normal ones, but without luck.
Did you do anything else except the steps already described?

alexmigf:
Try this:

1. Switch to the `main` branch.  
2. Navigate to **PDF Invoices > General** and check if a logo is set. If not, add one.  
3. Click **Save Changes**.  
4. Switch to the `857-setup-wizard-update` branch.  
5. Clear your browser cache.  
6. Go to **PDF Invoices > Tools > Run the Setup Wizard**.  
7. Proceed to Step 2 and observe that the logo is missing.

BrunoPavlinic98:
> @BrunoPavlinic98 Try this:
> 
> 1. Switch to the `main` branch.
> 2. Navigate to **PDF Invoices > General** and check if a logo is set. If not, add one.
> 3. Click **Save Changes**.
> 4. Switch to the `857-setup-wizard-update` branch.
> 5. Clear your browser cache.
> 6. Go to **PDF Invoices > Tools > Run the Setup Wizard**.
> 7. Proceed to Step 2 and observe that the logo is missing.

@alexmigf Tried multiple times, but I don't have any issues. This happens on your environment every time you try?

alexmigf:
@BrunoPavlinic98 no, only for the first try. I will debug and fix it, if you don't mind.

BrunoPavlinic98:
> @BrunoPavlinic98 no, only for the first try. I will debug and fix it, if you don't mind.

@alexmigf If you have a solution, sure, I don't mind since I can't replicate this.

alexmigf:
@BrunoPavlinic98 this was missing: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/895/commits/2562ba32e804614a92a78c96d7dc802626694481#diff-e268583d2c221dba966bff13b95da8756b094132176e33cb049a07e91305aa52R15

Should be fixed now.

BrunoPavlinic98:
> @BrunoPavlinic98 this was missing: [2562ba3#diff-e268583d2c221dba966bff13b95da8756b094132176e33cb049a07e91305aa52R15](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/2562ba32e804614a92a78c96d7dc802626694481#diff-e268583d2c221dba966bff13b95da8756b094132176e33cb049a07e91305aa52R15)
> 
> Should be fixed now.

@alexmigf That is why I could not replicate it, I have a fresh setup and did not setup multilingual setup.
@Terminator-Barbapapa Can you also please confirm it is working?

alexmigf:
@BrunoPavlinic98 I didn't had a multilingual setup either.

Terminator-Barbapapa:
I think the enhanced select should follow the new color scheme. Currently it looks like this:

Scherm­afbeelding 2025-01-31 om 16 24 44

But I think it should look like this:

Scherm­afbeelding 2025-01-31 om 16 37 21

For the selected elements you can use:

```
background-color: #6e1edc;
border-bottom: 1px solid white;
color: white;
```

And for the hover you can use:

```
background-color: #f1e9fc;
color: #000;
```

Terminator-Barbapapa:
The 'Finish' link at the end should point to the 'General' settings tab, and not to the 'Upgrade' tab as discussed. We should add an upgrade section in the final step of the Wizard, but we can do that in a separate issue.

BrunoPavlinic98:
> I think the enhanced select should follow the new color scheme. Currently it looks like this:
> 
> Scherm­afbeelding 2025-01-31 om 16 24 44
> But I think it should look like this:
> 
> Scherm­afbeelding 2025-01-31 om 16 37 21
> For the selected elements you can use:
> 
> ```
> background-color: #6e1edc;
> border-bottom: 1px solid white;
> color: white;
> ```
> 
> And for the hover you can use:
> 
> ```
> background-color: #f1e9fc;
> color: #000;
> ```

@Terminator-Barbapapa Done! I have only removed the border bottom styling because it moved the elements 1px above on hover, which looked weird.

Terminator-Barbapapa (approved):

alexmigf (approved):

assets/css/setup-wizard.css
```diff
@@ -34,12 +31,17 @@ body.wpo-wcpdf-setup a:hover {
...
-	border-color: #239bb9!important;
-	box-shadow: 0 0 0 1px #239bb9!important;
+	border-color: #6e1edc!important;
+	box-shadow: 0 0 0 1px #6e1edc!important;
 	outline: 2px solid transparent!important;
 	color: black!important;
 }
 
+body.wpo-wcpdf-setup .select2-search__field:focus {
+	border-color: none!important;
```
MohamadNateqi, line 41: It should be `unset`. 
Also, a space before `!important` would be good, same for other occurance.

views/setup-wizard/attach-to.php
MohamadNateqi, line 20: Maybe it would be good to extract this into a separate variable to improve readability.

MohamadNateqi (approved):
LGTM!

PR #901, opened 2024-10-31
Fix: template item meta styling

### LINKED ISSUE #894: Add backward compatibility with the "meta" class + Adjust item meta spacing
This issue addresses two reports, related to the styles of the target items:

### Backward compatibility with the "meta" class

Some users (e.g. [this one](https://wordpress.org/support/topic/remove-sku-weight-from-packing-slipenlarge-remaining-meta-notes/)) are using code snippets that are targeting to the "meta" class, which no longer exists since PR #810 was merged.

I have tested myself, adding this class next to item meta, in the following line, and it worked:
https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/928561a281fafd5631b444730543257b9dea4d73/templates/Simple/invoice.php#L121

### Adjust item meta spacing

While testing the issue above, I realized that a margin is still being below the product name (also reported by a user [here](https://wordpress.org/support/topic/new-upgrade-causing-excessive-line-spacing/)), even when none item meta is present:

image

This is because we are adding a margin to the whole container, here:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/928561a281fafd5631b444730543257b9dea4d73/templates/Simple/style.css#L226-L232

We may need to decrease the margin a bit here as well to make it look more like the spacing we had before the changes applied in PR #810:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/928561a281fafd5631b444730543257b9dea4d73/templates/Simple/style.css#L249-L254

alexmigf:
closes #894

### DIFF
--- a/templates/Simple/style.css
+++ b/templates/Simple/style.css
@@ -224,8 +224,9 @@ table.notes-totals {
 }
 
 /* item meta formatting */
-.item-meta {
-	margin: 8px 0 0 0;
+.item-meta,
+.meta {
+	margin: 4px 0 0 0;
 	font-size: 7pt;
 	line-height: 7pt;
 	overflow-wrap: anywhere;
@@ -248,9 +249,9 @@ table.notes-totals {
 
 .wc-item-meta p,
 .wc-item-meta dd,
-.item-meta p {
-	margin-top: 4px;
-	margin-left: 0;
+.item-meta p,
+.meta p {
+	margin: 0;
 }
 
 /* Notes & Totals */
--- a/templates/Simple/invoice.php
+++ b/templates/Simple/invoice.php
@@ -118,18 +118,18 @@
 				<td class="product">
 					<p class="item-name"><?php echo $item['name']; ?></p>
 					<?php do_action( 'wpo_wcpdf_before_item_meta', $this->get_type(), $item, $this->order ); ?>
-					<div class="item-meta">
+					<div class="item-meta meta">
 						<?php if ( ! empty( $item['sku'] ) ) : ?>
 							<p class="sku"><span class="label"><?php $this->sku_title(); ?></span> <?php echo esc_attr( $item['sku'] ); ?></p>
 						<?php endif; ?>
 						<?php if ( ! empty( $item['weight'] ) ) : ?>
 							<p class="weight"><span class="label"><?php $this->weight_title(); ?></span> <?php echo esc_attr( $item['weight'] ); ?><?php echo esc_attr( get_option( 'woocommerce_weight_unit' ) ); ?></p>
 						<?php endif; ?>
-						<!-- ul.wc-item-meta -->
+						<!-- .wc-item-meta -->
 						<?php if ( ! empty( $item['meta'] ) ) : ?>
 							<?php echo $item['meta']; ?>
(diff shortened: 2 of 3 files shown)


templates/Simple/style.css
```diff
@@ -224,8 +224,9 @@ table.notes-totals {
 }
 
 /* item meta formatting */
-.item-meta {
-	margin: 8px 0 0 0;
+.item-meta,
+.meta {
```
MohamadNateqi, line 228: Could we have meta items without the `item-meta` class?

MohamadNateqi (approved):
LGTM!

templates/Simple/style.css
alexmigf: Can you explain better?

MohamadNateqi: Sure, sorry.
We already have the `.item-meta` selector, and you also added the `.meta`, which seems repetitive. I thought maybe we could have meta items that don't have the `.item-meta` class and only have `.meta`, so you added this selector also to support them.

YordanSoares (commented):
It looks better now, but I wonder if we could adjust the following rule from `0.75em` to `2pt`:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/0a1436d430511e9752c868f659f02de43e838bb3/templates/Simple/style.css#L47-L49

...and set the same value here for the `margin-top`:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/0a1436d430511e9752c868f659f02de43e838bb3/templates/Simple/style.css#L239-L242

I tried myself and it looks like this:

testing-pr-911

alexmigf:
@YordanSoares I don't notice any substantial difference:

Without your suggestion:

Screenshot from 2024-11-04 12-11-22

With your suggestion:

Screenshot from 2024-11-04 12-11-04

YordanSoares:
@alexmigf I do see differences. Here is another test with several items, to see the differences clearly:

testing-pr-911-2

alexmigf:
@MohamadNateqi can you replicate this?

MohamadNateqi:
I changed both to `2pt` as Yordan suggested, but it was a slight change for me:

image

YordanSoares:
Here's my system configuration:

```
### WordPress Environment ###

WordPress address (URL): https://pdf-invoices-pr-901.local
Site address (URL): https://pdf-invoices-pr-901.local
WC Version: 9.3.3
Legacy REST API Package Version: The Legacy REST API plugin is not installed on this site.
Action Scheduler Version: ✔ 3.8.1
Log Directory Writable: ✔
WP Version: 6.6.2

WP Memory Limit: 256 MB
WP Debug Mode: –
WP Cron: ✔
Language: en_US
External object cache: –

### Server Environment ###

Server Info: nginx/1.26.1
PHP Version: 8.2.23
PHP Post Max Size: 1,000 MB
PHP Time Limit: 1200
PHP Max Input Vars: 4000
cURL Version: 8.7.0-DEV
OpenSSL/3.0.13
```

YordanSoares (approved):
@alexmigf now it is looking great on my side! 🚀

PR #904, opened 2024-11-01
Fix zero value order items invoice lines

MohamadNateqi:
A customer reported an issue they saw when validating their XML invoice. 
When there is an order item with `0` as its total value, the `TaxSubtotal` details will not be displayed for that item, as shown below for the second `InvoiceLine`:

image


ubl/Handlers/Invoice/InvoiceLineHandler.php
```diff
@@ -23,9 +23,7 @@ public function handle( $data, $options = array() ) {
 			$line_tax_data    = $item[ $taxDataContainer ];
 
 			foreach ( $line_tax_data[ $taxDataKey ] as $tax_id => $tax ) {
-				if ( ! is_numeric( $tax ) ) {
-					continue;
-				}
+				$tax = absint( $tax );
```
alexmigf, line 26: Could you clarify the reason for this change? It seems the tax value might be a float string, but you're reassigning it as an integer. Additionally, zero is still a numeric value, so I'm curious about the intent behind this adjustment.

alexmigf (changes requested):

ubl/Handlers/Invoice/InvoiceLineHandler.php
MohamadNateqi: The `$tax` value was actually an empty string, and I wanted to get `0` if it's an empty string, but I forgot about float values. I'll update the code.

alexmigf: For me, it wasn't an empty string; it was `'0'`. Have you checked it again? Wouldn't it be helpful to know what the validation error was?

alexmigf (approved):
LGTM!

ubl/Handlers/Invoice/InvoiceLineHandler.php
alexmigf: I understand. The problem isn’t that the item’s value is zero; it’s that no tax has been applied to it.

PR #907, opened 2024-11-13
Upgrade links are not displayed

### LINKED ISSUE #903: Upgrade links are not displayed
I think since we updated the Upgrade tab, the Upgrade links don't show up as expected anymore. I just tested this with only the Pro extension active. And the Upgrade link for the bundle did not display.
I think we might need to adjust this logic? https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/928561a281fafd5631b444730543257b9dea4d73/views/upgrade-table.php#L81-L86

BrunoPavlinic98:
closes #903


BrunoPavlinic98:
@Terminator-Barbapapa Can you please test this?

Terminator-Barbapapa (changes requested):
I just tested this for the four possible scenario's:

### 1. Only the Professional extension is installed and activated:
**Column 1:**  Show Professional extension features + Display 'Currently installed'.
**Column 2:** Show Bundle features + Display 'Upgrade now' button with personal upgrade link.

### 2. Only the Premium Templates extension is installed and activated:
**Column 1:**  Show Premium Templates extension features + Display 'Currently installed'.
**Column 2:** Show Bundle features + Display 'Upgrade now' button with personal upgrade link.

### 3. Both the Professional  and Premium Templates extensions are installed and activated:
**Column 1:**  Show Professional extension features.
**Column 2:** Show Bundle features + Display 'Currently installed'.

### 4. Neither the Professional  and Premium Templates extensions are installed and activated:
**Column 1:**  Show Professional extension features + Display 'Buy now' button with a link to the product page (including UTM parameters).
**Column 2:** Show Bundle features + Display 'Buy now' button with a link to the product page (including UTM parameters).

Here is what I found:

- **Scenario 1:** The link for the Bundle goes to the product page instead of providing a personal upgrade link for the customer.
- **Scenario 2:** The link for the Bundle goes to the product page instead of providing a personal upgrade link for the customer.
- **Scenario 3:** The column with Professional features is not showing. The 'Upgrade now' button for the Bundle is still visible. This should become the 'Currently installed' message.
- **Scenario 4:** Correct.

BrunoPavlinic98:
@Terminator-Barbapapa Since the personal upgrade link consists of `edd_license_key` and `download_id`, I was wondering if `download_id` is hardcoded somewhere and if it is connected to particular products (plugins) we are offering?

BrunoPavlinic98:
@Terminator-Barbapapa I think I covered all the cases now. I have added some new piece of code that should be (in my opinion) easier to read and debug.

includes/Settings/SettingsUpgrade.php
```diff
@@ -257,6 +257,7 @@ public function get_extension_license_infos( $ignore_cache = false ) {
 				if ( ! empty( $license_key ) ) {
 					$args['edd_action']  = 'check_license';
 					$args['license_key'] = trim( $license_key );
+					$license_info[$extension]['license_key'] = $license_key;
```
alexmigf, line 260: This can be added to the line above:

```php
$args['license_key'] = $license_info[ $extension ]['license_key'] = trim( $license_key );
```

includes/Settings/SettingsUpgrade.php
```diff
@@ -304,12 +305,16 @@ public function get_extension_license_infos( $ignore_cache = false ) {
 			if ( ! empty( $bundle_upgrade_link ) && $license_status == 'valid' ) {
 				$license_info[$extension]['url'] = $bundle_upgrade_link;
 			} else {
+				$license = ! empty( $license_info[$extension]['license_key'] ) ? $license_info[$extension]['license_key'] : ( ! empty( $license_info['pro']['license_key'] ) ? $license_info['pro']['license_key'] : $license_info['templates']['license_key'] );
 				switch ( $extension ) {
 					case 'pro':
-						$license_info[$extension]['url'] = 'https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional?utm_medium=plugin&utm_source=ips&utm_campaign=upgrade-tab&content=ips-pro-upgrade';
+						$license_info[$extension]['url'] = "https://wpovernight.com/checkout/?edd_license_key={$license}";
```
alexmigf, line 311: If all three options share the same URL, the `switch` statement is no longer necessary.

includes/Settings/SettingsUpgrade.php
```diff
@@ -304,12 +305,16 @@ public function get_extension_license_infos( $ignore_cache = false ) {
 			if ( ! empty( $bundle_upgrade_link ) && $license_status == 'valid' ) {
 				$license_info[$extension]['url'] = $bundle_upgrade_link;
 			} else {
+				$license = ! empty( $license_info[$extension]['license_key'] ) ? $license_info[$extension]['license_key'] : ( ! empty( $license_info['pro']['license_key'] ) ? $license_info['pro']['license_key'] : $license_info['templates']['license_key'] );
```
alexmigf, line 308: Maybe you don't need this line?

alexmigf (changes requested):

includes/Settings/SettingsUpgrade.php
alexmigf: If the license is empty, wouldn't it be better to provide the product link instead?

BrunoPavlinic98: @alexmigf It makes sense, but do you think we should redirect the user to the PRO plugin product link or to the bundle?
The PRO is the next step when it comes to upgrading from free, but bundle also makes sense because it is a better deal.

alexmigf: I believe the Professional should link to its respective product, and the Bundle should link to itself as well. Linking both to the Bundle might be problematic—it could give customers the impression that we're trying to mislead them.

Terminator-Barbapapa (approved):
LGTM! 🚀

PR #910, opened 2024-11-20
New: raised the minimum PHP version requirement to 7.4

### LINKED ISSUE #889: Bump PHP min version to `7.4`
We plan to upgrade the `sabre/xml` library, which requires a minimum of PHP `7.4`. Since WordPress has already discontinued support for PHP `7.1`, it's likely that `7.2` will follow soon. It might be a good time to consider raising our minimum PHP version requirement to `7.4`.

alexmigf:
closes #889

### DIFF
--- a/includes/Install.php
+++ b/includes/Install.php
@@ -82,7 +82,7 @@ class Install {
 	 */
 	protected function install() {
 		// only install when php version or higher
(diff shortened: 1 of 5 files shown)


MohamadNateqi (approved):
LGTM!

PR #913, opened 2024-12-02
Add user info to order note

### LINKED ISSUE #864: Add user info when log document creation
For the "Log to order notes", the user info could be added.

MohamadNateqi:
close #864 
This PR adds the user display name and ID to the order note for invoice creation and deletion:

image

image

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1270,7 +1270,15 @@ class Main {
 	public function log_document_creation_to_order_notes( $document, $trigger ) {
 		$triggers = $this->get_document_triggers();
 		if ( ! empty( $document ) && isset( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) && ! empty( $trigger ) && array_key_exists( $trigger, $triggers ) ) {
-			$user    = wp_get_current_user();
+			$user = wp_get_current_user();
+
+			if ( empty( $user ) ) {
+				$user = (object) array(
+					'display_name' => 'System',
+					'ID'           => 0,
+				);
+			}
+
 			/* translators: 1. document title, 2. creation trigger, 3. user display name, 4. user ID */
 			$message = __( 'PDF %1$s created via %2$s by %3$s(#%4$s).', 'woocommerce-pdf-invoices-packing-slips' );
 			$note    = sprintf( $message, $document->get_title(), $triggers[ $trigger ], $user->display_name, $user->ID );
@@ -1287,7 +1295,15 @@ class Main {
 	 */
 	public function log_document_deletion_to_order_notes( object $document ): void {
 		if ( ! empty( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) ) {
-			$user    = wp_get_current_user();
+			$user = wp_get_current_user();
+
+			if ( empty( $user ) ) {
+				$user = (object) array(
+					'display_name' => 'System',
+					'ID'           => 0,
+				);
+			}
+
 			/* translators: 1. document title 2. user display name, 3. user ID */
 			$message = __( 'PDF %1$s deleted by %2$s(#%3$s).', 'woocommerce-pdf-invoices-packing-slips' );
 			$note    = sprintf( $message, $document->get_title(), $user->display_name, $user->ID );
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1270,9 +1270,10 @@ class Main {
 	public function log_document_creation_to_order_notes( $document, $trigger ) {
 		$triggers = $this->get_document_triggers();
 		if ( ! empty( $document ) && isset( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) && ! empty( $trigger ) && array_key_exists( $trigger, $triggers ) ) {
+			$user    = wp_get_current_user();
 			/* translators: 1. document title, 2. creation trigger */
-			$message = __( 'PDF %1$s created via %2$s.', 'woocommerce-pdf-invoices-packing-slips' );
-			$note    = sprintf( $message, $document->get_title(), $triggers[$trigger] );
+			$message = __( 'PDF %1$s created via %2$s by %3$s(#%4$s).', 'woocommerce-pdf-invoices-packing-slips' );
+			$note    = sprintf( $message, $document->get_title(), $triggers[ $trigger ], $user->display_name, $user->ID );
 			$this->log_to_order_notes( $note, $document );
 		}
 	}
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1280,7 +1280,7 @@ class Main {
 		}
 
 		$user_note       = '';
-		$manual_triggers = $this->get_document_triggers( true );
+		$manual_triggers = $this->get_document_triggers( 'manual' );
 
 		// Add user information if the trigger is manual.
 		if ( array_key_exists( $trigger, $manual_triggers ) ) {
@@ -1447,25 +1447,36 @@ class Main {
 	/**
 	 * Get the document triggers
 	 *
-	 * @param bool $only_manual_triggers
+	 * @param string $trigger_type
 	 *
 	 * @return array
 	 */
-	public function get_document_triggers( bool $only_manual_triggers = false ): array {
-		$triggers = array(
-			'single'           => __( 'single order action', 'woocommerce-pdf-invoices-packing-slips' ),
-			'bulk'             => __( 'bulk order action', 'woocommerce-pdf-invoices-packing-slips' ),
-			'my_account'       => __( 'my account', 'woocommerce-pdf-invoices-packing-slips' ),
-			'email_attachment' => __( 'email attachment', 'woocommerce-pdf-invoices-packing-slips' ),
-			'document_data'    => __( 'order document data (number and/or date set manually)', 'woocommerce-pdf-invoices-packing-slips' ),
+	public function get_document_triggers( string $trigger_type = 'all' ): array {
+		$manual_triggers = array(
+			'single'        => __( 'single order action', 'woocommerce-pdf-invoices-packing-slips' ),
+			'bulk'          => __( 'bulk order action', 'woocommerce-pdf-invoices-packing-slips' ),
+			'my_account'    => __( 'my account', 'woocommerce-pdf-invoices-packing-slips' ),
+			'document_data' => __( 'order document data (number and/or date set manually)', 'woocommerce-pdf-invoices-packing-slips' ),
 		);
 
-		if ( $only_manual_triggers ) {
-			$user_triggers = array( 'single', 'bulk', 'my_account', 'document_data' );
-			$triggers      = array_intersect_key( $triggers, array_flip( $user_triggers ) );
+		$automatic_triggers = array(
+			'email_attachment' => __( 'email attachment', 'woocommerce-pdf-invoices-packing-slips' ),
+		);
+
+		switch ( $trigger_type ) {
+			case 'manual':
+				$triggers = $manual_triggers;
+				break;
+			case 'automatic':
+				$triggers = $automatic_triggers;
+				break;
+			case 'all':
+			default:
+				$triggers = array_merge( $manual_triggers, $automatic_triggers );
+				break;
 		}
 
-		return apply_filters( 'wpo_wcpdf_document_triggers', $triggers, $only_manual_triggers );
+		return apply_filters( 'wpo_wcpdf_document_triggers', $triggers, $trigger_type );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/Main.php
```diff
@@ -1270,9 +1270,10 @@ public function set_phpmailer_validator( $mailArray ) {
 	public function log_document_creation_to_order_notes( $document, $trigger ) {
 		$triggers = $this->get_document_triggers();
 		if ( ! empty( $document ) && isset( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) && ! empty( $trigger ) && array_key_exists( $trigger, $triggers ) ) {
-			/* translators: 1. document title, 2. creation trigger */
-			$message = __( 'PDF %1$s created via %2$s.', 'woocommerce-pdf-invoices-packing-slips' );
-			$note    = sprintf( $message, $document->get_title(), $triggers[$trigger] );
+			$user    = wp_get_current_user();
+			/* translators: 1. document title, 2. creation trigger, 3. user display name, 4. user ID */
+			$message = __( 'PDF %1$s created via %2$s by %3$s(#%4$s).', 'woocommerce-pdf-invoices-packing-slips' );
```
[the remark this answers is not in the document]
MohamadNateqi: I don't see any order notes for automatic workflow! It only adds notes on manual processes.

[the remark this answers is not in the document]
alexmigf: Apologies for not mentioning this earlier; it's utilized in the Pro version here: https://github.com/wpovernight/woocommerce-pdf-ips-pro/blob/968f66fa24a711700ad1a5524306410cd01e82a6/includes/wcpdf-pro-functions.php#L1243-L1275

alexmigf (changes requested):

includes/Main.php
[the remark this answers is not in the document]
alexmigf: Apologies for not mentioning this earlier; it's utilized in the Pro version here: https://github.com/wpovernight/woocommerce-pdf-ips-pro/blob/main/includes/wcpdf-pro-functions.php#L1243-L1275

MohamadNateqi:
@alexmigf What do you think about the last commit?

includes/Main.php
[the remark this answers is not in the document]
MohamadNateqi: I was not getting order notes until I checked the "archive PDF" setting. I think the log code you mentioned in the Pro extension should be out of that condition. What do you think?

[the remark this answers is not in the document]
alexmigf: You likely just need to catalog the user triggers and use the user object whenever one of those triggers occurs.

alexmigf (changes requested):

MohamadNateqi:
@alexmigf Did you have this idea in mind that is implemented in the last commit?

alexmigf:
> Did you have this idea in mind that is implemented in the last commit?

@MohamadNateqi Could you clarify? I want to make sure I understood correctly.

includes/Main.php
```diff
@@ -1265,16 +1265,41 @@ public function set_phpmailer_validator( $mailArray ) {
...
+
+		// Add user information if the trigger is manual.
+		if ( array_key_exists( $trigger, $manual_triggers ) ) {
+			$user = wp_get_current_user();
+
+			if ( ! empty( $user->display_name ) && ! empty( $user->ID )   ) {
+				$user_note_message = sprintf(
+					/* translators: 1. user display name, 2. user ID */
+					__( ' by %1$s(#%2$s)', 'woocommerce-pdf-invoices-packing-slips' ),
+					$user->display_name,
```
alexmigf, line 1293: Maybe this needs to be escaped.

includes/Main.php
```diff
@@ -1265,16 +1265,41 @@ public function set_phpmailer_validator( $mailArray ) {
...
+					$user->ID
```
alexmigf, line 1294: Same.

includes/Main.php
```diff
@@ -1265,16 +1265,41 @@ public function set_phpmailer_validator( $mailArray ) {
...
+				);
+			}
 		}
+
+		/* translators: 1. document title, 2. creation trigger, 3. user info */
+		$message = __( 'PDF %1$s created via %2$s%3$s.', 'woocommerce-pdf-invoices-packing-slips' );
```
alexmigf, line 1300: Could you consider removing the last placeholder? It seems unnecessary and might not aid in understanding the translation—you could concatenate it at the end instead.

includes/Main.php
```diff
@@ -1286,9 +1311,10 @@ public function log_document_creation_to_order_notes( $document, $trigger ) {
 	 */
 	public function log_document_deletion_to_order_notes( object $document ): void {
```
alexmigf, line 1312: Maybe we can make this more modular and pass the trigger too?

alexmigf (changes requested):

includes/Main.php
MohamadNateqi: The issue is the period at the end of the sentence that made me put that placeholder, but I made some changes in the structure.

includes/Main.php
MohamadNateqi: In this case, we need to add a `$trigger` parameter to the `delete()` method to pass it to the log method. I'm not sure if that's a good idea. What do you think?

alexmigf: Got it. Let’s leave it as it is for now.

includes/Main.php
```diff
@@ -1286,20 +1286,19 @@ public function log_document_creation_to_order_notes( object $document, string $
...
 
-			if ( ! empty( $user->display_name ) && ! empty( $user->ID )   ) {
+			if ( ! empty( $user->user_login ) ) {
 				$user_note_message = sprintf(
-					/* translators: 1. user display name, 2. user ID */
-					__( ' by %1$s(#%2$s)', 'woocommerce-pdf-invoices-packing-slips' ),
-					$user->display_name,
-					$user->ID
+					/* translators: user login name */
+					__( ' (User: %s)', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 1292: The string should not include a leading space here.

includes/Main.php
```diff
@@ -1286,20 +1286,19 @@ public function log_document_creation_to_order_notes( object $document, string $
...
 			}
 		}
 
-		/* translators: 1. document title, 2. creation trigger, 3. user info */
-		$message = __( 'PDF %1$s created via %2$s%3$s.', 'woocommerce-pdf-invoices-packing-slips' );
-		$note    = sprintf( $message, $document->get_title(), $triggers[ $trigger ], $user_note_message );
-		$this->log_to_order_notes( $note, $document );
+		/* translators: 1. document title, 2. creation trigger */
+		$message = __( 'PDF %1$s created via %2$s.', 'woocommerce-pdf-invoices-packing-slips' );
+		$note    = sprintf( $message, $document->get_title(), $triggers[ $trigger ] );
```
alexmigf, line 1300: Why not include the string directly here, as we usually do?

includes/Main.php
```diff
@@ -1311,11 +1310,21 @@ public function log_document_creation_to_order_notes( object $document, string $
...
 	public function log_document_deletion_to_order_notes( object $document ): void {
 		if ( ! empty( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) ) {
-			$user = wp_get_current_user();
+			$user_note_message = '';
+			$user              = wp_get_current_user();
+
+			if ( ! empty( $user->user_login ) ) {
+				$user_note_message = sprintf(
+					/* translators: user login name */
+					__( ' (User: %s)', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 1319: Same.

includes/Main.php
```diff
@@ -1311,11 +1310,21 @@ public function log_document_creation_to_order_notes( object $document, string $
...
+					esc_html( $user->user_login )
+				);
+			}
+
 			/* translators: 1. document title 2. user display name, 3. user ID */
-			$message = __( 'PDF %1$s deleted by %2$s(#%3$s).', 'woocommerce-pdf-invoices-packing-slips' );
+			$message = __( 'PDF %1$s deleted.', 'woocommerce-pdf-invoices-packing-slips' );
 			$note    = sprintf( $message, $document->get_title(), $user->display_name, $user->ID );
```
alexmigf, line 1326: Same.

alexmigf (changes requested):

includes/Main.php
```diff
@@ -1279,26 +1279,30 @@ public function log_document_creation_to_order_notes( object $document, string $
...
-		$note    = sprintf( $message, $document->get_title(), $triggers[ $trigger ] );
-		$this->log_to_order_notes( $note . $user_note_message, $document );
+		$note = sprintf(
+			/* translators: 1. document title, 2. creation trigger */
+			__( 'PDF %1$s created via %2$s.', 'woocommerce-pdf-invoices-packing-slips' ),
+			$document->get_title(),
+			$triggers[ $trigger ]
+		);
+
+		$this->log_to_order_notes( $note . ' ' . $user_note, $document );
```
alexmigf, line 1305: Maybe you should concatenate only if the `$user_note` is not empty.

includes/Main.php
```diff
@@ -1310,21 +1314,24 @@ public function log_document_creation_to_order_notes( object $document, string $
...
-			$message = __( 'PDF %1$s deleted.', 'woocommerce-pdf-invoices-packing-slips' );
-			$note    = sprintf( $message, $document->get_title(), $user->display_name, $user->ID );
-			$this->log_to_order_notes( $note . $user_note_message, $document );
+			$note = sprintf(
+				/* translators: document title  */
+				__( 'PDF %s deleted.', 'woocommerce-pdf-invoices-packing-slips' ),
+				$document->get_title()
+			);
+
+			$this->log_to_order_notes( $note . ' ' . $user_note, $document );
```
alexmigf, line 1334: Same.

alexmigf (changes requested):

includes/Main.php
MohamadNateqi: Sorry missed that.

includes/Main.php
```diff
@@ -1287,7 +1287,7 @@ public function log_document_creation_to_order_notes( object $document, string $
 			$user = wp_get_current_user();
 
 			if ( ! empty( $user->user_login ) ) {
-				$user_note = sprintf(
+				$user_note = ' ' . sprintf(
 					/* translators: user login name */
 					__( '(User: %s)', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 1292: Perhaps you could also consider removing the brackets from the string?

alexmigf (changes requested):

includes/Main.php
MohamadNateqi: Would it be good like this:

image

Or maybe this is better?

image

alexmigf: Either approach works for me; I just wanted to suggest that we might be able to skip passing them to the translation function.

includes/Main.php
```diff
@@ -1405,16 +1447,25 @@ public function log_document_creation_trigger_to_order_meta( $document, $trigger
...
 	 * Get the document triggers
 	 *
+	 * @param bool $only_manual_triggers
+	 *
 	 * @return array
 	 */
-	public function get_document_triggers() {
-		return apply_filters( 'wpo_wcpdf_document_triggers', [
+	public function get_document_triggers( bool $only_manual_triggers = false ): array {
+		$triggers = array(
```
alexmigf, line 1455: Consider improving this by separating the manual and automatic triggers, keeping them distinct and only combining them when necessary. Additionally, introducing separate filters for each group could enhance clarity and functionality.

alexmigf (changes requested):

includes/Main.php
alexmigf: The function parameter could be enhanced by allowing it to accept values like 'manual', 'automatic', or 'all'.

MohamadNateqi: Since changing the filter could make Pro modifications not work, I didn't change it, but let me know if you have any ideas!

alexmigf: Actually I was suggesting adding new filters for the groups, and not changing the one in the return.

MohamadNateqi: Got it, done!

includes/Main.php
```diff
@@ -1447,25 +1447,36 @@ public function log_document_creation_trigger_to_order_meta( $document, $trigger
 	/**
 	 * Get the document triggers
 	 *
-	 * @param bool $only_manual_triggers
+	 * @param string $trigger_type
```
alexmigf, line 1450: Would it be helpful to add a brief description here outlining the available options?

MohamadNateqi: Nice idea! Thanks

alexmigf (approved):
LGTM!

PR #914, opened 2024-12-04
Fix requirement admin notice issue

MohamadNateqi:
The admin notice for server requirements was displayed on `admin_init` instead of `admin_notices`.

### DIFF
--- a/includes/Settings/SettingsDebug.php
+++ b/includes/Settings/SettingsDebug.php
@@ -1113,7 +1113,7 @@ class SettingsDebug {
 	 *
 	 * @return void
 	 */
-	public function display_server_requirement_notice() {
+	public function display_server_requirement_notice(): void {
 		$status_page_url = admin_url( 'admin.php?page=wpo_wcpdf_options_page&tab=debug&section=status' );
 		$dismiss_url     = wp_nonce_url( add_query_arg( 'wpo_dismiss_requirements_notice', true ), 'dismiss_requirements_notice' );
 		$notice_message  = sprintf(


includes/Settings/SettingsDebug.php
```diff
@@ -1105,24 +1105,26 @@ public function display_server_requirement_notice(): void {
...
-		$status_page_url = admin_url( 'admin.php?page=wpo_wcpdf_options_page&tab=debug&section=status' );
-		$dismiss_url     = wp_nonce_url( add_query_arg( 'wpo_dismiss_requirements_notice', true ), 'dismiss_requirements_notice' );
-		$notice_message  = sprintf(
-			/* translators: 1: Plugin name, 2: Open anchor tag, 3: Close anchor tag */
-			__( 'Your server does not meet the requirements for %1$s. Please check the %2$sStatus page%3$s for more information.', 'woocommerce-pdf-invoices-packing-slips' ),
-			'<strong>PDF Invoices & Packing Slips for WooCommerce</strong>',
-			'<a href="' . esc_url( $status_page_url ) . '">',
-			'</a>'
-		);
+		add_action( 'admin_notices', function () {
```
alexmigf, line 1108: Consider moving the code into a well-defined function for better structure and maintainability.

alexmigf (changes requested):

MohamadNateqi:
@alexmigf I was considering that using options to store the dismissal decision might not be the best approach. Since it's constant, if the server configuration changes later on, the notice won't be displayed anymore. Therefore, it might be better to use a transient instead. What do you think?

alexmigf:
> I was considering that using options to store the dismissal decision might not be the best approach. Since it's constant, if the server configuration changes later on, the notice won't be displayed anymore. Therefore, it might be better to use a transient instead. What do you think?

It depends on the option, as some might be permanent. Perhaps it's best to open a new issue to discuss this further.

includes/Settings/SettingsDebug.php
```diff
@@ -1066,7 +1066,7 @@ public function get_server_config(): array {
 	 *
 	 * @return void
 	 */
-	public function display_server_requirement_notice(): void {
+	public function handle_server_requirement_notice(): void {
```
alexmigf, line 1069: Perhaps we don’t need this function at all. Instead, we could integrate the check directly into the notice.

MohamadNateqi: Wouldn't it be better to separate the logic from the function that outputs the HTML? Also, this function is called on `admin_init`.

alexmigf: We’ve already implemented similar checks for other notices. For example, take a look at this one: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/406406fe3d8c305ec059bb4573c19411a0eb1c6a/woocommerce-pdf-invoices-packingslips.php#L424-L454. I think the current approach is perfectly fine—it’s just a simple conditional check.

alexmigf: I apologize for the confusion. After taking another look, I realize the previous implementation was actually better. I initially overlooked the remaining checking code since it was hidden on the review page.

MohamadNateqi: No problem. I'll revert it now.

alexmigf (approved):
LGTM!

PR #921, opened 2024-12-10
Add shop phone number field for e-Invoice support

### LINKED ISSUE #920: Add a Store Phone Field to General Settings
To enhance the plugin's support for e-invoicing, we need to implement a store phone field in the General Settings, as WooCommerce does not currently offer this functionality.

MohamadNateqi:
close #920


includes/Settings/SettingsGeneral.php
```diff
@@ -197,6 +197,19 @@ public function init_settings() {
...
+				'type'     => 'setting',
+				'id'       => 'shop_phone_number',
+				'title'    => __( 'Shop Phone Number', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'text_input',
+				'section'  => 'general_settings',
+				'args'     => array(
+					'option_name'  => $option_name,
+					'id'           => 'shop_phone_number',
+					'translatable' => true,
+					'description'  => __( 'Required for e-Invoice.', 'woocommerce-pdf-invoices-packing-slips' ) . ' ' . $requires_pro        ,
```
alexmigf, line 210: Let's skip adding a description for now and comment out the code instead.

includes/Settings/SettingsGeneral.php
```diff
@@ -197,6 +197,19 @@ public function init_settings() {
 					'description'  => __( 'Required for UBL output format.<br>You can display this number on the invoice from the document settings.', 'woocommerce-pdf-invoices-packing-slips' ) . ' ' . $requires_pro        ,
 				)
 			),
+			array(
+				'type'     => 'setting',
+				'id'       => 'shop_phone_number',
```
alexmigf, line 202: Maybe without the `shop_` prefix like the others?

includes/Settings/SettingsGeneral.php
```diff
@@ -197,6 +197,19 @@ public function init_settings() {
 					'description'  => __( 'Required for UBL output format.<br>You can display this number on the invoice from the document settings.', 'woocommerce-pdf-invoices-packing-slips' ) . ' ' . $requires_pro        ,
```
alexmigf, line 197: Something is wrong with the commas here.

includes/Settings/SettingsGeneral.php
alexmigf: Will not require Pro at this stage.

alexmigf (changes requested):

includes/Settings/SettingsGeneral.php
MohamadNateqi: I actually wanted to make it similar to the `shop_address`, but without the prefix, it would also be good.

alexmigf: I see. I agree with you then.

MohamadNateqi: Done.

alexmigf (commented):
Maybe you should create functions to retrieve the phone number like we do for other data: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/60a1a00bc1bf865cd69a36cf8aff33b150a5aa36/includes/Documents/OrderDocument.php#L1222-L1230

alexmigf (approved):
LGTM!

PR #923, opened 2024-12-10
New: renames maker class and temp path from UBL to XML

### LINKED ISSUE #922: Rename some stuff from "ubl" to "xml" to support other XML formats
As far I can tell:

* UBLMaker => XMLMaker
* Temp dir: ubl => xml

To be continue...

alexmigf:
closes #922


alexmigf:
I’ve decided to take a different approach, so this is no longer needed. Closing.

PR #927, opened 2024-12-13
New: adds support for multiple UBL formats

alexmigf:


ubl/Documents/UblDocument.php
```diff
@@ -9,6 +9,10 @@
 }
 
 class UblDocument extends Document {
+	
+	public function get_root_element() {
+		return apply_filters( 'wpo_wc_ubl_document_root_element', 'Invoice', $this );
+	}
```
MohamadNateqi, line 15: I think the abstract method should be declared in the parent class.

assets/js/admin-script.js
```diff
@@ -55,6 +55,17 @@ jQuery( function( $ ) {
...
+	// disable encrypted pdf option for non UBL 2.1 formats
+	$( "[name='wpo_wcpdf_documents_settings_invoice_ubl[ubl_format]']" ).on( 'change', function( event ) {
+		let $encryptedPdfCheckbox = $( this ).closest( 'form' ).find( "[name='wpo_wcpdf_documents_settings_invoice_ubl[include_encrypted_pdf]']" );
+		
+		if ( $( this ).val() !== 'ubl_2_1' ) {
+			$encryptedPdfCheckbox.prop( 'checked', false ).prop( 'disabled', true );
+		} else {
+			$encryptedPdfCheckbox.prop( 'disabled', false );
+		}
+	} ).trigger( 'change' );
```
MohamadNateqi, line 68: Maybe it's good to display text under the "Include encrypted PDF:" option to explain why the option is disabled. It might be confusing for users why it is disabled.

MohamadNateqi (changes requested):

MohamadNateqi (approved):
LGTM!

PR #928, opened 2024-12-13
Fix: AJAX preview loading when disabled on settings pages

### LINKED ISSUE #735: Prevent Preview from being called in certain settings tabs
Currently the PDF preview is requested in every settings tab, we should create a way to prevent that on certain tabs, for example from Pro extension tabs.

alexmigf:
closes #735


assets/js/admin-script.js
```diff
@@ -441,6 +448,7 @@ jQuery( function( $ ) {
 
 	// Load the Preview with AJAX
 	function ajaxLoadPreview() {
+		console.log( 'Loading preview...' );
```
MohamadNateqi, line 451: Would you like to keep it?

MohamadNateqi (approved):
LGTM!

assets/js/admin-script.js
alexmigf: Yes, it was intentional.

PR #933, opened 2024-12-17
Replace `function_exists` with `extension_loaded`

### LINKED ISSUE #932: Fix checking extension load
For some of the requirements displayed on the status page, we check the existence of functions used in the extension to see if they are loaded. We need to use `extension_loaded` for that.

MohamadNateqi:
close #932

### DIFF
--- a/includes/Settings/SettingsDebug.php
+++ b/includes/Settings/SettingsDebug.php
@@ -1012,7 +1012,7 @@ class SettingsDebug {
 			),
 			'WP Memory Limit' => array(
 				/* translators: <a> tags */
-				'required' => sprintf( __( 'Recommended: 128MB (more for plugins with heavy setups) See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
+				'required' => sprintf( __( 'Recommended: 128MB (more for plugins with heavy setups)<br/>See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
 				'value'    => sprintf( 'WordPress: %s, PHP: %s', WP_MEMORY_LIMIT, $php_mem_limit ),
 				'result'   => $memory_limit > 67108864,
 			),
--- a/includes/Settings/SettingsDebug.php
+++ b/includes/Settings/SettingsDebug.php
@@ -1094,7 +1094,7 @@ class SettingsDebug {
 		$server_configs          = $this->get_server_config();
 
 		foreach ( $server_configs as $config_name => $config ) {
-			if ( in_array( $config_name, array( 'opcache', 'GMagick or IMagick' ), true ) ) {
+			if ( in_array( $config_name, array( 'opcache', 'GMagick or IMagick', 'WP Memory Limit' ), true ) ) {
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


MohamadNateqi:
> Perhaps we should also ignore the memory in the notice since it's recommended, not required.

We had users reached out to us because of a memory limit issue. Would it be better to consider it for notification?

includes/Settings/SettingsDebug.php
```diff
@@ -1007,7 +1012,7 @@ public function get_server_config(): array {
 			),
 			'WP Memory Limit' => array(
 				/* translators: <a> tags */
-				'required' => sprintf( __( 'Recommended: 128MB (more for plugin-heavy setups<br/>See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
+				'required' => sprintf( __( 'Recommended: 128MB (more for plugins with heavy setups)<br/>See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
```
alexmigf, line 1015: You might also want to consider removing the `<br/>` HTML tag from the string while you're at it.

alexmigf:
> We had users reached out to us because of a memory limit issue. Would it be better to consider it for notification?

This generally occurs during bulk exports or when generating PDFs with large images. For regular usage, it’s not usually a concern. Since memory usage can vary greatly between setups, notifying customers might create unnecessary alerts and lead to an increase in support tickets.

includes/Settings/SettingsDebug.php
```diff
@@ -1007,7 +1012,7 @@ public function get_server_config(): array {
 			),
 			'WP Memory Limit' => array(
 				/* translators: <a> tags */
-				'required' => sprintf( __( 'Recommended: 128MB (more for plugin-heavy setups<br/>See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
+				'required' => sprintf( __( 'Recommended: 128MB (more for plugins with heavy setups) See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
```
alexmigf, line 1015: What I meant was to remove it from the string because it’s something that doesn’t need to be translated. Just place it outside the translation function. Right now, it looks strange.

Screenshot from 2024-12-17 08-23-16

alexmigf (changes requested):

includes/Settings/SettingsDebug.php
MohamadNateqi: Sorry, I misunderstood your point. Fixed.

includes/Settings/SettingsDebug.php
```diff
@@ -1007,7 +1012,12 @@ public function get_server_config(): array {
 			),
 			'WP Memory Limit' => array(
 				/* translators: <a> tags */
-				'required' => sprintf( __( 'Recommended: 128MB (more for plugin-heavy setups<br/>See: %1$sIncreasing the WordPress Memory Limit%2$s', 'woocommerce-pdf-invoices-packing-slips' ), '<a href="https://docs.woocommerce.com/document/increasing-the-wordpress-memory-limit/" target="_blank">', '</a>' ),
+				'required' => __( 'Recommended: 128MB (more for plugins with heavy setups)', 'woocommerce-pdf-invoices-packing-slips' ) . '<br/>' . sprintf(
```
alexmigf, line 1015: This sentence doesn't seem quite right. How about this instead?

> Recommended: 128MB (especially for plugin-heavy setups)

MohamadNateqi: Good suggestion.
I actually wanted to change it to `(more for websites with heavy setups)` but I for missed replacing plugins with websites.

alexmigf (approved):
LGTM!

PR #942, opened 2024-12-20
New: Displays a notice if the yearly reset action is not scheduled

### LINKED ISSUE #941: Display notice if yearly reset action not set
Display a notice if the yearly reset action is not scheduled within the last 30 days of the year.

alexmigf:
closes #941

Screenshot from 2024-12-20 16-45-51

### DIFF
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -490,8 +490,7 @@ class WPO_WCPDF {
 				wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_wcpdf_schedule_yearly_reset_action' );
 			} else {
 				$this->settings->schedule_yearly_reset_numbers();
-				$message = esc_html__( 'Yearly reset numbering system rescheduled!', 'woocommerce-pdf-invoices-packing-slips' );
-				wcpdf_log_error( $message, 'info' );
+				wcpdf_log_error( 'Yearly reset numbering system rescheduled!', 'info' );
 			}
 			
 			wp_redirect( 'admin.php?page=wpo_wcpdf_options_page&tab=debug&section=status' );
--- a/woocommerce-pdf-invoices-packingslips.php
+++ b/woocommerce-pdf-invoices-packingslips.php
@@ -76,6 +76,7 @@ class WPO_WCPDF {
 		add_action( 'admin_notices', array( $this, 'nginx_detected' ) );
 		add_action( 'admin_notices', array( $this, 'mailpoet_mta_detected' ) );
 		add_action( 'admin_notices', array( $this, 'rtl_detected' ) );
+		add_action( 'admin_notices', array( $this, 'yearly_reset_action_missing_notice' ) );
 		add_action( 'admin_notices', array( $this, 'legacy_addon_notices' ) );
 		add_action( 'init', array( '\\WPO\\IPS\\Semaphore', 'init_cleanup' ), 999 ); // wait AS to initialize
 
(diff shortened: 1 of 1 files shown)


woocommerce-pdf-invoices-packingslips.php
```diff
@@ -452,6 +453,51 @@ public function rtl_detected() {
...
+			return;
+		}
+		
+		$current_date   = new \DateTime();
+		$end_of_year    = new \DateTime( 'last day of December' );
+		$days_remaining = $current_date->diff( $end_of_year )->days;
+
+		// Check if the current date is within the last 30 days of the year
+		if ( $days_remaining <= 30 && ! $this->settings->yearly_reset_action_is_scheduled() ) {
+			ob_start();
```
MohamadNateqi, line 477: Why using buffer and not just using echo?

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -452,6 +453,51 @@ public function rtl_detected() {
...
+			?>
+			<div class="notice notice-error">
+				<p><?php esc_html_e( "The year-end is approaching, and we noticed that your PDF Invoices & Packing Slips for WooCommerce plugin doesn't have the scheduled action to reset invoice numbers annually, even though you've explicitly enabled this setting in the document options. Click the button below to schedule the action before the year ends.", 'woocommerce-pdf-invoices-packing-slips' ); ?></p>
+				<p><a class="button" href="<?php echo esc_url( wp_nonce_url( add_query_arg( 'wpo_wcpdf_schedule_yearly_reset_action', 'true' ), 'schedule_yearly_reset_action_nonce' ) ); ?>"><?php esc_html_e( 'Schedule the action now', 'woocommerce-pdf-invoices-packing-slips' ); ?></a></p>
```
MohamadNateqi, line 481: Maybe also adding a dimiss button would be good?

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -452,6 +453,51 @@ public function rtl_detected() {
...
+			echo wp_kses_post( ob_get_clean() );
+		}
+		
+		// Schedule yearly reset action
+		if ( isset( $_REQUEST['wpo_wcpdf_schedule_yearly_reset_action'] ) && isset( $_REQUEST['_wpnonce'] ) ) {
+			if ( ! wp_verify_nonce( sanitize_text_field( wp_unslash( $_REQUEST['_wpnonce'] ) ), 'schedule_yearly_reset_action_nonce' ) ) {
+				wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_wcpdf_schedule_yearly_reset_action' );
+			} else {
+				$this->settings->schedule_yearly_reset_numbers();
+				$message = esc_html__( 'Yearly reset numbering system rescheduled!', 'woocommerce-pdf-invoices-packing-slips' );
```
MohamadNateqi, line 493: Does it need to be translatable?

woocommerce-pdf-invoices-packingslips.php
alexmigf: I think all notices are using it, we could improve in a new issue.

woocommerce-pdf-invoices-packingslips.php
alexmigf: I prefer not to, as it could be clicked accidentally, either by the admin or a collaborator. Instead, clicking the button should dismiss the notice while still allowing the issue to be resolved.

woocommerce-pdf-invoices-packingslips.php
alexmigf: Since this notice will be permanent, it might be helpful to make it translatable.

MohamadNateqi: But isn't the `$message` only for logs?

MohamadNateqi (approved):
LGTM!

woocommerce-pdf-invoices-packingslips.php
alexmigf: Apologies, I thought you were referring to the notice message. I'll go ahead and remove it.

PR #945, opened 2024-12-24
Fix: Incorrect treatment of UBL format setting as historical

### LINKED ISSUE #944: The UBL format is treated as a historical configuration
A customer reported an issue with the UBL XRechnung format, which consistently output UBL 2.1. The problem stemmed from the format setting being incorrectly treated as historical when it should not have been.

alexmigf:
closes #944


MohamadNateqi (approved):
LGTM!

PR #953, opened 2024-12-27
Fix: fatal error in the Sabre library caused by the Strauss namespace handling

### LINKED ISSUE #952: Call to undefined function `WPO\IPS\Vendor\Sabre\Xml\Serializer\standardSerializer()`
> Fatal error: Call to undefined function WPO\IPS\Vendor\Sabre\Xml\Serializer\standardSerializer() (.../www/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/vendor/sabre/xml/lib/Writer.php:104)

alexmigf:
closes #952

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -5,7 +5,7 @@ Tags: woocommerce, pdf, ubl, invoices, packing slips
 Requires at least: 4.4
 Tested up to: 6.7
 Requires PHP: 7.4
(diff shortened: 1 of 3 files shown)


alexmigf:
The customer reported that the issue is related to an Amelia plugin. Closing this for now.

PR #956, opened 2024-12-30
New: Add `cac:Price` support to UBL

### LINKED ISSUE #955: Add `cac:Price` support to UBL
Implement the `cac:Price` element to represent unit pricing details, including `cbc:PriceAmount`.

alexmigf:
closes #955

### DIFF
--- a/ubl/Handlers/Invoice/InvoiceLineHandler.php
+++ b/ubl/Handlers/Invoice/InvoiceLineHandler.php
@@ -140,6 +140,18 @@ class InvoiceLineHandler extends UblHandler {
 							),
 						),
 					),
+					array(
+						'name'  => 'cac:Price',
+						'value' => array(
+							array(
+								'name'       => 'cbc:PriceAmount',
+								'value'      => round( $this->get_item_unit_price( $item ), 2 ),
+								'attributes' => array(
+									'currencyID' => $this->document->order->get_currency(),
+								),
+							),
+						),
+					),
 				),
 			);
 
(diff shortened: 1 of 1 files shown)


ubl/Handlers/Invoice/InvoiceLineHandler.php
```diff
@@ -152,4 +164,20 @@ public function handle( $data, $options = array() ) {
...
 
+	/**
+	 * Get the unit price of an item
+	 *
+	 * @param WC_Order_Item $item
+	 * @return int|float
+	 */
+	private function get_item_unit_price( $item ) {
+		if ( is_a( $item, 'WC_Order_Item_Product' ) ) {
+			return $item->get_subtotal() / $item->get_quantity();
```
MohamadNateqi, line 175: Maybe you can use `get_item_subtotal()` method?
`$this->document->order->get_item_subtotal( $item, false, true );`

alexmigf: Good catch!

MohamadNateqi: Yes, it's the same. The only advantage might be using the provided source function or the applied filter. If those don't matter, the current code is fine, as it doesn't change over time.

alexmigf: This seems good for now, and we can always tweak it later!

MohamadNateqi (approved):
LGTM!

PR #962, opened 2024-12-31
Fix postcode and city defaults in UBL tax output

### LINKED ISSUE #961: UBL tax postcodes not reflecting WooCommerce tax settings
Postcodes defined in WooCommerce tax settings are not fully reflected in UBL taxes. Only one postcode is displayed when multiple postcodes are expected.

alexmigf:
closes #961

### DIFF
--- a/ubl/Settings/TaxesSettings.php
+++ b/ubl/Settings/TaxesSettings.php
@@ -62,7 +62,7 @@ class TaxesSettings {
 						foreach ( $results as $result ) {
 							$locationResults = $wpdb->get_results( $wpdb->prepare( "SELECT * FROM {$wpdb->prefix}woocommerce_tax_rate_locations WHERE tax_rate_id = %d;", $result->tax_rate_id ) );
 							$postcode        = array();
-							$city            = '';
+							$city            = array();
 							
 							foreach ( $locationResults as $locationResult ) {
 								if ( ! isset( $locationResult->location_type ) ) {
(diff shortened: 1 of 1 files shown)


ubl/Settings/TaxesSettings.php
```diff
@@ -61,20 +61,26 @@ public function output_table_for_tax_class( $slug, $name ) {
 					if ( ! empty( $results ) ) {
 						foreach ( $results as $result ) {
 							$locationResults = $wpdb->get_results( $wpdb->prepare( "SELECT * FROM {$wpdb->prefix}woocommerce_tax_rate_locations WHERE tax_rate_id = %d;", $result->tax_rate_id ) );
-							$postcode = $city = '';
-
+							$postcode        = array();
```
MohamadNateqi, line 64: Could you please explain to me why this changed to an array?
Since the query is getting all tax locations for a `tax_rate_id`, and the `tax_rate_id` seems to be unique in the `wp_woocommerce_tax_rates` table, so how can we have several tax locations for one tax rate?

ubl/Settings/TaxesSettings.php
```diff
@@ -61,20 +61,26 @@ public function output_table_for_tax_class( $slug, $name ) {
...
+										break;
 								}
 							}
-
+							
+							$postcode = empty( $postcode ) ? '*' : implode( '; ', $postcode );
+							$city     = empty( $city ) ? '*' : $city;
 							$scheme   = isset( $this->settings['rate'][ $result->tax_rate_id ]['scheme'] )   ? $this->settings['rate'][ $result->tax_rate_id ]['scheme']   : '';
 							$category = isset( $this->settings['rate'][ $result->tax_rate_id ]['category'] ) ? $this->settings['rate'][ $result->tax_rate_id ]['category'] : '';
 							$reason   = isset( $this->settings['rate'][ $result->tax_rate_id ]['reason'] )   ? $this->settings['rate'][ $result->tax_rate_id ]['reason']   : '';
```
MohamadNateqi, line 86: Replacing the empty string with a `*` would be good.

MohamadNateqi (changes requested):

ubl/Settings/TaxesSettings.php
alexmigf: The `tax_rate_id` is unique in `wp_woocommerce_tax_rates`, but in `wp_woocommerce_tax_rate_locations`, it allows multiple entries to associate one tax rate with multiple locations (e.g., postcodes or cities). This is why `$postcode` is an array—it captures all postcodes linked to the same tax rate.

ubl/Settings/TaxesSettings.php
alexmigf: Why? Keep in mind that these variables are used in selectors.

MohamadNateqi: I thought it's better to show a `*` instead of an empty string here:

image

Same as WooCommerce.

alexmigf: Done.

MohamadNateqi (approved):
All good!

PR #973, opened 2025-01-03
New: use dynamic UBL tax settings instead of saving to order

### LINKED ISSUE #972: Use dynamic UBL tax settings if latest settings
Use UBL tax settings dynamically instead of saving them to the order when the latest settings are in use.

alexmigf:
closes #972

### DIFF
--- a/ubl/Documents/Document.php
+++ b/ubl/Documents/Document.php
@@ -29,6 +29,7 @@ abstract class Document {
 
 	public function set_order_document( OrderDocument $order_document ) {
 		$this->order_document = $order_document;
+		$this->set_order( $order_document->order );
 	}
 
 	abstract public function get_root_element();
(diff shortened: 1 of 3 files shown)


ubl/Documents/Document.php
```diff
@@ -81,38 +89,38 @@ public function get_tax_rates() {
...
-					$reason = $this->get_tax_data_from_fallback( 'reason', $rate_id );
-					wc_update_order_item_meta( $tax_item_key, '_wcpdf_ubl_tax_reason', $reason );
+				foreach ( $fields as $key => $meta_key ) {
+					$value = wc_get_order_item_meta( $tax_item_key, $meta_key, true );
+					
+					if ( empty( $value ) || ! $use_historical_settings ) {
+						$value = $this->get_tax_data_from_fallback( $key, $rate_id );
+					}
+					
+					if ( $use_historical_settings ) {
```
MohamadNateqi, line 119: What if `$use_historical_settings` is `false` and also the `$value` return on line `113` is also empty? 
On line `116`, a new value is retrieved but not updated for the order item meta in this case. Right?

alexmigf: Yes, for both.

MohamadNateqi (approved):
LGTM!

PR #975, opened 2025-01-03
New: add utility function for dynamic string translation

### LINKED ISSUE #974: Dynamic string translation function
Add a utility function to handle dynamic string translations.

alexmigf:
closes #974

This function uses the `translate()` function to retrieve translations for runtime-determined strings, with a fallback to the original string if no translation exists. It ensures compatibility with existing `.po` translations while safely handling cases where the `translate()` function is unavailable.


MohamadNateqi (approved):
LGTM!

May I know where this function is going to be used?

alexmigf:
For dynamic strings where traditional functions bypass gettext, such as in cases like:

```php
__( $string, 'textdomain' );
```

PR #978, opened 2025-01-03
Tweak: improves the dynamic translate function.

### LINKED ISSUE #977: Improve dynamic translate function
Enhance the dynamic translation function by incorporating a Pro translation feature.

alexmigf:
New: translation logs setting

closes #977

### DIFF
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -998,7 +998,7 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
  * @return string
  */
 function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): string {
-	$log_enabled		= isset( WPO_WCPDF()->settings->debug_settings['translation_logs'] );
+	$log_enabled		= isset( WPO_WCPDF()->settings->debug_settings['log_missing_translations'] );
 	$log_message        = "Missing translation for: {$string} in textdomain: {$textdomain}";
 	$multilingual_class = '\WPO\WC\PDF_Invoices_Pro\Multilingual_Full';
 	$translation        = '';
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -1016,7 +1016,7 @@ function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): stri
 	}
 	
 	// If multilingual didn't change the string, fall back to native translate()
-	if ( $translation === $string && function_exists( 'translate' ) ) {
+	if ( ( empty( $translation ) || $translation === $string ) && function_exists( 'translate' ) ) {
 		$translation = translate( $string, $textdomain );
 	}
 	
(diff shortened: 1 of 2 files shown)


wpo-ips-functions.php
```diff
@@ -999,10 +999,22 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
...
+	}
+
+	// Check for multilingual support class
+	$multilingual_class = '\WPO\WC\PDF_Invoices_Pro\Multilingual_Full';
+	if ( class_exists( $multilingual_class ) && method_exists( $multilingual_class, 'maybe_get_string_translation' ) ) {
+		$translation = $multilingual_class::maybe_get_string_translation( $string, $textdomain );
+	}
+
+	// Fallback to translate()
+	$translation = $translation ?? translate( $string, $textdomain );
```
MohamadNateqi, line 1012: Do you think it should also use the fallback option if the translation is not null, but is the same as the given string?

wpo-ips-functions.php
```diff
@@ -999,10 +999,22 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
  */
 function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): string {
 	if ( ! function_exists( 'translate' ) ) {
-		return $string;
+		return esc_html( $string );
```
MohamadNateqi, line 1002: Is it necessary to escape HTML?

wpo-ips-functions.php
```diff
@@ -999,10 +999,22 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
...
+	if ( class_exists( $multilingual_class ) && method_exists( $multilingual_class, 'maybe_get_string_translation' ) ) {
+		$translation = $multilingual_class::maybe_get_string_translation( $string, $textdomain );
+	}
+
+	// Fallback to translate()
+	$translation = $translation ?? translate( $string, $textdomain );
+
+	// Log missing translations for debugging
+	if ( $translation === $string && isset( WPO_WCPDF()->settings->debug_settings['translation_logs'] ) ) {
+		error_log( "Missing translation for: {$string} in textdomain: {$textdomain}" );
```
MohamadNateqi, line 1016: Maybe `wcpdf_log_error()`?

wpo-ips-functions.php
```diff
@@ -998,11 +998,32 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
...
 	}
 	
-    $translation = translate( $string, $textdomain );
+	// Check for multilingual support class
+	if ( class_exists( $multilingual_class ) && method_exists( $multilingual_class, 'maybe_get_string_translation' ) ) {
+		$translation = $multilingual_class::maybe_get_string_translation( $string, $textdomain );
+	}
+	
+	// If multilingual didn't change the string, fall back to native translate()
+	if ( $translation === $string && function_exists( 'translate' ) ) {
```
MohamadNateqi, line 1019: I think it also needs to check if the `$translation` is empty.

MohamadNateqi (changes requested):

wpo-ips-functions.php
alexmigf: Since we already know that `$string` is not empty, I don't believe is necessary.

alexmigf: You're right. Fixed.

MohamadNateqi (approved):
LGTM!

Nice new feature. 👌

PR #980, opened 2025-01-03
Fix: migrate UBL Tax Scheme settings

### LINKED ISSUE #979: Migrate UBL tax scheme settings
The legacy schemes used lowercase values, but we are now transitioning to uppercase. As a result, the settings need to be updated accordingly.

alexmigf:
closes #979


MohamadNateqi (approved):
LGTM!

PR #987, opened 2025-01-06
Tweak: removes standard default selector for UBL Tax Reason

### LINKED ISSUE #986: Include a "None" option in the Reason dropdown for UBL Taxes.
Add a "None" option to the selector because "Default," while its value is also empty, forces the use of the default line value, which may differ from truly empty.

alexmigf:
closes #986

This PR removes the default "standard" selector for the UBL Tax Reason. In most cases, specifying a tax reason is unnecessary, and having a universal default for all taxes is generally impractical.


MohamadNateqi (approved):
LGTM!

PR #993, opened 2025-01-07
Fix setting categories not applied for disabled documents

### LINKED ISSUE #992: Setting categories not applied for disabled documents
When a document is not enabled, setting categories is not applied!

image

MohamadNateqi:
close #992


alexmigf (approved):
LGTM!

PR #994, opened 2025-01-08
Logo height setting can break document layout

### LINKED ISSUE #976: Logo height setting can break document layout
Currently, setting the 'Logo height' setting to a large number can cause the document layout to break.

 
Scherm­afbeelding 2025-01-03 om 11 45 48

<br>Also logos with a very wide aspect ratio (for instance 10:2) can cause the preview to turn completely blank even when the logo setting is relatively low, like `3cm`. This can confuse users.

We could set a maximum width in the CSS to prevent this.

```
td.header img {
	max-height: 3cm; /* may be overridden by the settings */
	width: auto;
	max-width: 100%;
}
```

BrunoPavlinic98:
closes #976 

**IMPORTANT NOTE - Do the same changes in templates plugin for all templates.**
Done - https://github.com/wpovernight/woocommerce-pdf-ips-templates/issues/364

### DIFF
--- a/templates/Simple/style.css
+++ b/templates/Simple/style.css
@@ -114,8 +114,8 @@ table.head {
 
 td.header img {
 	max-height: 3cm; /* may be overridden by the settings */
-	width: auto;
 	max-width: 100%;
+	width: auto;
 }
 
 /* .mpdf td.header img { */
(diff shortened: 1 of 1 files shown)


BrunoPavlinic98:
@alexmigf This is OK, I tested it, but we also need to do the same changes to the templates plugin. I will do that as well once this is merged.

BrunoPavlinic98:
> @BrunoPavlinic98 You can go ahead and open the issue there.

@alexmigf Done - https://github.com/wpovernight/woocommerce-pdf-ips-templates/issues/364

alexmigf (approved):
LGTM!

PR #999, opened 2025-01-10
Tweak: improves UBL get tax data from fallback function

### LINKED ISSUE #998: Improvements to UBL tax data from fallback
The fallback function for UBL tax data needs refinement to ensure it returns the standard default value when the tax rate ID is not specified.

alexmigf:
closes #998

### DIFF
--- a/wpo-ips-functions-ubl.php
+++ b/wpo-ips-functions-ubl.php
@@ -25,35 +25,37 @@ function wpo_ips_ubl_sanitize_string( string $string ): string {
 /**
  * Get UBL tax data from fallback
  *
- * @param string $key      Can be category, scheme, or reason
- * @param int    $rate_id  The tax rate ID
+ * @param string   $key      Can be category, scheme, or reason
+ * @param int|null $rate_id  The tax rate ID
  * @return string
  */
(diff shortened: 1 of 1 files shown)


wpo-ips-functions-ubl.php
```diff
@@ -25,35 +25,37 @@ function wpo_ips_ubl_sanitize_string( string $string ): string {
...
 	if ( ! in_array( $key, array( 'category', 'scheme', 'reason' ) ) ) {
 		return $result;
 	}
-
+	
+	$tax_rate_class   = '';
+	$ubl_tax_settings = get_option( 'wpo_wcpdf_settings_ubl_taxes', array() );
+	
 	if ( class_exists( '\WC_TAX' ) && is_callable( array( '\WC_TAX', '_get_tax_rate' ) ) ) {
 		$tax_rate = \WC_Tax::_get_tax_rate( $rate_id, OBJECT );
```
MohamadNateqi, line 43: I think you need to check if the `$rate_id` is not `null` here before using it?

alexmigf: Good catch! Fixed here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/0959ff768338db77f7e137473b0994f3cb81c81b

PR #1028, opened 2025-01-23
Tweak: Replace `translate()` with `gettext` filters in dynamic translation function

### LINKED ISSUE #1027: Replace `translate()` with `gettext` filters for dynamic translations
Replaced the use of `translate()` in the `wpo_wcpdf_dynamic_translate()` function with the `gettext` filters.

alexmigf:
closes #1027

### DIFF
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -1051,7 +1051,7 @@ function wpo_wcpdf_escape_url_path_or_base64( string $url_path_or_base64 ): stri
  * @return string
  */
 function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): string {
-	$log_enabled		= isset( WPO_WCPDF()->settings->debug_settings['log_missing_translations'] );
+	$log_enabled        = isset( WPO_WCPDF()->settings->debug_settings['log_missing_translations'] );
 	$log_message        = "Missing translation for: {$string} in textdomain: {$textdomain}";
 	$multilingual_class = '\WPO\WC\PDF_Invoices_Pro\Multilingual_Full';
 	$translation        = '';
@@ -1068,9 +1068,10 @@ function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): stri
 		$translation = $multilingual_class::maybe_get_string_translation( $string, $textdomain );
 	}
 	
-	// If multilingual didn't change the string, fall back to native translate()
-	if ( ( empty( $translation ) || $translation === $string ) && function_exists( 'translate' ) ) {
-		$translation = translate( $string, $textdomain );
+	// If no translation found, apply gettext filters
+	if ( empty( $translation ) || $translation === $string ) {
+		$translation = apply_filters( 'gettext', $string, $string, $textdomain );
+		$translation = apply_filters( "gettext_{$textdomain}", $translation, $string, $textdomain );
 	}
 	
 	// Log missing translations for debugging if it's still untranslated
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -1054,7 +1054,7 @@ function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): stri
 	static $cache       = array();
 	static $logged      = array();
 	
-	$cache_key          = $textdomain . '::' . $string;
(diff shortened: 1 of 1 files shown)


MohamadNateqi (approved):
LGTM!

May I ask why this has been replaced?

alexmigf:
Since `translate()` requires the actual string to be passed directly instead of a variable, it can be somewhat limiting. While this isn't a significant issue, the function already leverages `gettext` filters, so it made sense to use them directly for dynamic translations instead.

MohamadNateqi (approved):

wpo-ips-functions.php
```diff
@@ -1051,35 +1051,56 @@ function wpo_wcpdf_escape_url_path_or_base64( string $url_path_or_base64 ): stri
  * @return string
  */
 function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): string {
-	$log_enabled        = isset( WPO_WCPDF()->settings->debug_settings['log_missing_translations'] );
+	static $cache       = array();
+	static $logged      = array();
```
MohamadNateqi, line 1055: Maybe the `$logged` is not necessary and only `$cached` can be used?

wpo-ips-functions.php
```diff
@@ -1051,35 +1051,56 @@ function wpo_wcpdf_escape_url_path_or_base64( string $url_path_or_base64 ): stri
...
 	$multilingual_class = '\WPO\WC\PDF_Invoices_Pro\Multilingual_Full';
 	$translation        = '';
 	
-	if ( empty( $string ) ) {
-		if ( $log_enabled ) {
+	// Return early if empty string
+	if ( '' === $string ) {
+		if ( $log_enabled && ! isset( $logged[ $cache_key ] ) ) {
 			wcpdf_log_error( $log_message, 'warning' );
+			$logged[ $cache_key ] = true;
```
MohamadNateqi, line 1067: I think it's not necessary to keep an empty string in the log?

wpo-ips-functions.php
```diff
@@ -1051,35 +1051,56 @@ function wpo_wcpdf_escape_url_path_or_base64( string $url_path_or_base64 ): stri
...
 	$log_message        = "Missing translation for: {$string} in textdomain: {$textdomain}";
 	$multilingual_class = '\WPO\WC\PDF_Invoices_Pro\Multilingual_Full';
 	$translation        = '';
 	
-	if ( empty( $string ) ) {
-		if ( $log_enabled ) {
+	// Return early if empty string
+	if ( '' === $string ) {
+		if ( $log_enabled && ! isset( $logged[ $cache_key ] ) ) {
 			wcpdf_log_error( $log_message, 'warning' );
```
MohamadNateqi, line 1066: Would it be better to use a different log message when the string is empty?

MohamadNateqi (changes requested):

wpo-ips-functions.php
alexmigf: I believe it's crucial—translations can be a real headache, so the more information we have, the better.

wpo-ips-functions.php
alexmigf: I think it is necessary; this way, you can clearly see when an empty string is being passed, which can be important for debugging or tracking.

wpo-ips-functions.php
MohamadNateqi: Actually I wanted to say that the `$cache` can be used instead of `$logged`, but it's not so much different. So all good.

MohamadNateqi (approved):
LGTM!

wpo-ips-functions.php
```diff
@@ -1054,7 +1054,7 @@ function wpo_wcpdf_dynamic_translate( string $string, string $textdomain ): stri
 	static $cache       = array();
 	static $logged      = array();
 	
-	$cache_key          = $textdomain . '::' . $string;
+	$cache_key          = md5( $textdomain . '::' . $string );
```
MohamadNateqi, line 1057: May I ask why `md5`?

MohamadNateqi (approved):
All good!

wpo-ips-functions.php
alexmigf: To create a short, unique string from potentially long or complex values.

PR #1039, opened 2025-01-30
Add document link to emails for guest customers

### LINKED ISSUE #1035: Add Setting to Include Document Link in Email for Guest Customers
Introduce a new setting to include a document link in email notifications when "Guest" access is enabled. The link should only be sent to Guest customers.

MohamadNateqi:
close #1035


includes/Documents/Invoice.php
```diff
@@ -243,6 +243,22 @@ public function get_pdf_settings_fields( $option_name ) {
 					'description'	  => ! $wp_filesystem->is_writable( WPO_WCPDF()->main->get_tmp_path( 'attachments' ) ) ? '<span class="wpo-warning">' . sprintf( __( 'It looks like the temp folder (<code>%s</code>) is not writable, check the permissions for this folder! Without having write access to this folder, the plugin will not be able to email invoices.', 'woocommerce-pdf-invoices-packing-slips' ), WPO_WCPDF()->main->get_tmp_path( 'attachments' ) ).'</span>':'',
 				)
 			),
+			array(
+				'type'     => 'setting',
+				'id'       => 'add_download_link_to_emails',
+				'title'    => __( 'Add download link to emails', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 249: Maybe better "Include document link in guest emails"?

includes/Documents/Invoice.php
```diff
@@ -243,6 +243,22 @@ public function get_pdf_settings_fields( $option_name ) {
...
+				'callback' => 'select',
+				'section'  => $this->type,
+				'args'     => array(
+					'option_name'      => $option_name,
+					'id'               => 'add_download_link_to_emails',
+					'options_callback' => array( WPO_WCPDF()->main, 'get_wc_emails' ),
+					'multiple'         => true,
+					'enhanced_select'  => true,
+					'disabled'         => 'guest' !== WPO_WCPDF()->endpoint->get_document_link_access_type(),
+					'description'      => __( 'Select emails to include the document download link. This applies only to emails sent to "Guest" customers when the "Guest" access type is selected.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 259: I'd include a link here to the document access type page.

includes/Documents/Invoice.php
```diff
@@ -243,6 +243,22 @@ public function get_pdf_settings_fields( $option_name ) {
...
 			),
+			array(
+				'type'     => 'setting',
+				'id'       => 'add_download_link_to_emails',
+				'title'    => __( 'Add download link to emails', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'select',
+				'section'  => $this->type,
+				'args'     => array(
+					'option_name'      => $option_name,
+					'id'               => 'add_download_link_to_emails',
```
alexmigf, line 254: Maybe better: `include_link_guest_emails`

includes/Main.php
```diff
@@ -1770,6 +1773,80 @@ function_exists( 'WPO_WCPDF_Templates' ) &&
 			</tr>';
 		}
 	}
+
+	/**
+	 * Get WC emails as array of id => title.
+	 *
+	 * @return array
+	 */
+	public function get_wc_emails(): array {
```
alexmigf, line 1782: There are other instances where we retrieve WC emails, so it might be better to move this to the functions file.

includes/Endpoint.php
```diff
@@ -87,20 +87,21 @@ public function handle_document_requests() {
 		}
 	}
 
-	public function get_document_link( $order, $document_type, $additional_vars = array() ) {
+	public function get_document_link( $order, $document_type, $additional_vars = array(), bool $force_guest_access_key = false ) {
```
alexmigf, line 90: Maybe `$bypass_login_access` instead?

includes/Main.php
```diff
@@ -1770,6 +1773,80 @@ function_exists( 'WPO_WCPDF_Templates' ) &&
...
+
+			$is_allowed = in_array( $document->get_type(), $allowed_document_types, true ) && in_array( $email->id, $selected_emails, true );
+
+			if ( ! apply_filters( 'wpo_wcpdf_add_document_download_link_to_email_is_allowed', $is_allowed, $order, $sent_to_admin, $plain_text, $email ) ) {
+				continue;
+			}
+
+			$document = wcpdf_get_document( $document->get_type(), $order );
+
+			if ( ! $document ) {
```
alexmigf, line 1828: Maybe it's better to add `! $document->exists()` here.

includes/Main.php
```diff
@@ -1770,6 +1773,80 @@ function_exists( 'WPO_WCPDF_Templates' ) &&
...
+			$link_url  = WPO_WCPDF()->endpoint->get_document_link( $order, $document->get_type(), array(), true );
+
+			$document_link = sprintf(
+				'<p><a id="%s" href="%s" target="_blank">%s</a></p>',
+				esc_attr( 'wpo_wcpdf_' . $document->get_type() . '_download_link' ),
+				esc_url( $link_url ),
+				esc_html( $link_text )
+			);
+
+			echo apply_filters( 'wpo_wcpdf_add_document_download_link_to_email_link', $document_link, $document, $order, $sent_to_admin, $plain_text, $email );
```
alexmigf, line 1846: Maybe there's an extra "link" word here?

includes/Main.php
```diff
@@ -1770,6 +1773,80 @@ function_exists( 'WPO_WCPDF_Templates' ) &&
...
+			$link_text = sprintf(
+				/* translators: %s: Document type */
+				__( 'Download %s (PDF)', 'woocommerce-pdf-invoices-packing-slips' ),
+				wp_kses_post( $document->get_type() )
+			);
+			$link_url  = WPO_WCPDF()->endpoint->get_document_link( $order, $document->get_type(), array(), true );
+
+			$document_link = sprintf(
+				'<p><a id="%s" href="%s" target="_blank">%s</a></p>',
+				esc_attr( 'wpo_wcpdf_' . $document->get_type() . '_download_link' ),
```
alexmigf, line 1841: I suggest renaming all instances of `download_link` to `document_link`.

alexmigf (changes requested):
Some remarks:

* This setting does not need to trigger a Preview refresh.
* It might be useful to allow selecting the email template hook where the link is displayed.
* Instead of disabling the field, we could display it only when the "Guest" access type is enabled. In this case, we would also need to update the field description.

includes/Main.php
MohamadNateqi: Wouldn't it be better to allow document creation if it's not yet created?

includes/Main.php
MohamadNateqi: The rest of the cases are the function name. Do you also suggest to change the function name to `add_document_link_to_email`?

MohamadNateqi:
@alexmigf 

> It might be useful to allow selecting the email template hook where the link is displayed.

Do you suggest such a setting:
image
It's a multiselect to allow choosing several hooks.


> Instead of disabling the field, we could display it only when the "Guest" access type is enabled. In this case, we would also need to update the field description.

I thought this way, customers would know that such a setting exists easily.
So, do you believe hiding it and adding a description for the access type setting to inform about this feature would be better?

alexmigf:
> It's a multiselect to allow choosing several hooks.

I believe a standard selector is more appropriate since the user is unlikely to select more than one option.

> I thought this way, customers would know that such a setting exists easily.
So, do you believe hiding it and adding a description for the access type setting to inform about this feature would be better?

I understand your point, but I believe we should avoid promoting the "Guest" access type, as it is less secure. Customers already using "Guest" will see the option, and I believe they are the intended target.

includes/Main.php
MohamadNateqi: @alexmigf what do you think about this?

MohamadNateqi: I like the new setting idea, but I have concerns about overcrowding with various options.
How about we choose one as the default behavior for now and provide a filter instead of introducing a new setting?

alexmigf: I agree.

MohamadNateqi: So, which one do you vote for as the default behavior?
1. Allow document creation if it doesn't exist.
2. Don't allow.

alexmigf: I would go with the more defensive.

includes/Documents/Invoice.php
```diff
@@ -553,6 +553,50 @@ public function get_pdf_settings_fields( $option_name ) {
...
+				'type'     => 'setting',
+				'id'       => 'include_link_guest_emails_placement',
+				'title'    => __( 'Placement of document link in guest emails', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'select',
+				'section'  => $this->type,
+				'args'     => array(
+					'option_name' => $option_name,
+					'id'          => 'include_link_guest_emails_placement',
+					'options'     => apply_filters( 'wpo_wcpdf_document_link_guest_emails_template_hooks_options', array(
+						'order_details'                     => 'Order details',
```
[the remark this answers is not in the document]
MohamadNateqi: I enjoy space! Just like I enjoy big American cars that have plenty of room... Just kidding! 😁

includes/Documents/Invoice.php
```diff
@@ -557,7 +557,7 @@ public function get_pdf_settings_fields( $option_name ) {
 			$settings_fields[] = array(
 				'type'     => 'setting',
 				'id'       => 'include_link_guest_emails',
```
alexmigf, line 559: Perhaps we should consider omitting "guest" from the slugs and function names as well.

alexmigf (changes requested):

includes/Assets.php
```diff
@@ -166,8 +166,8 @@ public function backend_scripts_styles( $hook ) {
 						'mark_printed',
 						'unmark_printed',
 						'include_encrypted_pdf',
-						'include_link_guest_emails',
-						'include_link_guest_emails_placement',
+						'include_link_emails',
```
alexmigf, line 169: Maybe `include_email_link` is better?

includes/Documents/Invoice.php
```diff
@@ -553,6 +553,50 @@ public function get_pdf_settings_fields( $option_name ) {
...
+					'id'          => 'include_email_link_placement',
+					'options'     => apply_filters( 'wpo_wcpdf_document_link_guest_emails_template_hooks_options', array(
+						'order_details'            => 'Order details',
+						'order_meta'               => 'Order meta',
+						'before_order_table'       => 'Before order table',
+						'after_order_table'        => 'After order table',
+						'customer_address_section' => 'Customer address section',
+						'customer_details'         => 'Customer details',
+					), $this ),
+					'description' => __( 'Select the placement of the document link in the guest emails.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf, line 595: Maybe "guest customer emails"

includes/Main.php
```diff
@@ -1776,6 +1779,87 @@ function_exists( 'WPO_WCPDF_Templates' ) &&
 			</tr>';
 		}
 	}
+
+	function handle_document_link_in_emails( ): void {
```
alexmigf, line 1783: Space between `()`

includes/Main.php
```diff
@@ -1776,6 +1779,87 @@ function_exists( 'WPO_WCPDF_Templates' ) &&
...
+
+			if ( ! apply_filters( 'wpo_wcpdf_add_document_link_to_email_is_allowed', $is_allowed, $order, $sent_to_admin, $plain_text, $email ) ) {
+				continue;
+			}
+
+			$document = wcpdf_get_document( $document->get_type(), $order );
+
+			if (
+				( ! $document || ! $document->exists() ) &&
+				! apply_filters( 'wpo_wcpdf_add_document_link_to_email_allow_missing_documents', false, $document, $order, $sent_to_admin, $plain_text, $email )
```
alexmigf, line 1840: Doesn't seem to be working because of this check.

includes/Main.php
```diff
@@ -1835,8 +1835,12 @@ public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sen
...
 			$document = wcpdf_get_document( $document->get_type(), $order );
 
+			if ( ! $document ) {
+				continue;
+			}
+
 			if (
-				( ! $document || ! $document->exists() ) &&
+				! $document->exists() &&
 				! apply_filters( 'wpo_wcpdf_add_document_link_to_email_allow_missing_documents', false, $document, $order, $sent_to_admin, $plain_text, $email )
```
alexmigf, line 1844: Your change had no effect. You need to remove the exclamation mark before the filter.

alexmigf (changes requested):

includes/Main.php
alexmigf: Sorry, I was attaching the invoice to the same email, but it seems it hadn’t been generated yet when the link was added, and I didn’t double-check. However, while testing, I realized that your initial idea of generating the document might actually make sense.

MohamadNateqi: I can revert this filter to display the document link if it doesn't exist by default, and so users can also disable it by this filter. What do you think?

alexmigf (approved):
LGTM!

PR #1057, opened 2025-02-11
Use `wp_kses_post` instead of `wpo_wcpdf_sanitize_html_content` for footer & extras

### LINKED ISSUE #1056: Relax sanitization for footer and extras
Currently, the function `wpo_wcpdf_sanitize_html_content` is used to sanitize the footer and extras in PDF output, which prevents some of the required tags, such as `img`. However, since users manually fill these fields in the backend and are not exposed to customers, strict sanitization may not be necessary. Instead, using `wp_kses_post` should be sufficient to prevent malicious inputs while allowing more flexibility.

MohamadNateqi:
close #1056

### DIFF
--- a/includes/Documents/OrderDocument.php
+++ b/includes/Documents/OrderDocument.php
@@ -1338,7 +1338,7 @@ abstract class OrderDocument {
 		return ob_get_clean();
 	}
 	public function footer() {
-		echo wp_kses_post( $this->get_footer() ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo wp_kses_post( $this->get_footer() );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/Documents/OrderDocument.php
```diff
@@ -1338,7 +1338,7 @@ public function get_footer() {
 		return ob_get_clean();
 	}
 	public function footer() {
-		echo wpo_wcpdf_sanitize_html_content( $this->get_footer() ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo wp_kses_post( $this->get_footer() ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
alexmigf, line 1341: I think you could remove the `phpcs` comments.

alexmigf (changes requested):

alexmigf (approved):

PR #1073, opened 2025-02-18
Show inactive status for packing slip yearly reset

### LINKED ISSUE #1071: Conditonal "Documents' Status" values
The Documents' Status panel under PDF Invoices > Advanced > Status, shows the yearly reset value of currently available documents. Since only the invoice has  number sequence in the free version of the plugin, could we hide the packing slip yearly reset value since there is no number sequence for it?

Image

Additionally, if a user has the Professional extension and has activated the yearly reset value the packing slip, upon deactivate the Professional extension, the value in the Documents' Status panel shows it as active. The reset might then trigger for the packing slip even though the Professional extension is deactivated. This could also be the same for the other documents when their respective plugins are activated, such as the Proposal plugin.

MohamadNateqi:
close #1071

image


alexmigf (commented):
How about linking to the Upgrade tab instead of displaying a dash? This could be a great opportunity to promote the upgrade, such as: 'Upgrade to our Professional extension.'

MohamadNateqi:
@alexmigf Nice idea as always! I'll do it now.

alexmigf (changes requested):
image

views/advanced-status.php
```diff
@@ -16,8 +16,8 @@
 	<tbody>
 		<tr>
 			<td class="title">PDF Invoices & Packing Slips for WooCommerce</td>
-			<td><?php esc_html( WPO_WCPDF()->version ); ?></td>
-			<td class="valid-status"><?php esc_html_e( 'Active', 'woocommerce-pdf-invoices-packing-slips' ); ?></td>
+			<td><?php esc_html_e( WPO_WCPDF()->version ); ?></td>
```
alexmigf, line 19: I believe this should be `esc_attr` since it's not translatable.

MohamadNateqi: I think the `esc_attr()` is for an attribute value, but in this case, since it's not translatable, would applying any escaping function would work?

alexmigf: You could use `esc_html` with `echo` instead.

alexmigf (approved):
LGTM!

PR #1078, opened 2025-02-20
New: hybrid filesystem support

### LINKED ISSUE #1052: Issue with temporary directory after updating to version `4.0.0`
A [user](https://wordpress.org/support/topic/after-update-4-0-couldnt-be-created-or-is-not-writable/) reported issues with temporary sub directories after updating to version [4.0.0](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/releases/tag/v4.0.0). I suggested regenerating the temporary directory, but that didn’t resolve the problem. Although the customer confirmed that permissions are correctly set, I was unable to reproduce the issue locally. 

There are no logs available for debugging, making it more challenging to diagnose the issue.

Since granting temporary access isn’t ideal, I will attempt to replicate the problem across different environments, including an external hosting setup.

### User Environment

* WC: 9.6.1
* WP: 6.7.1
* PHP 7.4.33
* Server: Apache

### LINKED ISSUE #1074: Implement interface for hybrid filesystem support
Our plugin currently uses `WP_Filesystem` with the `direct` method, but some users need alternatives like `ssh2`. To provide flexibility, we will introduce a hybrid filesystem solution with a setting to switch between:

* WP_Filesystem (default, using direct method)
* PHP file functions (fallback mode)

To manage this, we'll create a new class, acting as an interface for all file operations in the plugin. This will ensure a consistent and extendable approach to handling file interactions.

### LINKED ISSUE #1114: Some users fonts go missing after updating to v4.2.0
Following the [thread here](https://wordpress.org/support/topic/typography-layout-issue/#post-18393709), our new font check in `v4.2.0` may not be working for everyone. It's apparently removing the fonts for some users upon updating to `v4.2.0`. The fonts are then not auto reinstalled either.

alexmigf:
closes #1074
closes #1052
closes #1114

image

I've added a new filter that allows you to force the filesystem method, providing a workaround if the settings page is inaccessible.

```php
add_filter( 'wpo_wcpdf_filesystem_method', function( $method ) {
	return 'php';
} );
```


MohamadNateqi (changes requested):

includes/Main.php
```diff
@@ -616,18 +615,16 @@ public function get_tmp_path( $type = '' ) {
...
 			if ( ! $dir ) {
 				update_option( 'wpo_wcpdf_no_dir_error', $tmp_path );
 				wcpdf_log_error( "Unable to create folder {$tmp_path}", 'critical' );
 				return false;
 			}
-		} elseif( ! wp_is_writable( $tmp_path ) ) {
+		} elseif( ! WPO_WCPDF()->file_system->is_writable( $tmp_path ) ) {
 			update_option( 'wpo_wcpdf_no_dir_error', $tmp_path );
 			wcpdf_log_error( "Temp folder {$tmp_path} not writable", 'critical' );
 			return false;
```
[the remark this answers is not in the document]
alexmigf: Although I think it's a good idea, it might be better to address it in a separate issue.

MohamadNateqi (approved):
Nice addition! LGTM!

What do you think about checking the WP filesystem and adding an admin notice if it is selected but cannot be initiated? Or a warning under its setting?
Maybe in another issue of course.

alexmigf:
> What do you think about checking the WP filesystem and adding an admin notice if it is selected but cannot be initiated? Or a warning under its setting?

I believe the current approach is sufficient. We log the situation and fall back to PHP methods, ensuring that your store continues to function normally without any issues. Since there are only two available options, there’s not much more you could do I think.

MohamadNateqi (approved):
LGTM!

alexmigf:
I've updated the first comment to include the new filter.

alexmigf:
A [user](https://wordpress.org/support/topic/filesystem-direct-method/#post-18393957) provided positive feedback on these changes.

wpo-ips-functions.php
```diff
@@ -1012,27 +1009,15 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
...
-
-	if ( ! WP_Filesystem() || ! $wp_filesystem ) {
-		$error = 'Failed to initialize WP_Filesystem.';
-		wcpdf_log_error( $error, 'critical' );
-		throw new \RuntimeException( esc_html( $error ) );
+	wcpdf_deprecated_function( 'wpo_wcpdf_get_wp_filesystem', '4.2.0', '\WPO\IPS\Compatibility\FileSystem::instance()->wp_filesystem' );
+	
+	if ( class_exists( '\\WPO\\IPS\\Compatibility\\FileSystem' ) ) {
+		$filesystem = \WPO\IPS\Compatibility\FileSystem::instance();
+		$filesystem->initialize_wp_filesystem();
```
MohamadNateqi, line 1016: @alexmigf The `initialize_wp_filesystem()` is a protected method.

alexmigf: Fixed [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/commit/63474f5eec483ffa11193d47aae2faf1e8edcbde), thank you!

PR #1082, opened 2025-03-07
Remove guest access type

### LINKED ISSUE #1043: Review necessity of "Full" document access type
Assess whether the "Full" document access type option is necessary, given that the "Guest" access type offers a similar functionality.

MohamadNateqi:
close #1043

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -370,8 +370,8 @@ class Main {
 			wcpdf_safe_redirect_or_die( $redirect_url, $message );
 		}
 
-		// Check the nonce
-		if ( is_user_logged_in() && $access_type === 'logged_in' && ! $valid_nonce ) {
+		// Check the nonce for logged in users
+		if ( is_user_logged_in() && 'logged_in' === $access_type && ! $valid_nonce ) {
 			$message = esc_attr__( 'You do not have sufficient permissions to access this page. Reason: invalid nonce', 'woocommerce-pdf-invoices-packing-slips' );
 			wcpdf_safe_redirect_or_die( $redirect_url, $message );
 		}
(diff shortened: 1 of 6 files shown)


includes/Main.php
```diff
@@ -1811,12 +1806,12 @@ function handle_document_link_in_emails(): void {
 	 * @return void
 	 */
 	public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sent_to_admin, bool $plain_text, \WC_Email $email ): void {
-		// Check if document access type is 'guest' and customer is a guest.
-		$is_guest_access_type = 'guest' === WPO_WCPDF()->endpoint->get_document_link_access_type();
-		$is_customer_guest    = 0 === $order->get_customer_id();
+		// Check if document access type is 'full' and customer is a guest.
+		$is_full_access_type = 'full' === WPO_WCPDF()->endpoint->get_document_link_access_type();
+		$is_customer_guest   = 0 === $order->get_customer_id();
```
alexmigf, line 1811: Maybe this variable is no longer needed?

alexmigf (changes requested):

includes/Main.php
MohamadNateqi: You mean not checking if the customer is a guest?

alexmigf: Yes, we could consider removing it. The setting is already visible in the document settings when 'Full' is enabled, which might be confusing.

MohamadNateqi: Maybe we can also not check the access type and allow it in any case?

alexmigf: You should verify if it's set to 'full,' as using the nonce will cause validation to fail when accessing the document, even for logged-in users.

alexmigf (approved):
LGTM!

PR #1087, opened 2025-03-11
New: Add plugin-specific filter hook for order item name

### LINKED ISSUE #1086: Add plugin-specific filter hook to translate the order item name
WooCommerce provides the `woocommerce_order_item_name` filter, but a plugin-specific hook is needed to allow targeted modifications without affecting global behavior. This ensures better compatibility with TranslatePress and other translation plugins.

alexmigf:
closes #1086


MohamadNateqi (approved):
LGTM!

PR #1092, opened 2025-03-12
Introduce a function to reload wc webhooks

### LINKED ISSUE #1083: Webhooks not working
We're loading our classes using the `init` hook with priority `9`, and within that, we register our custom webhook hooks at the same priority (`init`, `9`) here:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/6c9105a193d82a17813bf7b8e41181f51eb091be/includes/Main.php#L80

However, WooCommerce initializes its webhook hooks much earlier, at `init` priority `0`. As a result, our custom webhooks do not get registered in time and fail to work properly.

Previously, this setup worked because we were using the `plugins_loaded` hook, which runs before `init`.

Reported by a [customer](https://secure.helpscout.net/conversation/2867086440/69592?viewId=221731).

MohamadNateqi:
close #1083


includes/Main.php
```diff
@@ -1707,9 +1707,48 @@ public function wc_webhook_topics( $topics ) {
...
+
+		$wpo_topic_hooks = $this->wc_webhook_topic_events( array() );
+		$data_store      = \WC_Data_Store::load( 'webhook' );
+		$webhooks        = $data_store->get_webhooks_ids( 'active' );
+
+		foreach ( $webhooks as $webhook_id ) {
+			$webhook = new \WC_Webhook( $webhook_id );
+			$topic   = str_replace( 'order.', '', $webhook->get_topic() );
+
+			if ( in_array( $topic, $wpo_topic_hooks, true ) ) {
```
alexmigf, line 1746: Maybe you could check if the webhook is pending delivery before enqueue it?

```php
if ( $webhook->get_pending_delivery() ) {
	continue;
}
```

includes/Main.php
```diff
@@ -1707,9 +1707,48 @@ public function wc_webhook_topics( $topics ) {
...
+	private function reload_wpo_custom_webhooks() {
+		if (
+			! apply_filters( 'wpo_wcpdf_reload_wpo_custom_webhooks', true ) ||
+			! class_exists( 'WC_Data_Store' ) ||
+			! class_exists( 'WC_Webhook' )
+		) {
+			return;
+		}
+
+		$wpo_topic_hooks = $this->wc_webhook_topic_events( array() );
```
alexmigf, line 1738: Maybe you could move this empty array to the function arg.

includes/Main.php
```diff
@@ -1707,9 +1707,48 @@ public function wc_webhook_topics( $topics ) {
...
+		$data_store      = \WC_Data_Store::load( 'webhook' );
+		$webhooks        = $data_store->get_webhooks_ids( 'active' );
```
alexmigf, line 1740: Maybe you could return earlier if this is empty.

includes/Main.php
```diff
@@ -1707,9 +1707,48 @@ public function wc_webhook_topics( $topics ) {
...
+
+		foreach ( $webhooks as $webhook_id ) {
+			$webhook = new \WC_Webhook( $webhook_id );
+			$topic   = str_replace( 'order.', '', $webhook->get_topic() );
```
alexmigf, line 1744: Maybe you should check if this is a string and not empty.

alexmigf (changes requested):

alexmigf (approved):
LGTM!

PR #1100, opened 2025-03-18
New: Improve fonts check and reinstatement in temporary fonts directory

### LINKED ISSUE #1099: Improve check if fonts need reinstallation
Improve check if the fonts are available or need reinstallation. We could improve this so the fonts are always available.

alexmigf:
closes #1099

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -680,52 +680,50 @@ class Main {
 	/**
 	 * Checks if the tmp subfolder has files
 	 *
-	 * @param string $subfolder  can be 'attachments', 'fonts' or 'dompdf'
-	 *
+	 * @param string $subfolder Can be 'attachments', 'fonts', or 'dompdf'.
 	 * @return bool
 	 */
-	public function tmp_subfolder_has_files( $subfolder ) {
-		$has_files = false;
-
-		if ( empty( $subfolder ) || ! in_array( $subfolder, $this->subfolders ) ) {
+	public function tmp_subfolder_has_files( string $subfolder ): bool {
+		if ( empty( $subfolder ) || ! in_array( $subfolder, $this->subfolders, true ) ) {
 			wcpdf_log_error( sprintf( 'The directory %s is not a default tmp subfolder from this plugin.', $subfolder ), 'critical' );
-			return $has_files;
+			return false;
 		}
 
-		// we have a cached value
-		if ( get_transient( "wpo_wcpdf_subfolder_{$subfolder}_has_files" ) !== false ) {
-			return wc_string_to_bool( get_transient( "wpo_wcpdf_subfolder_{$subfolder}_has_files" ) );
-		}
+		$cache_key = "wpo_wcpdf_subfolder_{$subfolder}_has_files";
+		$tmp_path  = untrailingslashit( $this->get_tmp_path( $subfolder ) );
 
-		if ( ! function_exists( 'glob' ) ) {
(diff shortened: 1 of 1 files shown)


alexmigf:
@dwalkerpriv Testing this can be tricky. In the previous implementation, if all `.ttf` files were deleted from the fonts directory but other files remained, the directory was still considered non-empty. The new code fixes this issue while also introducing memory and performance optimizations for this check.

includes/Main.php
```diff
@@ -680,52 +680,50 @@ public function get_wp_upload_base () {
...
-
-		if ( empty( $subfolder ) || ! in_array( $subfolder, $this->subfolders ) ) {
+	public function tmp_subfolder_has_files( string $subfolder ): bool {
+		if ( empty( $subfolder ) || ! in_array( $subfolder, $this->subfolders, true ) ) {
 			wcpdf_log_error( sprintf( 'The directory %s is not a default tmp subfolder from this plugin.', $subfolder ), 'critical' );
-			return $has_files;
-		}
-
-		// we have a cached value
-		if ( get_transient( "wpo_wcpdf_subfolder_{$subfolder}_has_files" ) !== false ) {
```
MohamadNateqi, line 696: Why not checking cache first in the refactored code?

includes/Main.php
```diff
@@ -680,52 +680,50 @@ public function get_wp_upload_base () {
...
-			case 'fonts':
-				if ( ! empty( glob( $tmp_path.'/*.ttf' ) ) ) {
-					$has_files = true;
+		try {
+			$iterator = new \FilesystemIterator( $tmp_path, \FilesystemIterator::SKIP_DOTS );
+			
+			foreach ( $iterator as $file ) {
+				// If we don't have a file extension restriction, return true immediately
+				if ( empty( $allowed_extensions[ $subfolder ] ) ) {
+					set_transient( $cache_key, 'yes', HOUR_IN_SECONDS );
```
MohamadNateqi, line 708: Isn't setting the cache duration to just one hour too short? I'm concerned this might increase server load due to repeated checks.

includes/Main.php
```diff
@@ -680,52 +680,57 @@ public function get_wp_upload_base () {
...
-			case 'attachments':
-				if ( ! empty( glob( $tmp_path.'/*.pdf' ) ) ) {
-					$has_files = true;
-				}
-				break;
-			case 'fonts':
-				if ( ! empty( glob( $tmp_path.'/*.ttf' ) ) ) {
-					$has_files = true;
+		// Define allowed extensions per subfolder
+		$allowed_extensions = array(
```
MohamadNateqi, line 703: What do you think about `required_extensions`?
Since we're not checking if only these extensions exist and nothing else.

MohamadNateqi (approved):
LGTM!

includes/Main.php
alexmigf: If you rename it to `$required_extensions`, it might imply that only these extensions are permitted and that the function is enforcing a strict validation, which it isn't.

YordanSoares (approved):
LGTM!

alexmigf:
@dwalkerpriv The fonts will be reinstalled automatically when you open the PDF.

dwalkerpriv (approved):
I couldn't get the missing font error, but i did manage to have the fonts reinstalled after removing them from the folder with this PR 👍

PR #1105, opened 2025-03-25
Fix: Refactor SQL queries for safe table name handling

### LINKED ISSUE #1102: Refactor direct SQL queries to properly escape table names using `%i` or manual sanitization
David pointed out that we're incorrectly using `esc_sql()` to escape table names in SQL queries. This should be replaced with `%i` for WordPress 6.2+ and manual sanitization for older versions.

alexmigf:
closes #1102

I spent some time working on this by creating a helper class to handle similar cases in database queries. However, it turned into a large PR, and I’m not feeling confident about making such a big change at the moment. It’s something we can always revisit and improve in the future.


MohamadNateqi:
@alexmigf Instead of creating a helper class, could we introduce a custom wrapper function, something like `safe_prepare()`, to replace direct calls to prepare()? 
The idea would be to write queries using the new style placeholders (like `%i`), and have the wrapper automatically fall back to manual escaping for older WordPress versions (below 6.2).

alexmigf:
@MohamadNateqi Done.

wpo-ips-functions.php
```diff
@@ -1212,3 +1212,35 @@ function wpo_wcpdf_get_order_customer_vat_number( \WC_Abstract_Order $order ): ?
...
+
+/**
+ * Prepare an identifier query for use with $wpdb->prepare().
+ *
+ * @param string $query
+ * @param array $identifiers
+ * @param array $values
+ * @return string|void
+ */
+function wpo_wcpdf_prepare_identifier_query( string $query, array $identifiers = array(), array $values = array() ) {
```
MohamadNateqi, line 1224: Would it be better to use one argument for both identifiers and values? 
I think it can make calling easier and reduce the likelihood of errors.

alexmigf: I see how a single-array approach might simplify the call. However, keeping identifiers and values separate clarifies the logic, avoids placeholder mix-ups, and remains safer to maintain in the long run. I also think having them separated aligns better with the concerns you’re raising.

wpo-ips-functions.php
```diff
@@ -1212,3 +1212,35 @@ function wpo_wcpdf_get_order_customer_vat_number( \WC_Abstract_Order $order ): ?
...
+	global $wpdb;
+
+	$has_identifier_escape = version_compare( get_bloginfo( 'version' ), '6.2', '>=' );
+
+	if ( $has_identifier_escape ) {
+		return $wpdb->prepare( $query, ...array_merge( $identifiers, $values ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+	}
+
+	foreach ( $identifiers as &$id ) {
+		$id = '`' . preg_replace( '/[^a-zA-Z0-9_]/', '', $id ) . '`';
```
MohamadNateqi, line 1234: Could it cause an issue in case the identifier had some other character such as a `.` or a `-` which I think is not so uncommon?

For example, [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1105/files#diff-bd49c6b6fbe7a6920acaf94a4a3de786f05bd21c2a67f9cef8d8b1f40d54971bR1270) we are also sending the table prefix, which could have a `-` in the name.

includes/Settings/SettingsDebug.php
```diff
@@ -1371,20 +1371,18 @@ public function fetch_number_table_data( string $table_name, string $orderby = '
...
-				$data['orderby'],
-				$data['order'],
-				$chunk_size,
-				$offset
-			)
+		$order       = ( 'DESC' === strtoupper( $data['order'] ) ) ? 'DESC' : 'ASC';
+		$table_name  = $data['table_name'];
+		$orderby     = $data['orderby'];
+
+		$query = wpo_wcpdf_prepare_identifier_query(
```
MohamadNateqi, line 1378: Got this for it:
```php
WordPress database error Unknown column '2025-03-26 23:59:59 23:59:59' in 'order clause' for query SELECT * FROM `wp_wcpdf_invoice_number` WHERE date BETWEEN 'id' AND '2023-03-26 00:00:00 00:00:00' ORDER BY `2025-03-26 23:59:59 23:59:59` DESC LIMIT 100 OFFSET 0 made by do_action('wp_ajax_wpo_wcpdf_numbers_data'), WP_Hook->do_action, WP_Hook->apply_filters, WPO\IPS\Settings\SettingsDebug->ajax_numbers_data, WPO\IPS\Settings\SettingsDebug->fetch_number_table_data
```

Which is related to the orders sent to `prepare()` [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1105/files#diff-5b636c8fa043f3604ee932009a26f40ebbc86e97e3974688fa590a5410d3e681R1230).

MohamadNateqi (changes requested):

wpo-ips-functions.php
alexmigf: Yes, it can happen, but I don't think it's very common – by default, the prefix is usually `wp_`. I won’t add support for dots by default. However, I did add a filter so it can be customized if needed.

wpo-ips-functions.php
```diff
@@ -1212,3 +1212,51 @@ function wpo_wcpdf_get_order_customer_vat_number( \WC_Abstract_Order $order ): ?
...
+		$split            = preg_split( '/(%[%a-zA-Z])/', $query, -1, PREG_SPLIT_DELIM_CAPTURE );
+		
+		foreach ( $split as $part ) {
+			if ( '%i' === $part ) {
+				$all_placeholders[] = $identifiers[ $identifier_index++ ] ?? '';
+			} elseif ( preg_match( '/^%[sdfb]/', $part ) ) {
+				$all_placeholders[] = $values[ $value_index++ ] ?? '';
+			}
+		}
+
```
MohamadNateqi, line 1243: It might be good to check the number of `$all_placeholders` and the number of `$split` to return if not match.

MohamadNateqi (approved):
LGTM!

PR #1106, opened 2025-03-25
New: Add filter to allow customization of document link

### LINKED ISSUE #1104: Add a new filter hook to handle the document link
Currently, there is no way to filter the value returned by the `get_document_link` function, which limits the ability to customize the document link in advanced use cases. Adding a new filter hook would allow users to modify or override the default link behavior. 

This would be particularly useful for scenarios where, for instance, users need to replace the default document link with a third-party URL for specific orders. As a reference, in the following WordPress support topic, you can see an actual case in which this filter could be used: [How to store PDF existing invoices forever](https://wordpress.org/support/topic/how-to-store-pdf-existing-invoices-forever/).

alexmigf:
closes #1104


MohamadNateqi (approved):
LGTM!

PR #1107, opened 2025-03-25
New: Display latest stable and unstable GitHub releases on status page and show admin notice for new unstable versions

### LINKED ISSUE #1069: Include latest and beta plugin versions on Status table
It might be helpful to include both the latest stable version and the latest beta version in the plugin status table.

Image

Starting with the free plugin, as it’s simpler, and later expanding to the paid plugins in coordination with @pwovernight .

alexmigf:
closes #1069

Screenshot from 2025-03-25 14-49-11
Screenshot from 2025-03-25 14-48-10


MohamadNateqi:
@alexmigf Should I see that unstable version if there is a stable version after that?

image

alexmigf:
@MohamadNateqi improved.

MohamadNateqi (changes requested):

includes/Settings/SettingsDebug.php
```diff
@@ -891,6 +898,23 @@ public function init_settings() {
...
+				'title'    => __( 'Check for unstable versions', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'checkbox',
+				'section'  => 'debug_settings',
+				'args'     => array(
+					'option_name' => $option_name,
+					'id'          => 'check_unstable_versions',
+					'description' => sprintf(
+						/* translators: %1$s: open status page link anchor, %2$s: close status page link anchor */
+						__( 'Enable this to check for new beta or pre-release versions and display them in the %1$sstatus page%2$s. A notice will appear when a new version is available.', 'woocommerce-pdf-invoices-packing-slips' ),
+						'<a href="' . esc_url( admin_url( 'admin.php?page=wpo_wcpdf_options_page&tab=debug&section=status' ) ) . '">',
```
MohamadNateqi, line 913: Would it be good to add the `#check_unstable_versions` at the end of the URL?

MohamadNateqi (changes requested):
When the setting is not checked:
Last stable is empty for the free plugin:

image

<hr>

Also, the current version of the free plugin for me is `4.2.1-beta.1`, but it's not displayed in the table.

<hr>

Getting this in logs:

```php
PHP Notice:  Undefined index: stable in ...\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\views\advanced-status.php on line 28
PHP Notice:  Trying to access array offset on value of type null in ...\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\views\advanced-status.php on line 28
```

alexmigf:
@MohamadNateqi 

> Also, the current version of the free plugin for me is 4.2.1-beta.1, but it's not displayed in the table.

I'm a bit confused—since I haven't merged `main` yet, it should still be `4.2.0.` See [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e504a00dd25ed3edba9b558b08be09b3b15c015c/woocommerce-pdf-invoices-packingslips.php#L25) for reference.

includes/Settings/SettingsDebug.php
alexmigf: Why?

MohamadNateqi: Yes, exactly this.

MohamadNateqi:
> I'm a bit confused—since I haven't merged main yet, it should still be 4.2.0.

I actually merged main on my environment. I'll look into this again.

includes/Settings/SettingsDebug.php
```diff
@@ -88,10 +88,12 @@ public function display_settings() {
...
-		$server_configs        = $this->get_server_config();
-		$premium_plugins       = $this->get_premium_plugins();
-		$directory_permissions = $this->get_directory_permissions();
-		$yearly_reset_schedule = $this->get_yearly_reset_schedule();
+		$server_configs         = $this->get_server_config();
+		$premium_plugins        = $this->get_premium_plugins();
+		$directory_permissions  = $this->get_directory_permissions();
+		$yearly_reset_schedule  = $this->get_yearly_reset_schedule();
+		$debug_settings         = WPO_WCPDF()->settings->debug_settings;
+		$latest_github_releases = wpo_wcpdf_get_latest_releases_from_github();
```
MohamadNateqi, line 96: I think the condition you added previously was good. It could prevent the execution of `wpo_wcpdf_get_latest_releases_from_github()`.

Of course, if the "last stable" version should be displayed, executing this function is necessary, and you can ignore my comment.
But if we consider removing the "last stable" column when that setting is not checked, then we don't also need to execute this function to get the latest releases.

I am concerned about running a process to retrieve data from an external resource while a user may not want it.

What do you think?

alexmigf: I think it makes sense to display the last stable version. Since this only loads on the status page and I’ve implemented a caching system, the impact should be minimal. I don’t see a problem with the external request for now, but we can consider an alternative solution if it becomes an issue in the future.

MohamadNateqi (approved):
LGTM!

MohamadNateqi (approved):
LGTM!

dwalkerpriv (commented):
It works on my end. One question I have is when the first notice appears that lets the user know we've added the new table, the Settings button goes to the Advanced > Settings page. Should this take them to the Advanced > Status page so they can see and know where the table is instead?

alexmigf:
@dwalkerpriv The first notice after the upgrade simply informs you about the new setting, which is disabled by default. Unstable versions will only appear in the status table after you enable this setting.

dwalkerpriv (approved):
All is good 👍

PR #1111, opened 2025-03-27
Fix: Numbers tab datepicker style when `SCRIPT_DEBUG` is defined

### LINKED ISSUE #1110: Date picker not displaying correctly on Numbers tab
The date picker on the Numbers tab appears broken. It's not rendering the full UI properly, making it difficult to select dates.

Image

alexmigf:
closes #1110

For testing, Premium Templates extension must be disabled because it loads jQuery UI.


MohamadNateqi (approved):
LGTM!

Strangely, I cannot replicate this issue on the `main` branch.

alexmigf:
Have you defined `SCRIPT_DEBUG` in your `wp-config.php` file? If not, WordPress will load the minified version of the stylesheet, so you shouldn't see the issue. However, if `SCRIPT_DEBUG` is enabled, it tries to load the unminified version — which doesn't exist locally — and that’s why the issue appears.

MohamadNateqi:
Yeah, that was the issue. Good point.

PR #1121, opened 2025-04-04
Allow `<style>` tag for footer

### LINKED ISSUE #1120: HTML is not rendered properly in the footer
Following the [thread here](https://wordpress.org/support/topic/html-table-in-footer-issue-since-version-4/), HTML code in the footer under PDF Invoices > General > Footer, do not render properly.

We had this issue before with a user that also had the Premium Templates that printed the table in a custom block in the footer section. A fix in `v4.1.0` of the free plugin in combination with the Custom Block's raw HTML mode solved the issue there.

Maybe our sanitization methods needs further revision.

MohamadNateqi:
close #1120

### DIFF
--- a/includes/Documents/OrderDocument.php
+++ b/includes/Documents/OrderDocument.php
@@ -1337,8 +1337,16 @@ abstract class OrderDocument {
 		do_action( 'wpo_wcpdf_after_footer', $this->get_type(), $this->order );
 		return ob_get_clean();
 	}
-	public function footer() {
-		echo wp_kses_post( $this->get_footer() );
+	public function footer(): void {
+		$allowed_tags = wp_kses_allowed_html( 'post' );
+
+		// Allow <style> with type and media attributes.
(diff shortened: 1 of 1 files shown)


MohamadNateqi:
@alexmigf 
The customer issue is that the `wp_kses_post()` will strip out `<style>` tags, and so their table has no style.

We can also remove `wp_kses_post()` and not apply any escape functions to the footer since the website owner sets the footer and has nothing to do with customer input. What do you think?

MohamadNateqi:
@alexmigf It makes sense to consider other approaches for entering styles, but should we permit customers to input their footer styles in the footer field or not?

alexmigf:
@MohamadNateqi No, I think we should suggest alternative methods. This field is intended for text or markup only.

MohamadNateqi:
@alexmigf Okay, I will close this PR as the changes are unnecessary.

PR #1123, opened 2025-04-09
New: Helper function to handle UBL file writing with error handling

### LINKED ISSUE #1122: Create helper function to handle UBL file writing
Extract UBL file writing logic into a new function to centralize the code currently duplicated in two places. Include strict type declarations, basic PHPDoc, and error handling for missing dependencies.

alexmigf:
closes #1122


wpo-ips-functions.php
```diff
@@ -1401,3 +1401,61 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+function wpo_ips_write_ubl_file( object $document, bool $attachment = false, bool $contents_only = false ) {
+	$ubl_maker = wcpdf_get_ubl_maker();
+
+	if ( ! $ubl_maker ) {
+		return wcpdf_error_handling( 'UBL Maker not available. Cannot write UBL file.' );
+	}
+
+	if ( $attachment ) {
+		$tmp_path = WPO_WCPDF()->main->get_tmp_path( 'attachments' );
+		$ubl_maker->set_file_path( $tmp_path );
```
MohamadNateqi, line 1422: Maybe you should check if the `$tmp_path` is not `false`.

wpo-ips-functions.php
```diff
@@ -1401,3 +1401,61 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+	}
+
+	$ubl_document = new \WPO\IPS\UBL\Documents\UblDocument();
+	
+	if ( ! $ubl_document ) {
```
MohamadNateqi, line 1427: I think this condition is always 'false' because the `new` never returns `false`.

wpo-ips-functions.php
```diff
@@ -1401,3 +1401,61 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+		return wcpdf_error_handling( 'UBL Document not available. Cannot write UBL file.' );
+	}
+	
+	$ubl_document->set_order_document( $document );
+
+	$builder = new \WPO\IPS\UBL\Builders\SabreBuilder();
+	
+	if ( ! $builder ) {
```
MohamadNateqi, line 1435: Same

wpo-ips-functions.php
```diff
@@ -1401,3 +1401,61 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+ * @param object $document
+ * @param bool   $attachment
+ * @param bool   $contents_only
+ * 
+ * @return string|bool
+ */
+function wpo_ips_write_ubl_file( object $document, bool $attachment = false, bool $contents_only = false ) {
+	$ubl_maker = wcpdf_get_ubl_maker();
+
+	if ( ! $ubl_maker ) {
```
MohamadNateqi, line 1416: Same

wpo-ips-functions.php
```diff
@@ -1401,3 +1401,61 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+/**
+ * Write UBL file
+ * 
+ * @param object $document
+ * @param bool   $attachment
+ * @param bool   $contents_only
+ * 
+ * @return string|bool
+ */
+function wpo_ips_write_ubl_file( object $document, bool $attachment = false, bool $contents_only = false ) {
```
MohamadNateqi, line 1413: Maybe `OrderDocument` instead of `object` for `$document`?

MohamadNateqi (changes requested):

wpo-ips-functions.php
alexmigf: Of course, my bad.

wpo-ips-functions.php
```diff
@@ -1401,3 +1401,60 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+			return wcpdf_error_handling( 'Temporary path not available. Cannot write UBL file.' );
+		}
+		
+		$ubl_maker->set_file_path( $tmp_path );
+	}
+
+	$ubl_document = new \WPO\IPS\UBL\Documents\UblDocument();
+	$ubl_document->set_order_document( $document );
+
+	$builder = new \WPO\IPS\UBL\Builders\SabreBuilder();
```
BrunoPavlinic98, line 1433: I think you are missing one space before '=' to follow WPCS.

wpo-ips-functions.php
```diff
@@ -1401,3 +1401,60 @@ function wpo_wcpdf_get_latest_plugin_version( string $plugin_slug ) {
...
+		$document->get_filename(
+			'download',
+			array( 'output' => 'ubl' )
+		),
+		$document
+	);
+
+	$full_filename = $ubl_maker->write( $filename, $contents );
+
+	return $full_filename;
```
BrunoPavlinic98, line 1458: Should we use a filter here, and provide `$filename`, `$contents`, `$document`, `$attachment`, and `$contents_only` as additional parameters being passed down to the filter? I am not sure if that makes sense because I lack business logic on how users use their UBL's, but maybe it will be easier for customers to have their custom UBL logic in this hook only, rather then change the UBL content using `wpo_ips_ubl_contents` filter, and then alter the filename using `wpo_ips_ubl_filename` filter. I am not sure if this is a valid use case.

What do you think?

BrunoPavlinic98: Actually, now when I look at it, it does not make much sense to include `$contents_only` because if this is true, we won't even get to that line.

alexmigf: Thank you for your suggestion. I don't think it's particularly important to add a filter here, as we already provide one for the `filename`. This is just the path combined with the `filename`, so I'm not sure it would be very useful at this point. However, we could consider adding it in the future if a customer needs it for a specific use case.

BrunoPavlinic98: @alexmigf Agreed!

BrunoPavlinic98 (approved):
LGTM!

PR #1131, opened 2025-04-11
Fix: "Gapped" invoices shown when other documents used the invoice numbering

### LINKED ISSUE #1130: Advanced > Numbers: Document "Type" is displayed as "gapped"
I'm seeing the "gapped" label within the "Type" column for orders in which the document numbers are correct, which seems to be a bug:

Image

I'm not sure if this is relevant, but please note that I'm formatting the number with a specific prefix, suffix, and padding, like this:

Image

alexmigf:
closes #1130

Invoices were incorrectly marked as "gapped" in **Advanced > Numbers** when other document types reused the invoice numbering system.

### DIFF
--- a/includes/Tables/NumberStoreListTable.php
+++ b/includes/Tables/NumberStoreListTable.php
@@ -64,8 +64,9 @@ class NumberStoreListTable extends \WP_List_Table {
 					$value = $item->id;
 					break;
 				case 'type':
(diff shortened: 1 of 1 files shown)


YordanSoares (approved):
LGTM!

image
_I have edited the invoice numbers102 and 104 and confirmed that the number tool is detecting the changes._

MohamadNateqi (approved):
LGTM!

PR #1137, opened 2025-04-15
AVIF image format support

### LINKED ISSUE #964: AVIF image format support
Currently, neither DOMPDF nor mPDF support AVIF. As a temporary solution, we could consider converting AVIF images to JPG until proper support is implemented.

BrunoPavlinic98:
closes #964


MohamadNateqi (commented):
I added an image using the Customizer custom block:

image

And this is the document PDF:

image

Should this work with remote images? Or is it only a fix for local images?

includes/Documents/OrderDocument.php
```diff
@@ -1464,7 +1464,8 @@ public function get_pdf() {
 			'paper_orientation'	=> apply_filters( 'wpo_wcpdf_paper_orientation', 'portrait', $this->get_type(), $this ),
 			'font_subsetting'	=> $this->get_setting( 'font_subsetting', false ),
 		);
-		$pdf_maker    = wcpdf_get_pdf_maker( $this->get_html(), $pdf_settings, $this );
+		
+		$pdf_maker    = wcpdf_get_pdf_maker( $this->replace_avif_images( $this->get_html() ), $pdf_settings, $this );
```
alexmigf, line 1468: This should be tied to a new advanced setting, as we haven’t received many requests for it and I don’t think it’s justified as a default just yet.

includes/Documents/OrderDocument.php
```diff
@@ -1477,6 +1478,59 @@ public function get_pdf() {
 		return apply_filters( 'wpo_wcpdf_get_pdf', $pdf, $this );
 	}
 
+	/**
+	 * Replace .avif images with .jpg images in HTML
+	 * @param string $html The HTML to process
+	 * 
+	 * @return string The processed HTML
+	 */
+	public function replace_avif_images( string $html ): string {
```
alexmigf, line 1487: This should be placed in the `functions.php` file so it could be reused elsewhere if needed.

MohamadNateqi (commented):
In my test with a local .avif image, I couldn't see the logo:

image

includes/Documents/OrderDocument.php
```diff
@@ -1477,6 +1478,59 @@ public function get_pdf() {
...
+		$replaced_images = array();
+
+		if ( ! empty( $html ) ) {
+			$dom = new \DOMDocument();
+			$dom->loadHTML( $html );
```
alexmigf, line 1492: Maybe you should clear errors here if HTML is malformed, see example: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e8581e1c1484d8cf41e1e4817e2e70b6f322f8e3/wpo-ips-functions.php#L630-L632

includes/Documents/OrderDocument.php
```diff
@@ -1477,6 +1478,59 @@ public function get_pdf() {
...
+						$replaced_images[$src] = substr( $src, 0, -5 ) . '.jpg';
+					}
+				}
+			}
+			
+			if ( ! empty( $replaced_images ) ) {
+				$html = strtr( $html, $replaced_images );
+				foreach ( $replaced_images as $avif_src => $jpg_src ) {
+					// Get the server path from URL
+					if ( ! file_exists( $avif_src ) ) {
```
alexmigf, line 1509: We have a filesystem class now to handle this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/e8581e1c1484d8cf41e1e4817e2e70b6f322f8e3/includes/Compatibility/FileSystem.php#L168-L180

includes/Documents/OrderDocument.php
```diff
@@ -1477,6 +1478,59 @@ public function get_pdf() {
...
+						continue;
+					}
+	
+					// Skip if jpg already exists
+					if ( file_exists( $jpg_src ) ) {
```
alexmigf, line 1514: Same.

includes/Documents/OrderDocument.php
```diff
@@ -1477,6 +1478,59 @@ public function get_pdf() {
...
+					$src = $image->getAttribute( 'src' );
+					
+					if ( '.avif' === substr( $src, -5 ) ) {
+						$replaced_images[$src] = substr( $src, 0, -5 ) . '.jpg';
+					}
+				}
+			}
+			
+			if ( ! empty( $replaced_images ) ) {
+				$html = strtr( $html, $replaced_images );
```
alexmigf, line 1506: Wouldn't be faster and secure to replace the `src` directly on line `1500`?

`$image->setAttribute( 'src', $avif_src );`

includes/Documents/OrderDocument.php
```diff
@@ -1477,6 +1478,59 @@ public function get_pdf() {
...
+					}
+	
+					// Skip if jpg already exists
+					if ( file_exists( $jpg_src ) ) {
+						continue;
+					}
+					
+					// Load the avif image
+					if ( function_exists( 'imagecreatefromavif' ) ) {
+						$image = imagecreatefromavif( $avif_src );
```
alexmigf, line 1520: This is a PHP8+ function, but our plugin is 7.4+, you need to have that into account.

alexmigf: I think GD php extension is also needed.

alexmigf (changes requested):
This function only supports local files, so maybe you should check that when getting the `src` before continue with the code. I think this should be also mentioned in the new setting description.

includes/Documents/OrderDocument.php
BrunoPavlinic98: If I do it like this, I need to return that HTML using `$dom->saveHTML();` method. The problem with this is that `saveHTML` encodes URLs, and puts `%20` signs instead of empty spaces. Due to this, the path to the image is wrong, and no image in the document is being displayed properly. We could potentially replace all `%20`'s with empty space, but then we are back to the approach that is similar to `strtr`.

includes/Documents/OrderDocument.php
BrunoPavlinic98: If you agree, I can create a new setting in the advanced tab, and we can put the beta `sup`  sign there to indicate that this is a beta feature. With that, we can warn the user in the description that this is a PHP8+ function in case they have a lower version than that installed. I am not sure if GD php extension is needed - how do you check for this?

alexmigf: Here: https://www.php.net/manual/en/function.imagecreatefromavif.php (returns GdImage)

Also check this: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/56fb10364614b9e4dc072a5f39b8cf8596feb67e/includes/Settings/SettingsDebug.php#L1078

alexmigf: I agree with your idea for the setting, seems the best approach.

BrunoPavlinic98: @alexmigf This is done, feel free to check the code.

includes/Documents/OrderDocument.php
BrunoPavlinic98: I did not find any satisfactory solution related to `DOMDocument` class. An alternative would be to use some other library, but I think this is an overkill in this case. Our HTML's are not that large when dealing with single documents. The only performance wise issue might occur with bulk documents.

What do you think?

alexmigf: Instead of processing the entire HTML, why not target only the specific locations where an image might be used? In the free plugin, we could start by focusing on the logo, using an abstraction function to handle the AVIF conversion. Inside that function, we could implement multiple conversion methods — for example, if Imagick is available, we could use it for the conversion:

```php
$image = new Imagick('image.avif');
$image->setImageFormat('jpeg');
$image->writeImage('image.jpg');
$image->clear();
$image->destroy();
```

If the user is running PHP 8.1 or higher, we could use `imagecreatefromavif` for the conversion:

```php
$image = imagecreatefromavif('image.avif');
imagejpeg($image, 'image.jpg');
imagedestroy($image);
```

We should also consider adding a setting to enable this feature (using the [beta tag](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/1c8eaf7ad3307b2e179d85f017a63172f414a2cd/views/settings-page.php#L33)), and check its status inside the conversion function. Later on, we could reuse the same function to handle product thumbnails in the Premium Templates.

Let me know your thoughts.

BrunoPavlinic98: That is not a bad approach at all! I will try it, and let you know how it goes.
Thank you!

BrunoPavlinic98: @alexmigf I did an initial abstraction approach. Please check it out and let me know your opinion. If this is OK, than I will continue with other avif > jpg implementations.

alexmigf: @BrunoPavlinic98 When I said abstraction function, I meant something like the function you created earlier in the functions file. Sorry if I wasn’t very clear.

alexmigf: I don't think we need this level of complexity or separate classes for now; a function should be enough, as it's not a big deal at the moment.

BrunoPavlinic98: I understand, but since the code is already there, I think it would be best that I continue on the provided architecture. I don't think the complexity is too large, but helps a lot because it separates different implementations.

If we will ever consider refactoring this, I think we will do it in a similar or the same way as it is currently being done.
What do you say?

alexmigf: Thanks — just to clarify, I’m suggesting we don’t move forward with the structure in the PR (the abstract class and separate implementation), and instead go with a single function. AVIF support isn’t a major feature in the plugin right now, so introducing that level of abstraction feels like overengineering.

If image handling ever becomes more complex or central, we can revisit and design a more suitable structure then — potentially with a dedicated class. But for now, I think a simple function is the more pragmatic choice.

Open to discussing further, of course — just wanted to be clear on my reasoning.

BrunoPavlinic98: I understand.

Having one function is OK, but we need to be aware of the consequences:

1. One function should be responsible for taking avif image HTML, and returning the jpg one - I don't see an issue here.
2. The function from above should serve as a callback for multiple hooks (in this example, I used the logo hook) - this is also fine because more hooks are 'hooked' to one callback function.
3. We have multiple possible implementations on how to create a jpg image from avif - imagecreatefromavif and Imagick are 2 we are aware of - this is not OK because we will need to somehow decide on which one to use in that function, depending on PHP versions and PHP extensions available. This can easily lead to messy code.
4. If we have 2 possible implementations for avif --> jpg problem, I think it is only natural that we try to separate concerns in code, making one function for each approach - this is not bad, but why not have it as a part of an abstract class - it is much cleaner.

With all mentioned above, I think that by using a single function approach, we will naturally loop back because we will notice that we created lots of functions that are just wandering around our code. That is why I think the current approach is better. Regarding the overengineering, my opinion is simple - if you make an abstraction that has only one implementation, that is overengineered. Otherwise, you are using abstraction like you should.

alexmigf: Although I understand your concerns, I still believe a single function is sufficient to handle this for now. The proposed structure would make more sense if we were dealing with something critical like e-invoicing or another accounting-related feature—but that’s not the case with AVIF. This is just an initial step toward compatibility, and creating a full structure for it feels unnecessary and potentially problematic if we later decide to rename or remove it. Deprecating a function is much simpler.

That said, you just need to check whether the conversion can be applied based on the Imagick or PHP version, and then proceed accordingly. If you're concerned about performance, you can cache the result using a transient. I think we’re overengineering a solution that should be simple for this first iteration.

BrunoPavlinic98: I am not concerned about performance because I think performance won't be an issue since we will be using multiple small hooks and we will be dealing with small amount of HTML.
I am concerned that we are removing the abstraction code with the aim of simplification, but in reality, we are just building up a technical debt for this implementation. Sooner or later, we will want to make this abstract, especially if we find additional ways on how we can convert `avif` to `jpg`.

However, let's say that we go in the function approach. One function fill be located in `wpo-ips-functions.php` file, and then we will hook that function to multiple PDF hooks, one of them being `wpo_wcpdf_header_logo_img_element` Where do you propose that we define those hooks? I think the best place for it would be the `AvifCompatibility` class, which will have multiple hooks all pointing to the `wpo_ips_convert_avif_to_jpg` function, that will hold all the logic. Do you think that would be an okay solution?

alexmigf: No, just do like this:

image

BrunoPavlinic98: Done!

BrunoPavlinic98: Please let me know what you think.

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,41 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
 	return $full_filename;
 }
 
+function wpo_ips_convert_avif( string $image_html ): string {
+	$debug_settings = get_option( 'wpo_wcpdf_settings_debug', array() );
```
alexmigf, line 1483: I was thinking of keeping the function focused solely on its responsibilities and leaving the settings outside of it.

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,41 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
 	return $full_filename;
 }
 
+function wpo_ips_convert_avif( string $image_html ): string {
```
alexmigf, line 1482: I would rename this to `wpo_ips_maybe_convert_avif_image`, what you think?

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,41 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+		return $image_html;
+	}
+	
+	$src     = '';
+	$jpg_src = '';
+	
+	if ( preg_match('/src=["\'](.*?)["\']/', $image_html, $matches ) ) {
+		$src = $matches[1];
+		
+		if ( '.avif' === substr( $src, -5 ) && 'http' !== substr( $src, 0, 4 ) ) {
```
alexmigf, line 1494: Maybe you could use `preg_match` here too?

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,41 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+			$jpg_src = substr( $src, 0, -5 ) . '.jpg';
+			if ( WPO_WCPDF()->file_system->exists( $src ) && ! WPO_WCPDF()->file_system->exists( $jpg_src ) ) {
+				if ( version_compare( PHP_VERSION, '8.1.0', '<' ) && extension_loaded( 'gd' ) ) {
```
alexmigf, line 1497: Maybe you should check the `imagecreatefromavif` function here too.

alexmigf, line 1497: Maybe you could wrap both conversions in a try/catch with logging.

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,41 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+						$image->destroy();
+					}
+				} else {
+					return $image_html;
+				}
+			}
+		}
+	}
+	
+	return str_replace('.avif', '.jpg', $image_html);
```
alexmigf, line 1518: Maybe using `preg_replace` would be safer?

alexmigf (changes requested):

wpo-ips-functions.php
BrunoPavlinic98: I agree.

wpo-ips-functions.php
BrunoPavlinic98: It would work of course, but I don't see why we would use this. We always need to replace `'.avif'` with `'.jpg'`, and both are strings, not patterns. That is why I used `str_replace`.

Do you have negative experience with `str_replace`, or did it sometimes worked in a wrong way?

wpo-ips-functions.php
BrunoPavlinic98: Even tough I initially had a similar comment as [here](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1137#discussion_r2074801683), I think that here `preg_match` is OK because we can check the first and last characters with it, making the code a bit smaller, but still readable.

wpo-ips-functions.php
alexmigf: It's up to you, but I think it's more confusing where the function is called, since it's defined in another file and we can’t immediately see whether the setting is being applied.

wpo-ips-functions.php
alexmigf: Your code isn't incorrect, but it may cause unintended replacements. For example:

`logo.avif icon`

wpo-ips-functions.php
BrunoPavlinic98: Done!

wpo-ips-functions.php
BrunoPavlinic98: Can you please explain what would be wrong here? logo.avif would be replaced by logo.jpg on 2 places, but I don't see a major issue here. And I think that with `preg_replace,` we would still be replacing `.avif` with `.jpg`, because replacing the entire `src` attribute might be trickier to achieve via regex - I am not sure how would I achieve this with being 100% sure I did not trim the wrong part of the `src `.

I would like to understand this 100% so that I know this in the future. Thank you.

alexmigf: > Your code isn't incorrect ...

I'm not saying it's wrong—just that you might end up replacing strings you shouldn't.

BrunoPavlinic98: I see. I tried to play with regexes a bit to replace only the src attribute but I was not successful.

We can do the following:

```
	$dom = new \DOMDocument();
	@$dom->loadHTML( mb_convert_encoding( $image_html, 'HTML-ENTITIES', 'UTF-8' ), LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );
	$images = $dom->getElementsByTagName( 'img' );
	if ( $images->length > 0 ) {
		$images->item(0)->setAttribute( 'src', $jpg_src );
		
		$image_html = $dom->saveHTML();
	}
	
	return str_replace( '%20', ' ', $image_html );
```

With the code above, we are sure that we are updating the `SRC` attribute only, but in the end, we need to replace '%20' with an empty space due to `$image_html = $dom->saveHTML();` restrictions.

Do you think the provided way could be better?

BrunoPavlinic98:
@MohamadNateqi, can you please test again with local files only? Currently, we only support conversion for `avif` logo images.

alexmigf:
@BrunoPavlinic98 I made a few small improvements to the AVIF fallback logic you worked on — mainly added support for using an existing JPG if it's already there and resolving remote URLs to local paths when possible.
Your implementation was solid — just extended it a bit to cover more cases. Let me know what you think!

BrunoPavlinic98:
> @BrunoPavlinic98 I made a few small improvements to the AVIF fallback logic you worked on — mainly added support for using an existing JPG if it's already there and resolving remote URLs to local paths when possible. Your implementation was solid — just extended it a bit to cover more cases. Let me know what you think!

@alexmigf I agree with the implementation, thank you for the help. I added one small WPCS tweak to it - other than that, all good.

alexmigf (approved):

MohamadNateqi (changes requested):

includes/Documents/OrderDocument.php
```diff
@@ -1295,6 +1295,9 @@ public function header_logo(): void {
 			$img_src     = isset( WPO_WCPDF()->settings->debug_settings['embed_images'] ) ? wpo_wcpdf_get_image_src_in_base64( $src ) : $src;
 			$img_element = sprintf( '<img src="%1$s" alt="%2$s"/>', wpo_wcpdf_escape_url_path_or_base64( $img_src ), esc_attr( $company ) );
 			$img_element = apply_filters( 'wpo_wcpdf_header_logo_img_element', $img_element, $attachment_id, $this );
+			$img_element = isset( WPO_WCPDF()->settings->debug_settings['avif_fallback'] ) && wc_string_to_bool( WPO_WCPDF()->settings->debug_settings['avif_fallback'] )
```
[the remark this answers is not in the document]
alexmigf: Wouldn't that just skip the AVIF function?

includes/Settings/SettingsDebug.php
```diff
@@ -908,6 +908,26 @@ public function init_settings() {
...
+				'type'     => 'setting',
+				'id'       => 'avif_fallback',
+				'title'    => __( 'Convert AVIF to JPG', 'woocommerce-pdf-invoices-packing-slips' ) . ' <sup class="wcpdf_beta">beta</sup>',
+				'callback' => 'checkbox',
+				'section'  => 'debug_settings',
+				'args'     => array(
+					'option_name' => $option_name,
+					'id'          => 'avif_fallback',
+					'description' => __( 'Enables AVIF fallback support. Currently, this only applies to the company logo: if a local AVIF image is used, it will be replaced with a JPG version in the PDF documents.', 'woocommerce-pdf-invoices-packing-slips' )
+						. ( version_compare( PHP_VERSION, '8.1.0', '<' ) && ! extension_loaded( 'imagick' )
```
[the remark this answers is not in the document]
alexmigf: That's true, but I also wanted it to be properly positioned. Still, you should see the image correctly because the AVIF function has checks, and it falls back to the original if needed. Plus, the PHP version limitation should be mentioned in the description, right?

includes/Documents/OrderDocument.php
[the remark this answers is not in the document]
MohamadNateqi: I'm not sure if I understood your point, but I don't think it creates any issues if it goes before the filter applied on line 1297.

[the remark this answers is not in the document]
alexmigf: If you run the AVIF function before applying the filter, and then change the `img` element with the filter back to AVIF, wouldn't this bypass the AVIF function?

includes/Settings/SettingsDebug.php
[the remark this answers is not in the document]
MohamadNateqi: Yes, I agree to display the info in the description, and it also works fine if the requirements are not met.
I just thought it might enhance the UX, so users can understand why this setting is gray and do not allow them to change it.

includes/Documents/OrderDocument.php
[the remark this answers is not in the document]
MohamadNateqi: Yes, but isn't the purpose of providing that filter to allow modification of any changes to the final result? The bypass can happen, but I think maybe it should be the user's concern?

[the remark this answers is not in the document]
alexmigf: Generally, that’s true, but in this case, it can be misleading. You know you uploaded an AVIF file, but the filter displays it as a JPG. I’m not opposed to this approach, but I believe it could cause more confusion than clarity.

includes/Settings/SettingsDebug.php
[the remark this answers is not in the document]
alexmigf: This is currently in beta, and I don't think it's a solid solution yet. It will likely need further improvements later. For now, I wouldn’t invest more time in it.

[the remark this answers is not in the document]
MohamadNateqi: Thank you for the explanations.

MohamadNateqi (approved):
LGTM!

Issue #1139, opened 2025-04-15
Allow filtering of the default capability to access to the plugin's setting


YordanSoares:
Currently, we have a filter to extend the access to another user roles based on their capabilities ([see](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/da5d0b56384589f04ef8bf02683c86e67b84de8c/includes/Settings.php#L136)), but the `user_settings_capability` function always return `manage_woocommerce` as a default value, if the current user do not have the filtered capabilities.

Adding a new filter to customize the default capability, will allow adding more control on the access to the plugin's settings.

PR #1140, opened 2025-04-15
Improves the settings capability filter: `wpo_wcpdf_settings_user_role_capabilities`

YordanSoares:
Closes: #1139

### DIFF
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -133,8 +133,8 @@ class Settings {
 	 * @return string
 	 */
 	public function user_settings_capability() {
-		$user_capability       = '';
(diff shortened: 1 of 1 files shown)


includes/Settings.php
```diff
@@ -133,8 +133,8 @@ public function add_support_links( $links, $file ) {
 	 * @return string
 	 */
 	public function user_settings_capability() {
-		$user_capability       = 'manage_woocommerce';
-		$capabilities_to_check = apply_filters( 'wpo_wcpdf_settings_user_role_capabilities', array( $user_capability ) );
+		$user_capability       = '';
```
alexmigf, line 136: This will serve as the fallback if capability checks fail. I believe we should keep it as it was and instead add a filter to modify the default if needed.

alexmigf (approved):

PR #1144, opened 2025-04-17
Fix: Restrict edit document number input to valid positive integers

### LINKED ISSUE #1143: Document number field accepts invalid values
The manual document number input allows non-numeric values, which can break numbering logic and lead to inconsistent or conflicting numbers.

Image

alexmigf:
closes #1143

I couldn’t find any historical reason why the invoice number field was set as a text input, nor any indication that switching it to a number input would cause issues. Since the field is meant to accept only positive integers, and allowing arbitrary strings can interfere with the numbering logic, this change helps enforce valid input. I’m not seeing any downside at the moment.


MohamadNateqi (approved):
LGTM!

PR #1147, opened 2025-04-21
Fix: Deprecation warning for `null` value passed to `substr()` when extracting thumbnail URL

### LINKED ISSUE #1146: substr(): Passing null to parameter #1 ($string) of type string is deprecated
> PHP Deprecated:  substr(): Passing null to parameter #1 ($string) of type string is deprecated in wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Documents/OrderDocumentMethods.php on line 915

alexmigf:
closes #1146


MohamadNateqi (approved):
LGTM!

PR #1151, opened 2025-04-22
Tweak: Improve Invoice My Account link logic for custom status checks and code clarity

### LINKED ISSUE #1150: Improve `my_account_pdf_link()` logic for custom status checks and code clarity
The `my_account_pdf_link()` method could be improved by using strict comparisons for order statuses, avoiding redundant `array_keys()` calls when possible, and handling empty custom status settings more defensively. The logic can also be cleaned up for better readability and maintainability.

alexmigf:
closes #1150

### DIFF
--- a/includes/Frontend.php
+++ b/includes/Frontend.php
@@ -19,7 +19,7 @@ class Frontend {
 	}
 
 	public function __construct()	{
-		add_filter( 'woocommerce_my_account_my_orders_actions', array( $this, 'my_account_pdf_link' ), 999, 2 ); // needs to be triggered later because of Jetpack query string: https://github.com/Automattic/jetpack/blob/1a062c5388083c7f15b9a3e82e61fde838e83047/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-my-account.php#L235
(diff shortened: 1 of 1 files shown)


includes/Frontend.php
```diff
@@ -28,41 +28,45 @@ public function __construct()	{
...
+		$invoice         = wcpdf_get_invoice( $order );
+		$invoice_allowed = false;
+		
 		if ( $invoice && $invoice->is_enabled() ) {
-			$pdf_url = WPO_WCPDF()->endpoint->get_document_link( $order, 'invoice', array( 'my-account' => 'true' ) );
-
 			// check my account button settings
 			$button_setting = $invoice->get_setting( 'my_account_buttons', 'available' );
+			
 			switch ( $button_setting ) {
```
MohamadNateqi, line 47: Would it be better to move this `switch` outside of the `if` to have an early return if the `$invoice_allowed` is `false`?

MohamadNateqi (approved):
LGTM!

includes/Frontend.php
alexmigf: This approach aligns with the structure used [here](https://github.com/wpovernight/woocommerce-pdf-ips-pro/commit/931e6d83ead2ae305bf1af0104bbcda6576690fb), and it also avoids duplicating the final filter call.

PR #1156, opened 2025-04-25
Fix: Ensure fonts are properly reinstalled after upgrade when `tmp` folder is cleared

### LINKED ISSUE #1155: Fonts missing after upgrade due to reinstall not completing
When the plugin is upgraded, existing fonts in the temporary folder are correctly deleted, but they are not always reinstalled. This leaves the fonts directory empty, causing blank PDFs to be generated. The issue seems to occur during the reinstall step where font files are expected to be copied from the plugin’s assets directory. It appears that under certain conditions, the reinstall does not complete as intended, and no fonts are restored.

alexmigf:
closes #1155

This issue has been difficult to reproduce locally, but I'm hoping that forcing the font reinstall immediately after the upgrade routine code will resolve the problem.


MohamadNateqi (approved):
LGTM!

PR #1163, opened 2025-04-30
Introduce separated shop address fields

### LINKED ISSUE #1160: Split Seller Address into specific address fields
Currently, we only have a single address field. We need to split it into specific fields: Street 1, Street 2, City, Zip, and Country. This change is necessary to comply with e-invoicing formats and APIs. Default values should be pulled from WooCommerce, but users must be able to modify them, with an option to re-sync with WooCommerce if needed.

Image

MohamadNateqi:
close #1160
This PR should be merged following this PR: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/564.

This PR replaces the single input field for shop addresses with multiple address fields:

image

The setup wizard fields are also updated:

image

<hr>

**Notes**
- The current value of the **Shop Address** field will be set for the new **Shop Additional Info** field. This prevents any issues for the customer's migration to these new changes.
- An admin notice will be displayed to fill in the new shop address fields.
- A sync button will be added in a separate [issue](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/1167).


MohamadNateqi:
@alexmigf I didn’t plan a migration for the address field because customers enter addresses in all sorts of formats—complete, partial, inconsistent—so parsing and restructuring the existing data isn’t straightforward. That said, I can take it on if you feel it’s necessary.

Alternatively, we could just display an admin notice prompting users to update their shipping address in light of the changes. What do you think?

As for "States", I left them out and stuck with just the country, assuming it might simplify things, such as for E-Invoicing or external integrations, to obtain any part of the address they need.
Would you prefer to include states within the country field, add a separate field for them, or just leave them out entirely?

alexmigf:
@MohamadNateqi We could run into an issue if the user saves changes without completing the address fields. Should we consider syncing with WooCommerce in that case?

> Would you prefer to include states within the country field, add a separate field for them, or just leave them out entirely?

It might be worth including a state field, as some European countries like Romania use it. However, since these fields are also used to generate the company address in the PDF, changes could impact how addresses appear for other countries. I recommend adding a separate, manually editable field for this purpose.

MohamadNateqi:
> We could run into an issue if the user saves changes without completing the address fields. Should we consider syncing with WooCommerce in that case?

Do you suggest this only for this version where we are introducing these new changes? Or do you mean syncing address data if it's empty every time the user saves changes?

MohamadNateqi:
@alexmigf I believe some users add additional information to their shop address, such as a company number, VAT, or other details. Should we also introduce an "Additional address details" field or inform them to use existing filters to include their custom data?

MohamadNateqi:
@alexmigf If I understand correctly, you're suggesting changing the name of the **Shop Address** field to **Shop Additional Info** and keeping its value so we don't need to worry about migration. Am I right?

alexmigf:
@MohamadNateqi Yes, and if you continue displaying it alongside the new fields, users who already have a value set won’t notice any change.

We could display a notice to users who have a value in the additional shop info field but haven’t filled out the address fields, so they can update their information accordingly.

includes/Documents/OrderDocument.php
```diff
@@ -1329,8 +1329,8 @@ public function get_settings_text( $settings_key, $default = false, $autop = tru
 		}
 
 		// legacy filters
-		if ( in_array( $settings_key, array( 'shop_name', 'shop_address', 'footer', 'extra_1', 'extra_2', 'extra_3' ) ) ) {
-			$text = apply_filters( "wpo_wcpdf_{$settings_key}", $text, $this );
+		if ( in_array( $settings_key, array( 'shop_name', 'footer', 'extra_1', 'extra_2', 'extra_3' ) ) ) {
+			$text = apply_filters_deprecated( "wpo_wcpdf_{$settings_key}", array( $text, $this ), '4.4.1', "wpo_wcpdf_{$settings_key}_settings_text" );
```
alexmigf, line 1333: Shouldn't be `4.5.0`?

includes/SetupWizard.php
```diff
@@ -278,23 +278,23 @@ public function get_step( $delta ) {
...
+
 				foreach ( $request['wcpdf_settings'] as $option => $settings ) {
 					// sanitize posted settings
 					foreach ( $settings as $key => $value ) {
 						if ( 'attach_to_email_ids' === $key ) {
 							$value = array_fill_keys( $value, '1' );
 						}
 
-						if ( $key == 'shop_address' && function_exists( 'sanitize_textarea_field' ) ) {
+						if ( $key == 'shop_address_additional' && function_exists( 'sanitize_textarea_field' ) ) {
```
alexmigf, line 297: Strict comparison.

views/setup-wizard/shop-name.php
```diff
@@ -5,11 +5,64 @@
...
 <div class="wpo-setup-input">
 	<?php
+	// Set default values for current setting to be used in case the user has not set them yet.
 	$current_settings = wp_parse_args( get_option( 'wpo_wcpdf_settings_general', array() ), array(
-		'shop_name'    => array( 'default' => get_bloginfo( 'name' ) ),
-		'shop_address' => array( 'default' => '' ),
+		'shop_name'             => array( 'default' => get_bloginfo( 'name' ) ),
+		'shop_address_line_1'   => array( 'default' => get_option( 'woocommerce_store_address' ) ),
+		'shop_address_line_2'   => array( 'default' => get_option( 'woocommerce_store_address_2' ) ),
+		'shop_address_country'  => array( 'default' => get_option( 'woocommerce_store_country' ) ?? '' ),
```
alexmigf, line 13: You could use the second parameter of the `get_option()` to set the default value, no need to use `??`.

includes/Settings/SettingsGeneral.php
```diff
@@ -214,17 +217,104 @@ public function init_settings() {
...
+				'id'		=> 'shop_address_country',
+				'title'		=> __( 'Shop Country', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback'	=> 'select',
+				'section'	=> 'general_settings',
+				'args'		=> array(
+					'option_name'      => $option_name,
+					'options_callback' => ( function () {
+						$countries[''] = __( 'Select a country', 'woocommerce-pdf-invoices-packing-slips' );
+						$woo_countries = WC()->countries->get_countries();
+						$woo_countries = array_combine( array_values( $woo_countries ), $woo_countries );
```
alexmigf, line 255: Maybe there is some redundancy here?

alexmigf: Additionally, I don't see any way to clear this input if I decide not to select a country after having already chosen one.

includes/Settings/SettingsGeneral.php
```diff
@@ -454,6 +543,57 @@ public function find_templates() {
...
+			} else {
+				wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_dismiss_requirements_notice' );
+				return;
+			}
+		}
+
+		if (
+			! empty( WPO_WCPDF()->settings->general_settings['shop_address_additional'] ) &&
+			(
+				empty( WPO_WCPDF()->settings->general_settings['shop_address_line_1'] ) ||
```
alexmigf, line 567: This notice is never displayed:

image

includes/Documents/OrderDocument.php
```diff
@@ -1367,11 +1367,91 @@ public function shop_coc_number() {
...
+			$this->get_settings_text( 'shop_address_line_1', '', false),
+			$this->get_settings_text( 'shop_address_line_2', '', false),
+			$this->get_settings_text( 'shop_address_city', '', false),
+			$this->get_settings_text( 'shop_address_postcode', '', false),
+			$this->get_settings_text( 'shop_address_state', '', false),
+			$this->get_settings_text( 'shop_address_country', '', false),
+			$this->get_settings_text( 'shop_address_additional', '', false),
+		) );
+
+		return wpautop( implode( '<br>', $address ) );
```
alexmigf, line 1454: I'm not sure about this. For example, in Portugal, the postcode and city are usually on the same line. Maybe you should use the `get_address_formats()` method from WooCommerce Countries?

alexmigf (changes requested):
I'm not sure about the location of this fields:

image

Comparing to the Shop VAT Number. Maybe it needs an update on that.

includes/Settings/SettingsGeneral.php
MohamadNateqi: > Maybe there is some redundancy here?

The `WC()->countries->get_countries()` returns an array that has country code as keys and country name as its value.
For our settings, the keys and values both should be the country name, so I had to do that.

Do you have any suggestions?

views/setup-wizard/shop-name.php
```diff
@@ -5,11 +5,64 @@
 </div>
 <div class="wpo-setup-input">
 	<?php
+	// Set default values for current setting to be used in case the user has not set them yet.
 	$current_settings = wp_parse_args( get_option( 'wpo_wcpdf_settings_general', array() ), array(
-		'shop_name'    => array( 'default' => get_bloginfo( 'name' ) ),
-		'shop_address' => array( 'default' => '' ),
+		'shop_name'             => array( 'default' => get_bloginfo( 'name' ) ),
+		'shop_address_line_1'   => array( 'default' => get_option( 'woocommerce_store_address' ) ),
```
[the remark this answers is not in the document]
MohamadNateqi: Yeah, my eyes missed them! Strange!

includes/Settings/SettingsGeneral.php
MohamadNateqi: If I understood your point, you mean that instead of using the country name as the array key, we should use the country code and store the country code in the settings as well, while introducing a new function to convert the country code to the country name. Am I right?

includes/Settings/SettingsGeneral.php
```diff
@@ -454,6 +547,57 @@ public function find_templates() {
...
+			} else {
+				wcpdf_log_error( 'You do not have sufficient permissions to perform this action: wpo_dismiss_requirements_notice' );
+				return;
+			}
+		}
+
+		if (
+			! empty( WPO_WCPDF()->settings->general_settings['shop_address_additional']['default'] ) &&
+			(
+				empty( WPO_WCPDF()->settings->general_settings['shop_address_line_1']['default'] ) ||
```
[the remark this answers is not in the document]
MohamadNateqi: What do you think about checking all the keys in this array? This way, we can ensure that we send an admin notice if the address is missing for any languages.

includes/Settings/SettingsGeneral.php
alexmigf: Yes.

includes/Settings/SettingsGeneral.php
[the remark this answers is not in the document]
alexmigf: Yes, maybe it's better.

includes/Documents/OrderDocument.php
```diff
@@ -1364,11 +1364,123 @@ public function shop_coc_number() {
...
+			'state'          => $this->get_settings_text( 'shop_address_state', '', false ),
+			'country_code'   => $this->get_settings_text( 'shop_address_country', '', false ),
+			'additional'     => $this->get_settings_text( 'shop_address_additional', '', false ),
+		);
+		$address['country'] = $this->get_country_from_country_code( $address['country_code'] );
+
+		$address_format = $this->get_address_format_from_country( $address['country_code'] );
+
+		// Replace placeholder with $address values, and remove empty placeholders.
+		$formatted_address = preg_replace_callback( '/\{([a-zA-Z0-9_]+)}/', function ( $matches ) use ( $address ) {
```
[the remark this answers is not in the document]
MohamadNateqi: Do you mean a separate function to replace placeholders with values?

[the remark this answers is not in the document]
alexmigf: I would say this part:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1163/files#diff-18422edbbceedd754958dd82112333b64966cbd87587fd9bb7c09b6043dfe57aR1462-R1481

Maybe it can be placed in the functions file?

MohamadNateqi:
@n-rusu Based on this PR implementation, where do you think I should put the sync button to sync address fields with Woo address fields?

alexmigf:
@MohamadNateqi is this ready for a new review?

MohamadNateqi:
@alexmigf Yes.

alexmigf:
@MohamadNateqi I asked because it appears that the function declaration for `wpo_wcpdf_get_formatted_address` is missing.

MohamadNateqi:
@alexmigf Sorry, I forgot to add that file to my commit. Ready now.

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,59 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+	$formatted_address = preg_replace( '/,\s*,/', ',', $formatted_address );    // Handle repeated commas
+	$formatted_address = preg_replace( '/\n\s*\n/', '\n', $formatted_address ); // Remove empty lines
+
+	// Add additional info if provided.
+	if ( ! empty( $address['additional'] ) ) {
+		$formatted_address .= '\n' . $address['additional'];
+	}
+
+	// Convert to HTML line breaks.
+	return str_replace( '\n', '<br>', $formatted_address );
```
alexmigf, line 1536: Maybe you should also consider `<br/>`

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,59 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+	// Replace placeholder with $address values, and remove empty placeholders.
+	$formatted_address = preg_replace_callback( '/\{([a-zA-Z0-9_]+)}/', function ( $matches ) use ( $address ) {
+		return $address[ $matches[1] ] ?? '';
+	}, $address_format );
+
+	// Remove empty spaces and unnecessary commas.
+	$formatted_address = preg_replace( '/,\s*,/', ',', $formatted_address );    // Replace ", ," with ","
+	$formatted_address = preg_replace( '/,\s*$/', '', $formatted_address );     // Remove trailing commas
+	$formatted_address = preg_replace( '/,\s*,/', ',', $formatted_address );    // Handle repeated commas
+	$formatted_address = preg_replace( '/\n\s*\n/', '\n', $formatted_address ); // Remove empty lines
```
alexmigf, line 1528: Maybe `\n` should be double quoted? `"\n"`

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,59 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+
+	// Add additional info if provided.
+	if ( ! empty( $address['additional'] ) ) {
+		$formatted_address .= '\n' . $address['additional'];
```
alexmigf, line 1532: Same here.

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,59 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+function wpo_wcpdf_get_formatted_address( $country_code, $address ): string {
+	$address_format = wpo_wcpdf_get_address_format_for_country( $country_code );
+
+	// Replace placeholder with $address values, and remove empty placeholders.
+	$formatted_address = preg_replace_callback( '/\{([a-zA-Z0-9_]+)}/', function ( $matches ) use ( $address ) {
+		return $address[ $matches[1] ] ?? '';
+	}, $address_format );
+
+	// Remove empty spaces and unnecessary commas.
+	$formatted_address = preg_replace( '/,\s*,/', ',', $formatted_address );    // Replace ", ," with ","
```
alexmigf, line 1525: Maybe this regex could be improved: `'/,\s*,+/'`

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,59 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+
+/**
+ * Get the formatted address for a given country code.
+ *
+ * @param $country_code
+ * @param $address
+ *
+ * @return string
+ */
+function wpo_wcpdf_get_formatted_address( $country_code, $address ): string {
```
alexmigf, line 1516: Type hints.

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,59 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+ * @return string Country name or empty string if not found.
+ */
+function wpo_wcpdf_get_country_from_country_code( string $country_code ): string {
+	return \WC()->countries->get_countries()[ $country_code ] ?? '';
+}
+
+/**
+ * Get the address format for a given country.
+ *
+ * @param string $country_code Country name, like the Netherlands.
```
alexmigf, line 1496: Like `NL`?

alexmigf (changes requested):

alexmigf (changes requested):
Maybe the Shop Name deserves a small description too?

Screenshot from 2025-05-16 15-41-42

Make the selector full width. This only happens when the multilingual tabs are displayed.

Screenshot from 2025-05-16 15-40-30

wpo-ips-functions.php
```diff
@@ -1493,11 +1493,12 @@ function wpo_wcpdf_get_country_from_country_code( string $country_code ): string
 /**
  * Get the address format for a given country.
  *
- * @param string $country_code Country name, like the Netherlands.
+ * @param string $country_code Country code, like the NL.
  *
  * @return string
  */
 function wpo_wcpdf_get_address_format_for_country( string $country_code ): string {
+	$country_code    = strtoupper( trim( $country_code ) );
```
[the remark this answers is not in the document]
MohamadNateqi: I checked the emptiness of the `$country_code` in the return line, so if it's empty, it'll return the default address format.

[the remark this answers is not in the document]
alexmigf: I see, didn't show up on the review screen, sorry.

alexmigf (commented):
This issue remains for me:

image

MohamadNateqi:
@alexmigf, isn't this coming from cache?

alexmigf:
@MohamadNateqi no.

MohamadNateqi:
@alexmigf Should be fixed now. 
It was fine on Chrome, but I could see the issue after checking it with Firefox.

alexmigf:
@MohamadNateqi brilliant!

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,66 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
 }
 
+/**
+ * Get the country name from the country code.
+ *
+ * @param string $country_code
+ *
+ * @return string Country name or empty string if not found.
+ */
+function wpo_wcpdf_get_country_from_country_code( string $country_code ): string {
```
alexmigf, line 1489: Maybe `wpo_wcpdf_get_country_from_code`?

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,66 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+
+/**
+ * Get the formatted address for a given country code.
+ *
+ * @param string $country_code
+ * @param array $address
+ *
+ * @return string
+ */
+function wpo_wcpdf_get_formatted_address( string $country_code, array $address ): string {
```
alexmigf, line 1518: Maybe `wpo_wcpdf_format_country_address`?

alexmigf (changes requested):

wpo-ips-functions.php
MohamadNateqi: I wondered if the `code` itself might be confusing. Like, what code exactly?
Do you think it's redundant?

alexmigf: I don't see a problem with that. But maybe it should be: `wpo_wcpdf_get_country_name_from_code`

wpo-ips-functions.php
```diff
@@ -1479,3 +1479,66 @@ function wpo_ips_write_ubl_file( \WPO\IPS\Documents\OrderDocument $document, boo
...
+
+/**
+ * Get the formatted address for a given country code.
+ *
+ * @param string $country_code
+ * @param array $address
+ *
+ * @return string
+ */
+function wpo_wcpdf_format_country_address( string $country_code, array $address ): string {
```
alexmigf, line 1518: Since you're passing `country_code` inside the `$address`, maybe you don't need the first parameter? Maybe you could also create a fallback in case the code is not provided? For example, getting the code from the country name.

alexmigf (changes requested):

wpo-ips-functions.php
MohamadNateqi: I added the `$country_code` parameter as part of the design philosophy since it explicitly defines the function's contract. This makes it clear that the address formatting depends on the provided country code, independent of whether it exists in the `$address` array.

Also, the `$address` data might not always include a country code or name, and it may be unreliable, so I thought having this as a required input ensures consistency.

Do you think it's better to remove it?

alexmigf: No, but maybe in that case don't include it in the address.

MohamadNateqi: So in that case, I need to check if the country code is in the placeholders, since it's not included in the `$address` and will be removed.

alexmigf: Not sure if I understood correctly.

MohamadNateqi: If I understand your point correctly, you mentioned removing the country code from the `$address` parameter, right?
In that case, I said, since currently we're only using the data inside this `$address` array to replace placeholders with values, we also need to check the country code separately to replace its placeholder with its value.
I did some changes related to that in my last commit, could you please check?

alexmigf: Can you explain this placeholders and why we need them?

alexmigf: I see. You could add the country code to the address at the beginning of the function if you want.

alexmigf (changes requested):
Something seems off. I have my full address entered, but only the postcode, city, and country are being displayed.

image

Maybe we could promote the Postcode Checker plugin for address validation?

MohamadNateqi:
> Something seems off. I have my full address entered, but only the postcode, city, and country are being displayed.

You mean the address 1, address 2, and the state are not being displayed, right?

> Maybe we could promote the Postcode Checker plugin for address validation?

Since it's the customer's own address, would validation be useful here?

MohamadNateqi:
The state should not be displayed as it's not in the Portugal address format in Woo:

```php
{company}\n
{name}\n
{address_1}\n
{address_2}\n
{postcode} {city}\n
{country}
```

But for the address line 1 and 2, the last commit should resolve the issue. Could you please check again?

> I agree that, in most cases, it's not necessary. However, it could serve as a way to promote the Postcode plugin. Given that the PDF plugin is our most popular product, this approach might make sense. But let me know your opinion.

It would be great if we could promote the Postcode checker here.

What do you think about the implementation? For instance, displaying a message indicating that users can validate the address with the Postcode checker plugin? And ff they have this plugin, a message will appear below one of the fields showing the validation result?

alexmigf:
I agree.

MohamadNateqi:
@alexmigf What are your thoughts on adding a new tab for the Postcode checker in the settings section, along with a tool that allows customers to enter and validate an address? 

I think this would result in a cleaner and more effective implementation. Also, for multilingual setups, displaying the validation results on this page could be challenging.
However, if we create this tool, we could simply include a message below one of the fields to guide users there. 

What do you think?

alexmigf:
Yes, that could work. Please open an issue for it.

MohamadNateqi:
@alexmigf You're right. I created another issue for that: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/issues/1174.

alexmigf (changes requested):
I see a `<br>` before the address line 1, maybe because of the empty company name:

image

Maybe you should consider removing the leading names from the formats.

includes/Documents/OrderDocument.php
```diff
@@ -1364,11 +1364,104 @@ public function shop_coc_number() {
...
+	}
+	function shop_address_line_2(): void {
+		echo esc_html( $this->get_shop_address_line_2() );
+	}
+
+	/**
+	 * Return/Show shop/company address country if provided.
+	 */
+	function get_shop_address_country(): string {
+		return $this->get_country_from_country_code( $this->get_shop_address_country_code() );
```
alexmigf, line 1391: It appears that this function is undefined: `get_country_from_country_code()`

includes/Documents/OrderDocument.php
```diff
@@ -1364,11 +1364,104 @@ public function shop_coc_number() {
...
+
 	/**
 	 * Return/Show shop/company address if provided
 	 */
-	public function get_shop_address() {
-		return $this->get_settings_text( 'shop_address' );
+	public function get_shop_address(): string {
+		$country_code = $this->get_settings_text( 'shop_address_country', '', false );
+		$address      = array(
+			'address_1'  => $this->get_settings_text( 'shop_address_line_1', '', false ),
```
alexmigf, line 1453: Maybe you could use the getters above here?

alexmigf (changes requested):

includes/Documents/OrderDocument.php
MohamadNateqi: Since those getters add a paragraph, I cannot use them here.

alexmigf (approved):
LGTM!

PR #1166, opened 2025-05-01
Tweak: Set PHP file functions as default method in Filesystem class due to WP_Filesystem instability

### LINKED ISSUE #1165: Set default filesystem method to PHP due to instability with WP Filesystem API
Our `Filesystem` class currently supports both PHP native file functions and the WP_Filesystem API. However, we've observed recurring issues when using WP_Filesystem, especially with methods other than `direct` (e.g., `ftp`, `ssh`), including permission errors and inconsistent behavior across environments.

To improve stability and user experience, we should update the default method used by the `Filesystem` class to native PHP file functions. WP_Filesystem will remain available as a fallback or user-selectable option, but only in `direct` mode.

As discussed internally with @DavidAnderson684

alexmigf:
closes #1165

This PR also improves the reliability of the WP_Filesystem initialization process.

### DIFF
--- a/includes/Compatibility/FileSystem.php
+++ b/includes/Compatibility/FileSystem.php
@@ -106,24 +106,22 @@ class FileSystem {
 	 * @return void
 	 */
 	public function initialize_wp_filesystem(): void {
+		// Force the use of the direct filesystem method if not already defined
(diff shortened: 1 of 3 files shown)


MohamadNateqi (approved):
LGTM!

PR #1169, opened 2025-05-02
 Fix: Fatal error in Semaphore class when `wcpdf_log_error()` is undefined

### LINKED ISSUE #1168: Uncaught Error: Call to undefined function `wcpdf_log_error()` in `Semaphore.php:371`
A customer encountered a fatal error after renaming the `woocommerce` plugin directory. The error originates from a call to the `wcpdf_log_error()` function, which appears to be undefined in this context.

```
Fatal error: Uncaught Error: Call to undefined function wcpdf_log_error()  
in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Semaphore.php:371  
Stack trace:  
#0 Semaphore.php(413): WPO\IPS\Semaphore::is_cleanup_scheduled()  
#1 Semaphore.php(436): WPO\IPS\Semaphore::schedule_semaphore_cleanup()  
#2 class-wp-hook.php(324): WPO\IPS\Semaphore::init_cleanup('')  
#3 class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)  
#4 plugin.php(517): WP_Hook->do_action(Array)  
#5 wp-settings.php(704): do_action('init')  
...
```

alexmigf:
closes #1168

### DIFF
--- a/includes/Semaphore.php
+++ b/includes/Semaphore.php
@@ -365,12 +365,17 @@ class Semaphore {
 	 * @return bool - whether the cleanup is scheduled
 	 */
 	public static function is_cleanup_scheduled(): bool {
+		$error_message = 'Action Scheduler is not available. Cannot check if cleanup is scheduled.';
+		
 		if ( function_exists( '\\as_next_scheduled_action' ) ) {
 			return \as_next_scheduled_action( self::get_cleanup_hook_name() );
+		} elseif ( function_exists( 'wcpdf_log_error' ) ) {
+			\wcpdf_log_error( $error_message, 'critical' );
 		} else {
-			\wcpdf_log_error( 'Action Scheduler is not available. Cannot check if cleanup is scheduled.', 'critical' );
-			return false;
+			error_log( $error_message );
 		}
+		
+		return false;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/Semaphore.php
```diff
@@ -365,12 +365,17 @@ private static function get_cleanup_hook_name(): string {
...
+		$error_message = 'Action Scheduler is not available. Cannot check if cleanup is scheduled.';
+		
 		if ( function_exists( '\\as_next_scheduled_action' ) ) {
 			return \as_next_scheduled_action( self::get_cleanup_hook_name() );
+		} elseif ( function_exists( 'wcpdf_log_error' ) ) {
+			\wcpdf_log_error( $error_message, 'critical' );
 		} else {
-			\wcpdf_log_error( 'Action Scheduler is not available. Cannot check if cleanup is scheduled.', 'critical' );
-			return false;
+			error_log( $error_message );
```
MohamadNateqi, line 375: Maybe we could have a log method for this?

MohamadNateqi (approved):
LGTM!

includes/Semaphore.php
alexmigf: The class already has one, but I didn't want to mix them because these are static methods that might be called earlier.

MohamadNateqi: All good then.

PR #1170, opened 2025-05-05
Introduce categories to General settings

### LINKED ISSUE #1164: Add categories to general settings
We can add categories to the general settings, the same as we did for document settings.

MohamadNateqi:
close #1164


alexmigf (changes requested):
Can't close the sections.

image

MohamadNateqi:
@alexmigf 

> Can't close the sections.

Could you please clear your cache and try again? I cannot replicate this.

alexmigf:
I've tried it, but it's still not closing. It seems like the event isn't being attached properly.

image

MohamadNateqi:
@alexmigf Are the accordions in the Documents tab also not working?

MohamadNateqi:
This is really strange! I tried it with three browsers, and it worked fine!

alexmigf:
I fixed the issue with the settings accordion not working on the first load of the General tab. The problem was that the script wasn't attaching the event handlers because the tab parameter was missing in the URL, causing the function to exit early.

I adjusted the code to default to the General tab when the tab parameter is absent. This way, the accordion works properly, even on the first load.

MohamadNateqi:
I see that now. I always had the tab parameter in the URL.

Thank you for fixing this.

alexmigf (approved):
LGTM!

MohamadNateqi:
@alexmigf All good. Let's roll.

Issue #1179, opened 2025-05-26
Fix Shop Address not displaying


alexmigf:
The shop address is not displaying in documents after last version v4.5.0.

PR #1180, opened 2025-05-26
Fix: Shop Address not displaying in documents

alexmigf:
#1179

### DIFF
--- a/includes/Documents/OrderDocument.php
+++ b/includes/Documents/OrderDocument.php
@@ -1448,10 +1448,13 @@ abstract class OrderDocument {
 	 * Return/Show shop/company address if provided
 	 */
 	public function get_shop_address(): string {
-		if ( $this->use_historical_settings() ) {
-			return $this->get_settings_text( 'shop_address', '', false );
+		// Legacy compatibility until "Test mode" is enabled
+		$address = $this->get_settings_text( 'shop_address' );
+		if ( $this->use_historical_settings() && ! empty( $address ) ) {
+			return $this->get_settings_text( $address );
 		}
 		
+		// New
 		$country_code = $this->get_settings_text( 'shop_address_country', '', false );
 		$address      = array(
 			'address_1'  => $this->get_settings_text( 'shop_address_line_1', '', false ),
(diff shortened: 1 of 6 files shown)


includes/Install.php
```diff
@@ -614,6 +614,16 @@ protected function upgrade( $installed_version ) {
 				update_option( 'wpo_wcpdf_settings_general', $general_settings );
 			}
 		}
+		
+		// 4.5.1-beta.1: set shop address value for the new shop additional info field.
+		if ( version_compare( $installed_version, '4.5.1-beta.1', '<' ) ) {
+			$general_settings = get_option( 'wpo_wcpdf_settings_general', array() );
+
+			if ( ! empty( $general_settings['shop_address_additional'] ) ) {
+				$general_settings['shop_address'] = $general_settings['shop_address_additional'];
```
MohamadNateqi, line 623: I don't understand the reason for having this. Could you please explain it to me?
As I understand, for documents that use historical data, the fallback code added to the `get_shop_address()` should fix the issue, and for new documents, it uses the `shop_address_additional` key.

alexmigf: Because you removed that setting here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/94bd11274e4803c40a8311caa67e0cd5ad4282ba/includes/Install.php#L613

alexmigf: Which fallback?

alexmigf: If that setting is removed on the [migration](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/94bd11274e4803c40a8311caa67e0cd5ad4282ba/includes/Install.php#L613), it will be empty on that code right?

alexmigf: Ah, maybe not, let me confirm.

includes/Documents/OrderDocument.php
```diff
@@ -1448,6 +1448,13 @@ function shop_address_additional(): void {
 	 * Return/Show shop/company address if provided
 	 */
 	public function get_shop_address(): string {
+		// Ensures legacy compatibility until settings are saved after upgrading from version 4.5.0
```
MohamadNateqi, line 1451: Maybe needs to be updated.

MohamadNateqi (approved):
LGTM!

includes/Documents/OrderDocument.php
MohamadNateqi: I just wanted to point out that `until settings are saved after upgrading from version 4.5.0` may need to be removed because it was added for the migration, right?

PR #1188, opened 2025-05-27
Fix: Support for `{state_code}` in shop address formatting

### LINKED ISSUE #1187: Add support for `{state_code}` placeholder in shop address formatting
The `{state_code}` placeholder is used in some WooCommerce address formats but is not currently supported in our plugin shop address. We should resolve the correct ISO code from the state name and fall back when necessary.

alexmigf:
closes #1187

### DIFF
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -1520,42 +1520,33 @@ function wpo_wcpdf_format_country_address( string $country_code, array $address 
 
 	// Set default values for address fields if not provided.
 	$address['country_code'] = $address['country_code'] ?? $country_code;
+	$address['state']        = $address['state'] ?? '';
 
-	$state_name = $address['state'] ?? '';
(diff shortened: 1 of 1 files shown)


wpo-ips-functions.php
```diff
@@ -1521,26 +1521,44 @@ function wpo_wcpdf_format_country_address( string $country_code, array $address
...
 
-	// Replace placeholder with $address values, and remove empty placeholders.
+	$state_name = $address['state'] ?? '';
+	$states     = WC()->countries->get_states( $country_code );
+
+	// Reverse lookup: get state code from name, fallback to state name if not found
+	if ( is_array( $states ) ) {
+		$state_code = array_search( $state_name, $states, true );
+		if ( false === $state_code ) {
+			$state_code = $state_name;
```
MohamadNateqi, line 1531: Maybe you can add this state related values to the `$address` array?

PR #1198, opened 2025-05-29
Tweak: Update UBL tax filters to append values instead of overriding defaults

### LINKED ISSUE #1197: Ensure UBL tax filters append instead of override default values
Currently, the UBL tax-related filters allow full override of the default values when used. To ensure consistency and maintain the integrity of the default tax codes, we should update these filters to append additional values instead of replacing the defaults.

alexmigf:
closes #1197


alexmigf:
Using the filters incorrectly could lead to invalid or non-compliant XML, which might have serious consequences for the user. That’s why I believe we need to be strict here. Customers rely on us to provide accurate and standards-compliant data—they shouldn’t have to worry about whether core tax values were accidentally modified.

I’d rather limit flexibility in this case than risk incorrect submissions that undermine the trust in the software.

MohamadNateqi (approved):
LGTM!

PR #1208, opened 2025-06-10
New: Document data editing for Invoices and Credit Notes is now disabled by default. Use Advanced settings to enable.

### LINKED ISSUE #1207: Document number field on order page does not allow letters
The document number field on the order page no longer accepts letters, following a [previous change](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1144) made to avoid numbering issues and improve consistency. However, this limitation prevents users from using alphanumeric formats, which are still needed in some cases.

alexmigf:
closes #1207

This PR adds a new advanced setting for "Enable document data editing", which will be disabled by default.

image

When disabled:

image

When enabled:

image


includes/Settings/SettingsDebug.php
```diff
@@ -866,6 +866,19 @@ public function init_settings() {
 					),
 				)
 			),
+			array(
+				'type'     => 'setting',
+				'id'       => 'enable_alphanumeric_number',
```
MohamadNateqi: Maybe `enable_alphanumeric_document_number` would be better?

includes/Admin.php
```diff
@@ -1342,11 +1346,16 @@ public function process_order_document_form_data( $form_data, $document_slug )
...
 		if ( isset( $form_data['_wcpdf_' . $document_slug . '_number'] ) ) {
-			$document_number = absint( $form_data['_wcpdf_' . $document_slug . '_number'] );
+			$document_number = $form_data['_wcpdf_' . $document_slug . '_number'];
 			
-			if ( $document_number !== 0 ) {
-				$data['number'] = $document_number;
-			}			
+			if ( ! isset( WPO_WCPDF()->settings->debug_settings['enable_alphanumeric_number'] ) ) {		
+				$document_number = absint( $document_number );	
+				if ( $document_number !== 0 ) {
```
BrunoPavlinic98: I think that you need to use yoda condition here - `0 !== $document_number`

alexmigf: We are just doing it for strings.

MohamadNateqi:
@alexmigf Why not display two inputs, one to edit the plain number and one to edit the formatted number, without any settings?

includes/Admin.php
```diff
@@ -948,17 +948,49 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
-						</p>
+					<?php if ( ! empty( WPO_WCPDF()->settings->debug_settings['enable_document_data_editing'] ) || ! in_array( $document->get_type(), array( 'invoice', 'credit-note' ) ) ) : ?>
+						<?php if ( isset( $data['number'] ) ) : ?>
+							<p class="form-field <?php echo esc_attr( $data['number']['name'] ); ?>_field">
+								<label><?php printf( esc_html__( '%s plain number:', 'woocommerce-pdf-invoices-packing-slips' ), $document->get_title() ); ?></label>
```
MohamadNateqi: Why `for` attribute has been removed?

includes/Admin.php
```diff
@@ -948,17 +948,49 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+												esc_attr( $document->get_title() )
+											),
+											sprintf(
+												/* translators: %1$s: open anchor tag, %2$s: close anchor tag, %3$s: setting name */
+												esc_html__( 'If you need to enable this feature, you can do so in the %1$sAdvanced Settings%2$s section under %3$s.', 'woocommerce-pdf-invoices-packing-slips' ),
+												'<a href="' . esc_url( admin_url( 'admin.php?page=wpo_wcpdf_options_page&tab=debug' ) ) . '" target="_blank">',
+												'</a>',
+												'<strong>' . esc_html__( 'Enable document data editing', 'woocommerce-pdf-invoices-packing-slips' ) . '</strong>'
+											),
+											esc_html__( 'Please ensure that enabling this feature complies with the legal requirements of your country.', 'woocommerce-pdf-invoices-packing-slips' )
```
MohamadNateqi: What do you think about removing this sentence?

includes/Admin.php
```diff
@@ -948,17 +948,49 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
-						</p>
+					<?php if ( ! empty( WPO_WCPDF()->settings->debug_settings['enable_document_data_editing'] ) || ! in_array( $document->get_type(), array( 'invoice', 'credit-note' ) ) ) : ?>
+						<?php if ( isset( $data['number'] ) ) : ?>
+							<p class="form-field <?php echo esc_attr( $data['number']['name'] ); ?>_field">
+								<label><?php printf( esc_html__( '%s plain number:', 'woocommerce-pdf-invoices-packing-slips' ), $document->get_title() ); ?></label>
+								<input type="number" min="1" step="1" class="short" name="<?php echo esc_attr( $data['number']['name'] ); ?>" id="<?php echo esc_attr( $data['number']['name'] ); ?>" value="<?php echo absint( $data['number']['plain'] ); ?>" disabled="disabled">
+							</p>
+							<p class="form-field <?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number'; ?>_field">
+								<label><?php printf( esc_html__( '%s formatted number:', 'woocommerce-pdf-invoices-packing-slips' ), $document->get_title() ); ?></label>
```
MohamadNateqi: Adding `for` attribute would be good.

MohamadNateqi (changes requested):

includes/Admin.php
alexmigf: Can you explain?

MohamadNateqi: I thought it might not be our job to inform them to ensure compliance before enabling. We have already informed them about the compliance concern by disabling the feature. The rest is on them.

includes/Admin.php
```diff
@@ -948,17 +948,53 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+					<?php else : ?>
+						<div class="notice notice-warning inline" style="margin:0;">
+							<p>
+								<?php
+									echo wp_kses_post(
+										sprintf(
+											'%s %s',
+											sprintf(
+												/* translators: %s document title */
+												esc_html__( 'Editing of %s data is currently disabled to comply with legal requirements in some countries.', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi: What are your thoughts on modifying the **data** in the sentence to be more specific to the setting names? For instance:
"Editing of Invoice number and date is currently disabled..."
Customers can already edit and save the note. This could confuse unfamiliar users, leading them to believe they cannot edit the note when that option is disabled. They may also think that there is no option to edit the invoice number and date.

MohamadNateqi (approved):
LGTM!

MohamadNateqi (approved):

PR #1211, opened 2025-06-12
Add "Email Address" to the Shop Address fields

### LINKED ISSUE #1209: Add Email Address field to Shop Address fields
We need to add a new field to the shop address fields to get the shop email address.

MohamadNateqi:
close #1209


templates/Simple/packing-slip.php
```diff
@@ -144,4 +148,4 @@
 	</htmlpagefooter><!-- required for mPDF engine -->
 <?php endif; ?>
 
-<?php do_action( 'wpo_wcpdf_after_document', $this->get_type(), $this->order ); ?>
\ No newline at end of file
+<?php do_action( 'wpo_wcpdf_after_document', $this->get_type(), $this->order ); ?>
```
alexmigf: Maybe this doesn't need to be changed?

templates/Simple/invoice.php
```diff
@@ -191,4 +195,4 @@
 	</htmlpagefooter><!-- required for mPDF engine -->
 <?php endif; ?>
 
-<?php do_action( 'wpo_wcpdf_after_document', $this->get_type(), $this->order ); ?>
\ No newline at end of file
+<?php do_action( 'wpo_wcpdf_after_document', $this->get_type(), $this->order ); ?>
```
alexmigf: Same.

alexmigf (changes requested):
Maybe we should check if the email is valid somehow?

templates/Simple/packing-slip.php
MohamadNateqi: It was from the editorconfig setting, automatically applied. I'll revert.

MohamadNateqi:
> Maybe we should check if the email is valid somehow?

@alexmigf  I added the input type to the field. While it's not a strong validation, I can implement backend validation if you believe it's necessary. Since this field is intended for admin use, I thought front-end validation might suffice. Please let me know your thoughts!

alexmigf:
@MohamadNateqi While it doesn't cause issues with PDF rendering, it can lead to problems with e-invoicing—especially when integrating with Peppol or using validation tools.

includes/Settings/SettingsCallbacks.php
```diff
@@ -734,39 +734,42 @@ public function normalize_settings_args( $args ) {
...
 		foreach ( $input as $key => $value ) {
-
-			// Check to see if the current option has a value. If so, process it.
-			if ( isset( $input[$key] ) ) {
-				if ( is_array( $input[$key] ) ) {
-					foreach ( $input[$key] as $sub_key => $sub_value ) {
-						$output[$key][$sub_key] = $input[$key][$sub_key];
+			if ( is_array( $value ) ) {
+				foreach ( $value as $sub_key => $sub_value ) {
+					if ( 'shop_email_address' === $key ) {
```
alexmigf: Maybe use `strpos` to catch only `email`?

alexmigf (approved):
LGTM!

PR #1212, opened 2025-06-13
New: Improve document data editing on the order page

### LINKED ISSUE #1210: Improve editing of document numbers on the order page
@Terminator-Barbapapa  suggested enhancing the number editing by positioning the prefix and suffix inputs next to the plain number field. Upon saving, each component is stored separately, and the formatted number is generated automatically.

Image

@n-rusu Please also share your feedback on the user interface.

alexmigf:
closes #1210

This PR improves the document number and date editing experience on the order page. Instead of editing the formatted number directly, the UI now provides separate input fields for prefix, number, suffix, and date, along with a live preview of the final formatted result.

image


assets/js/order-script.js
```diff
@@ -207,5 +198,53 @@ jQuery( function( $ ) {
 			$( '.view-more' ).show();
 		}
 	} );
+	
+	function updatePreviewNumber( $table ) {
```
BrunoPavlinic98, line 219: Use `let` instead of `var` in this function.

includes/Admin.php
```diff
@@ -1380,9 +1562,54 @@ public function debug_enabled_warning( $wp_admin_bar ) {
 			$wp_admin_bar->add_node( $args );
 		}
 	}
+	
+	/**
+	 * AJAX handler to preview a formatted number
+	 * 
+	 * @return void
+	 */
+	public function ajax_preview_formatted_number(): void {
```
BrunoPavlinic98, line 1577: Do we need a nonce check here?

includes/Admin.php
```diff
@@ -1380,9 +1562,54 @@ public function debug_enabled_warning( $wp_admin_bar ) {
...
+	}
+
+	/**
+	 * Process the order document form data and return an array with the data to be saved.
+	 *
+	 * @param array $form_data The form data submitted via AJAX.
+	 * @param \WPO\IPS\Documents\OrderDocument $document The document object.
+	 * @return array Processed data ready to be saved.
+	 */
+	public function process_order_document_form_data( array $form_data, \WPO\IPS\Documents\OrderDocument $document ): array {
```
BrunoPavlinic98: Before, we used the document slug as the second parameter. Therefore, we would need to change that function call in every plugin this function is being used. Maybe this is not an ideal solution since currently, I receive the following error when creating a new order:

```
Fatal error: Uncaught Error: WPO\IPS\Admin::process_order_document_form_data(): Argument #2 ($document) must be of type WPO\IPS\Documents\OrderDocument, string given, called in C:\Users\Bruno\Desktop\WP Overnight\Sites\WPOvernight\app\public\wp-content\plugins\woocommerce-order-proposal\compatibility\class-woocommerce-pdf-invoices.php on line 108
in C:\Users\Bruno\Desktop\WP Overnight\Sites\WPOvernight\app\public\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 1612
```

I propose that we leave the old `process_order_document_form_data` function as it was for other plugins, and only use the new function with the document as the second parameter in this plugin.

alexmigf: Nice catch! This needs to be updated on the other plugins, I'm on it.

n-rusu (approved):

includes/Admin.php
BrunoPavlinic98: Changes on other plugins look great, we only need to pay attention that after the new release we put the `WPO_WCPDF` version to 4.6.1! Currently, we have '4.6.0-rc.4'.

assets/js/order-script.js
```diff
@@ -207,5 +198,54 @@ jQuery( function( $ ) {
...
+			},
+			error: function( xhr, status, error ) {
+				console.error( 'AJAX error:', status, error );
+				$table.find( '.preview-number code' ).text( 'Error loading preview' );
+			}
+		} );
+	}
+	
+	$( document ).on( 'input change', '.wcpdf-data-fields .number-specs-table input, .wcpdf-data-fields .number-table input', function() {
+		var $table = $( this ).closest( '.wcpdf-data-fields' );
```
BrunoPavlinic98: Sorry, I forgot to mention this var -> let previously.

includes/Admin.php
```diff
@@ -1380,9 +1562,58 @@ public function debug_enabled_warning( $wp_admin_bar ) {
...
-	public function process_order_document_form_data( $form_data, $document_slug )
-	{
+	/**
+	 * Process the order document form data and return an array with the data to be saved.
+	 *
+	 * @param array $form_data The form data submitted via AJAX.
+	 * @param \WPO\IPS\Documents\OrderDocument $document The document object.
+	 * @return array Processed data ready to be saved.
+	 */
+	public function process_order_document_form_data( array $form_data, \WPO\IPS\Documents\OrderDocument $document ): array {
```
BrunoPavlinic98: When I create a new order and when invoice gets automatically created, I don't have an invoice number. Then, I tried to delete all post meta related to the invoice and created the invoice manually (by clicking on the invoice link in the order edit page), and then, the invoice number is there. Is this an expected behaviour?

I did some debugging, and it seems that `$form_data` does not hold all necessary information related to document number.

includes/Admin.php
alexmigf: No, it's correct, next will be `4.6.0`.

BrunoPavlinic98: All good, you are correct.

includes/Admin.php
alexmigf: @BrunoPavlinic98 can you show me your document settings?

BrunoPavlinic98: [invoice-settings-export_2025-06-17_13-33-07.json](https://github.com/user-attachments/files/20776344/invoice-settings-export_2025-06-17_13-33-07.json)
@alexmigf I exported them.

alexmigf: I haven’t been able to reproduce the issue locally using Bruno’s settings. We’ve scheduled to continue investigating tomorrow.

BrunoPavlinic98: @alexmigf , I think I realized what is going on.

When I create a new order, the `process_order_document_form_data` method gets executed from the `save_invoice_number_date` method. The issue is that `save_invoice_number_date` will stop with the execution in case the `$document_data` is empty:
```
if ( empty( $document_data ) ) {
    return;
}
```
However, that is never the case since the `process_order_document_form_data` will always return some data due to the following lines:
```
$data['number']['document_type'] = $document->get_type();
$data['number']['order_id']      = $document->order->get_id();
```

Therefore, the invoice gets saved.

alexmigf: @BrunoPavlinic98 The function `save_invoice_number_date` is not called when you create a new order.

dpeyou (approved):
## Success
- Works with Invoice, Packing Slip, Proforma, Receipt and Credit Note :heavy_check_mark: 

## Suggestions
- Add hash to link using the HTML id of the input field. The clicked link then goes directly to the setting.

add ink to advanced settings permalink


add hash to link

add hash to link2

- Add warning/note & link to "Next document number" setting.

alexmigf:
@dpeyou I've implemented your first suggestion, but I'm not sure I understand the second one.

includes/Settings/SettingsDebug.php
```diff
@@ -877,8 +877,10 @@ public function init_settings() {
...
 					'description' => __( 'Allow editing of document number and date on the order page.', 'woocommerce-pdf-invoices-packing-slips' ) . '<br>' .
 									sprintf(
-										__( '%s: Changing document data is prohibited in some countries. This setting is disabled by default to comply with legal requirements. Only the following document types are affected: Invoice and Credit Note.', 'woocommerce-pdf-invoices-packing-slips' ),
-										'<strong>' . __( 'Note', 'woocommerce-pdf-invoices-packing-slips' ) . '</strong>'
+										/* translators: %1$s: note, %2$s: document types */
+										__( '%1$s: Changing document data is prohibited in some countries. This setting is disabled by default to comply with legal requirements. Only the following document types are affected: %2$s.', 'woocommerce-pdf-invoices-packing-slips' ),
+										'<strong>' . __( 'Note', 'woocommerce-pdf-invoices-packing-slips' ) . '</strong>',
+										'<code>' . __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' ) . '</code>' . ( function_exists( 'WPO_WCPDF_Pro' ) ? ', <code>' . __( 'Credit Note',  ) . '</code>' : 'woocommerce-pdf-invoices-packing-slips' )
```
YordanSoares: Currently, if you don't have Professional activated, it returns the slug at the end of the description, like this:

image

Also, the `'Credit Note'` string is not available in the free plugin, so I would update this string with something like this:
```php
'<code>' . __( 'Invoice', 'woocommerce-pdf-invoices-packing-slips' ) . '</code>' . ( function_exists( 'WPO_WCPDF_Pro' ) ? ', <code>' . __( 'Credit Note', 'wpo_wcpdf_pro' ) . '</code>' : '' )
```

includes/Admin.php
```diff
@@ -805,21 +807,97 @@ public function data_input_box_content( $post_or_order_object ) {
...
+					'name'      => "_wcpdf_{$document->slug}_date",
+				);
+			}
+		} else {
+			$number_store    = $document->get_sequential_number_store();
+			$number_settings = $document->get_number_settings();
+			
+			$current['number'] = array(
+				'prefix' => array(
+					'value' => $number_settings['prefix'] ?: null,
```
YordanSoares: I'm seeing these warnings that appear above the documents data panel, when the document doesn't exist yet:

```
Warning: Undefined array key "prefix" in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 864

Warning: Undefined array key "suffix" in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 872

Warning: Undefined array key "padding" in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 876
```

image

YordanSoares (changes requested):
@alexmigf I have tested the editing of both invoice and credit note numbers, and it worked as expected:

image

However, I found some issue that I'm pointing below.

includes/Admin.php
```diff
@@ -1380,9 +1602,58 @@ public function debug_enabled_warning( $wp_admin_bar ) {
...
-	public function process_order_document_form_data( $form_data, $document_slug )
-	{
+	/**
+	 * Process the order document form data and return an array with the data to be saved.
+	 *
+	 * @param array $form_data The form data submitted via AJAX.
+	 * @param OrderDocument $document The document object.
+	 * @return array Processed data ready to be saved.
+	 */
+	public function process_order_document_form_data( array $form_data, OrderDocument $document ): array {
```
YordanSoares: I'm getting this fatal error while processing a new refund, when having Professional `2.16.7` also installed:
```php
Fatal error: Uncaught TypeError: WPO\IPS\Admin::process_order_document_form_data(): Argument #2 ($document) must be of type WPO\IPS\Documents\OrderDocument, string given, called in \wp-content\plugins\woocommerce-pdf-ips-pro\includes\wcpdf-pro-writepanels.php on line 299 and defined in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php:1656 Stack trace: 
#0 \wp-content\plugins\woocommerce-pdf-ips-pro\includes\wcpdf-pro-writepanels.php(299): WPO\IPS\Admin->process_order_document_form_data(Array, 'packing_slip') 
#1 \wp-includes\class-wp-hook.php(324): WPO\WC\PDF_Invoices_Pro\Writepanels->save_numbers_dates(Array, Object(Automattic\WooCommerce\Admin\Overrides\Order), Object(WPO\IPS\Admin)) 
#2 \wp-includes\class-wp-hook.php(348): WP_Hook->apply_filters('', Array) 
#3 \wp-includes\plugin.php(517): WP_Hook->do_action(Array) 
#4 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php(1341): do_action('wpo_wcpdf_on_sa...', Array, Object(Automattic\WooCommerce\Admin\Overrides\Order), Object(WPO\IPS\Admin)) 
#5 \wp-includes\class-wp-hook.php(324): WPO\IPS\Admin->save_invoice_number_date(87, Object(Automattic\WooCommerce\Admin\Overrides\Order)) 
#6 \wp-includes\class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array) 
#7 \wp-includes\plugin.php(517): WP_Hook->do_action(Array) 
#8 \wp-content\plugins\woocommerce\includes\admin\class-wc-admin-meta-boxes.php(262): do_action('woocommerce_pro...', 87, Object(WP_Post)) 
#9 \wp-includes\class-wp-hook.php(326): WC_Admin_Meta_Boxes->save_meta_boxes(87, Object(WP_Post)) 
#10 \wp-includes\class-wp-hook.php(348): WP_Hook->apply_filters('', Array) 
#11 \wp-includes\plugin.php(517): WP_Hook->do_action(Array) 
#12 \wp-includes\post.php(5110): do_action('save_post', 87, Object(WP_Post), true) 
#13 \wp-includes\post.php(5212): wp_insert_post(Array, false, true) 
#14 \wp-admin\includes\post.php(453): wp_update_post(Array) 
#15 \wp-admin\post.php(227): edit_post() 
#16 {main} thrown in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 1656
```

alexmigf: Sorry, it requires this Pro branch: https://github.com/wpovernight/woocommerce-pdf-ips-pro/pull/587

wpo-ips-functions.php
```diff
@@ -1604,3 +1604,117 @@ function ( $matches ) use ( $address ) {
...
+ * @param int|null                         $plain_number The base document number (unformatted).
+ * @param string|null                      $prefix       The prefix string (may contain placeholders).
+ * @param string|null                      $suffix       The suffix string (may contain placeholders).
+ * @param int|null                         $padding      Number of digits for zero-padding the base number.
+ * @param \WPO\IPS\Documents\OrderDocument $document     The document object (e.g. invoice or credit note).
+ * @param \WC_Abstract_Order               $order        The WooCommerce order associated with the document.
+ *
+ * @return string The fully formatted document number.
+ */
+function wpo_wcpdf_format_document_number( ?int $plain_number, ?string $prefix, ?string $suffix, ?int $padding, \WPO\IPS\Documents\OrderDocument $document, \WC_Abstract_Order $order ): string {
```
YordanSoares, line 1628: I'm getting this error now:
```
Fatal error: Uncaught TypeError: wpo_wcpdf_format_document_number(): Argument #4 ($padding) must be of type ?int, string given, called in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 888 and defined in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\wpo-ips-functions.php:1628
Stack trace: 
#0 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php(888): wpo_wcpdf_format_document_number(3, '', '', '', Object(WPO\IPS\Documents\Invoice), Object(Automattic\WooCommerce\Admin\Overrides\Order)) A
#1 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php(945): WPO\IPS\Admin->get_current_values_for_document_data(Object(WPO\IPS\Documents\Invoice), Array) 
#2 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php(803): WPO\IPS\Admin->output_number_date_edit_fields(Object(WPO\IPS\Documents\Invoice), Array) 
#3 \wp-admin\includes\template.php(1456): WPO\IPS\Admin->data_input_box_content(Object(WP_Post), Array) 
#4 \wp-admin\edit-form-advanced.php(721): do_meta_boxes(Object(WP_Screen), 'normal', Object(WP_Post)) 
#5 \wp-admin\post.php(206): require('C:\\Users\\Yordan...') 
#6 {main} thrown in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\wpo-ips-functions.php on line 1628
```

YordanSoares (changes requested):

wpo-ips-functions.php
alexmigf: Fixed

YordanSoares (approved):
@alexmigf LGTM!

dpeyou (approved):
I was not able to reproduce the other issues :heavy_check_mark: 
It seems good to go :rocket: :guitar:

My other suggestion was probably not necessary and could potentially lead to bloat.
It was to add a link to the "Next document number" setting. 

There is already a link to the documentation for it however.

assets/js/order-script.js
```diff
@@ -86,25 +86,16 @@ jQuery( function( $ ) {
...
-			if ( $( this ).attr( 'name' ).endsWith( '_formatted_number_current' ) ) {
-				const name  = $( this ).attr( 'name' );
-				const value = $( this ).val();
-				serialized += '&' + encodeURIComponent( name ) + '=' + encodeURIComponent( value );
-			}
-		} );
-
 		// regenerate specific
-		if( action == 'regenerate' ) {
+		if ( action == 'regenerate' ) {
```
BrunoPavlinic98: === and the yoda condition

assets/js/order-script.js
```diff
@@ -86,25 +86,16 @@ jQuery( function( $ ) {
...
 			if ( window.confirm( wpo_wcpdf_ajax.confirm_regenerate ) === false ) {
 				return; // having second thoughts
 			}
 
 			$form.find('.wpo-wcpdf-regenerate-document').addClass('wcpdf-regenerate-spin');
 
 		// delete specific
-		} else if( action == 'delete' ) {
+		} else if ( action == 'delete' ) {
```
BrunoPavlinic98: === and the yoda condition

includes/Admin.php
```diff
@@ -1089,17 +1310,18 @@ public function save_invoice_number_date( $order_id, $order ) {
...
 
-			$form_data = [];
-
-			if ( $invoice = wcpdf_get_invoice( $order ) ) {
-				$is_new        = false === $invoice->exists();
+			$form_data = array();
+			$invoice   = wcpdf_get_invoice( $order );
+			
+			if ( $invoice ) {
+				$is_new        = ( false === $invoice->exists() );
```
BrunoPavlinic98, line 1279: `$invoice->exists()` should return a `bool`, so I think we are good if we only have `$is_new  =  $invoice->exists();`
However, the `$is_new` is only used once (in the `if` loop below) in the `save_invoice_number_date` method, so I think we don't need a variable for holding this value.

includes/Admin.php
```diff
@@ -1286,26 +1508,26 @@ public function ajax_crud_document() {
...
 					$document->delete();
 
 					$response = array(
 						'message' => $notice_messages[$notice]['success'],
 					);
 
 				// on save
-				} elseif ( $action_type == 'save' ) {
+				} elseif ( 'save' === $action_type ) {
 					$is_new = false === $document->exists();
```
BrunoPavlinic98: Same as the comment above, and it is also used only once in this method.

includes/Admin.php
```diff
@@ -1431,6 +1737,7 @@ public function process_order_document_form_data( $form_data, $document_slug )
 			$data['date'] = current_time( 'timestamp', true );
 		}
 
+		// Notes
 		if ( isset( $form_data['_wcpdf_' . $document_slug . '_notes'] ) ) {
 			// allowed HTML
 			$allowed_html = array(
```
BrunoPavlinic98, line 1716: I am not sure if we could use the `wp_kses_post` here instead of defining the `$allowed_html`, I did not test it, but this might be a better option.

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,159 @@
...
+		foreach ( $data as $key => $value ) {
+			if ( in_array( $key, $numeric_properties, true ) ) {
+				$value = (int) $value;
 
-		} elseif ( is_array( $number ) ) {
-			// loaded with full number data
-			foreach ($number as $key => $value) {
-				$this->{$key} = $value;
+				// Only treat 0 as null for numeric keys
+				if ( $value === 0 ) {
```
BrunoPavlinic98, line 70: 0 === $value

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,159 @@
...
 		}
 	}
 
-	public function __toString() {
+	/**
+	 * Returns the formatted string representation of the object.
+	 *
+	 * @return string
+	 */
+	public function __toString(): string {
```
BrunoPavlinic98, line 84: Do we need to cast the return value here? `get_formatted` should always return `string`.

includes/Settings/SettingsDebug.php
```diff
@@ -877,8 +877,10 @@ public function init_settings() {
 					'id'          => 'enable_document_data_editing',
 					'description' => __( 'Allow editing of document number and date on the order page.', 'woocommerce-pdf-invoices-packing-slips' ) . '<br>' .
 									sprintf(
-										__( '%s: Changing document data is prohibited in some countries. This setting is disabled by default to comply with legal requirements. Only the following document types are affected: Invoice and Credit Note.', 'woocommerce-pdf-invoices-packing-slips' ),
-										'<strong>' . __( 'Note', 'woocommerce-pdf-invoices-packing-slips' ) . '</strong>'
+										/* translators: %1$s: note, %2$s: document types */
+										__( '%1$s: Changing document data is prohibited in some countries. This setting is disabled by default to comply with legal requirements. Only the following document types are affected: %2$s.', 'woocommerce-pdf-invoices-packing-slips' ),
```
BrunoPavlinic98: The description says that only invoice and credit note documents are affected, but I can change numbers on proposals and proforma invoices as well. Is this a typo?

wpo-ips-functions.php
```diff
@@ -1604,3 +1604,117 @@ function ( $matches ) use ( $address ) {
...
+			continue;
+		}
+
+		$value = str_replace( '[order_year]', $order_year, $value );
+		$value = str_replace( '[order_month]', $order_month, $value );
+		$value = str_replace( '[order_day]', $order_day, $value );
+		$value = str_replace( "[{$document->slug}_year]", $document_year, $value );
+		$value = str_replace( "[{$document->slug}_month]", $document_month, $value );
+		$value = str_replace( "[{$document->slug}_day]", $document_day, $value );
+		$value = str_replace( '[order_number]', $order_number, $value );
```
BrunoPavlinic98, line 1682: Maybe something like this would be better? I did not test this, but it makes sense to use `str_replace` only once.

```
$replacements = [
    '[order_year]' => $order_year,
    '[order_month]' => $order_month,
    '[order_day]' => $order_day,
    "[{$document->slug}_year]" => $document_year,
    "[{$document->slug}_month]" => $document_month,
    "[{$document->slug}_day]" => $document_day,
    '[order_number]' => $order_number,
];

$value = str_replace( array_keys( $replacements ), array_values( $replacements ), $value );
```

BrunoPavlinic98 (changes requested):
The code review is below.
Functionality wise, I did not encounter on any issue.

includes/Admin.php
alexmigf: This isn’t related to any recent changes I’ve made.

includes/Documents/DocumentNumber.php
alexmigf: Is good.

includes/Settings/SettingsDebug.php
alexmigf: I don’t think that’s the case, as it’s stated to be disabled by default for invoices and credit notes.

BrunoPavlinic98 (approved):
LGTM!

n-rusu (approved):

includes/Admin.php
```diff
@@ -805,21 +807,97 @@ public function data_input_box_content( $post_or_order_object ) {
...
+	/**
+	 * Returns the current values for the document data.
+	 * 
+	 * @param OrderDocument $document The document instance.
+	 * @param array $data The data to be processed.
+	 * 
+	 * @return array The current values for the document data.
+	 */
+	public function get_current_values_for_document_data( OrderDocument $document, array $data ): array {
+		$document_number_instance = $document->get_number();
```
MohamadNateqi: These two can move inside the if condition.

includes/Admin.php
```diff
@@ -933,86 +1013,191 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
-								<input type="text" class="short" name="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number'; ?>" id="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number'; ?>" value="<?php echo esc_attr( $data['number']['formatted'] ); ?>" disabled="disabled">
-								<input type="hidden" class="short" name="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number_current'; ?>" id="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number_current'; ?>" value="<?php echo esc_attr( $data['number']['formatted'] ); ?>"> (<?php esc_html_e( 'Displayed in the PDF', 'woocommerce-pdf-invoices-packing-slips' ); ?>)
-							</p>
+				<div class="editable editable-number-date">
+					<?php if ( $document_data_editing_enabled ) : ?>
+						<?php if ( ! empty( $data['number'] ) ) : ?>
+							<table class="number-specs-table">
+								<thead>
+									<tr>
+										<th class="table-column-head <?php echo esc_attr( $data['number']['prefix']['name'] ); ?>_field">
```
MohamadNateqi: Is this `table-column-head` necessary? Since this element can be accessed through its parent (`thead`). Same for `table-column-data`.

includes/Admin.php
```diff
@@ -805,21 +807,97 @@ public function data_input_box_content( $post_or_order_object ) {
...
+				'plain' => array(
+					'value' => $number_store->get_next() ?? null,
+					'name'  => "_wcpdf_{$document->slug}_number_plain",
+				),
+				'suffix' => array(
+					'value' => $number_settings['suffix'] ?? null,
+					'name'  => "_wcpdf_{$document->slug}_number_suffix",
+				),
+				'padding' => array(
+					'value' => $number_settings['padding'] ?? null,
```
MohamadNateqi: When the document does not exist, the default value for this field will be `0`, also, the min value of the field is `1`.
So, maybe `?: 1` would be better?

includes/Admin.php
```diff
@@ -933,86 +1013,191 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+														__( 'If set, this value will be used as number prefix.' , 'woocommerce-pdf-invoices-packing-slips' ),
+														sprintf(
+															/* translators: 1. document type, 2-3 placeholders */
+															__( 'You can use the %1$s year and/or month with the %2$s or %3$s placeholders respectively.', 'woocommerce-pdf-invoices-packing-slips' ),
+															__( 'invoice', 'woocommerce-pdf-invoices-packing-slips' ),
+															'<strong>[invoice_year]</strong>',
+															'<strong>[invoice_month]</strong>'
+														)
+													);
+													echo wc_help_tip( wp_kses_post( $tip_text ), true );
```
MohamadNateqi: Tips are not being displayed after I save the edit form and open it again. I also noticed a new `data-tip` attr being added.

includes/Admin.php
```diff
@@ -933,86 +1013,191 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
-										);
-									?>
-								</label>
-								<input type="text" class="short" name="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number'; ?>" id="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number'; ?>" value="<?php echo esc_attr( $data['number']['formatted'] ); ?>" disabled="disabled">
-								<input type="hidden" class="short" name="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number_current'; ?>" id="<?php echo '_wcpdf_' . esc_attr( $document->slug ) . '_formatted_number_current'; ?>" value="<?php echo esc_attr( $data['number']['formatted'] ); ?>"> (<?php esc_html_e( 'Displayed in the PDF', 'woocommerce-pdf-invoices-packing-slips' ); ?>)
-							</p>
+				<div class="editable editable-number-date">
+					<?php if ( $document_data_editing_enabled ) : ?>
+						<?php if ( ! empty( $data['number'] ) ) : ?>
+							<table class="number-specs-table">
```
MohamadNateqi: I noticed that the TABLE element is being used to structure data that isn’t actually tabular. For better semantics and accessibility, it’s generally best to use TABLE only for actual tabular data. In cases like this, CSS layout methods might be more appropriate.
Since this is already implemented, I’m not suggesting a rewrite, but it could be worth considering for the future. Let me know your thoughts.

includes/Settings/SettingsDebug.php
MohamadNateqi: Maybe the first placeholder does not need to be in the translatable string.

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,159 @@
...
+	public function __construct( $number, array $settings = array(), ?\WPO\IPS\Documents\OrderDocument $document = null, ?\WC_Abstract_Order $order = null ) {
+		$number = apply_filters( 'wpo_wcpdf_raw_document_number', $number, $settings, $document, $order );
 
-	/**
-	 * Number prefix
-	 * @var string
-	 */
-	public $prefix;
+		// Normalize data from either a raw number or a full array
+		$data = is_array( $number ) ? $number : ( ! empty( $number ) ? array_merge( array( 'number' => $number ), $settings ) : array() );
```
MohamadNateqi: Could this be hard to read?

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,159 @@
...
-			$padding_string = (string) absint($this->padding);
+	/**
+	 * Applies formatting to the document number based on the settings and order/document data.
+	 *
+	 * @param \WPO\IPS\Documents\OrderDocument $document
+	 * @param \WC_Abstract_Order $order
+	 * @return string
+	 */
+	public function apply_formatting( \WPO\IPS\Documents\OrderDocument $document, \WC_Abstract_Order $order ): string {
+		if ( empty( $document ) || empty( $order ) ) {
```
MohamadNateqi: This if-statement can be removed.

includes/Admin.php
```diff
@@ -933,86 +1013,191 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+									<tr>
+										<td class="table-column-data <?php echo esc_attr( $data['number']['plain']['name'] ); ?>_field">
+											<input type="number" min="1" step="1" class="short" name="<?php echo esc_attr( $data['number']['plain']['name'] ); ?>" id="<?php echo esc_attr( $data['number']['plain']['name'] ); ?>" value="<?php echo absint( $data['number']['plain']['value'] ); ?>" disabled="disabled">
+										</td>
+									</tr>
+								</tbody>
+								<tfoot>
+									<tr>
+										<td colspan="3">
+											<label><?php esc_html_e( 'Preview:', 'woocommerce-pdf-invoices-packing-slips' ); ?></label>
```
MohamadNateqi: Maybe P tag? Since the LABEL should come with a form field.

assets/js/order-script.js
```diff
@@ -207,5 +198,54 @@ jQuery( function( $ ) {
...
+				}
+			},
+			error: function( xhr, status, error ) {
+				console.error( 'AJAX error:', status, error );
+				$table.find( '.preview-number code' ).text( 'Error loading preview' );
+			}
+		} );
+	}
+	
+	$( document ).on( 'input change', '.wcpdf-data-fields .number-specs-table input, .wcpdf-data-fields .number-table input', function() {
```
MohamadNateqi: Using both `input` and `change` will make twice AJAX request.

assets/js/order-script.js
```diff
@@ -207,5 +198,54 @@ jQuery( function( $ ) {
...
+		let $table = $( this ).closest( '.wcpdf-data-fields' );
+		updatePreviewNumber( $table );
```
MohamadNateqi: Any change in the fields having a spinner (like "Number padding") will send an AJAX request.
So, if the user holds the mouse down on the spinner, multiple requests will be sent.
Perhaps you could debounce the AJAX call and display a loading spinner beside it?

includes/Admin.php
```diff
@@ -933,86 +1013,191 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+							</table>
+							<table class="number-table">
+								<thead>
+									<tr>
+										<th class="table-column-head <?php echo esc_attr( $data['number']['plain']['name'] ); ?>_field">
+											<label for="<?php echo esc_attr( $data['number']['plain']['name'] ); ?>">
+												<?php
+													printf(
+														/* translators: %s document title */
+														esc_html__( '%s number', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi: Would it be useful to add a tip text beside this to warn about changing this field?

assets/js/order-script.js
```diff
@@ -207,5 +198,54 @@ jQuery( function( $ ) {
...
+				if ( response.success && response.data.formatted ) {
+					let $preview = $table.find( '.preview-number' );
+					let $code    = $preview.find( 'code' );
+					let current  = $preview.data( 'current' );
+					let updated  = response.data.formatted;
+
+					$code.text( updated );
+
+					if ( current !== updated ) {
+						$code.css( 'background-color', '#fff3cd' );
```
MohamadNateqi: It might be better to add a class to the element and define the color in the CSS stylesheet.

assets/js/order-script.js
```diff
@@ -207,5 +198,54 @@ jQuery( function( $ ) {
...
+					} else {
+						$code.css( 'background-color', '' ); // reset
+					}
+				}
+			},
+			error: function( xhr, status, error ) {
+				console.error( 'AJAX error:', status, error );
+				$table.find( '.preview-number code' ).text( 'Error loading preview' );
```
MohamadNateqi: Should the text be translatable?

includes/Admin.php
```diff
@@ -1071,9 +1275,26 @@ public function bulk_actions( $actions ) {
...
+		if (
+			( empty( $order ) || ! ( $order instanceof \WC_Order || is_subclass_of( $order, '\WC_Abstract_Order' ) ) ) &&
+			! empty( $order_id )
+		) {
 			$order = wc_get_order( $order_id );
-		} else {
+		}
+		
+		if (
+			! $order ||
```
MohamadNateqi: Can be removed.

includes/Admin.php
```diff
@@ -1029,10 +1214,29 @@ public function output_number_date_edit_fields( $document, $data ): void {
 					</div>
 					<!-- Editable -->
 					<div class="editable-notes">
-						<p class="form-field form-field-wide">
-							<label for="<?php echo esc_attr( $data['notes']['name'] ); ?>"><?php echo wp_kses_post( $data['notes']['label'] ); ?></label>
-							<p><textarea name="<?php echo esc_attr( $data['notes']['name'] ); ?>" class="<?php echo esc_attr( $data['notes']['name'] ); ?>" cols="60" rows="5" disabled="disabled"><?php echo wp_kses_post( $data['notes']['value'] ); ?></textarea></p>
-						</p>
+						<table class="notes-table">
```
MohamadNateqi: I believe the previous version without the TABLE structure was simpler and also correct in terms of HTML semantics.

includes/Admin.php
```diff
@@ -1071,9 +1275,26 @@ public function bulk_actions( $actions ) {
...
+		// Skip any auto-draft or draft request
+		if (
+			( isset( $_POST['original_post_status'] ) && in_array( $_POST['original_post_status'], array( 'auto-draft', 'draft' ), true ) ) ||
+			( isset( $_POST['post_status'] ) && in_array( $_POST['post_status'], array( 'auto-draft', 'draft' ), true ) )
+		) {
+			return;
+		}
+		
+		if (
+			( empty( $order ) || ! ( $order instanceof \WC_Order || is_subclass_of( $order, '\WC_Abstract_Order' ) ) ) &&
```
MohamadNateqi: This line of conditions could simplify to this: `$order instanceof \WC_Order`. The rest seems unnecessary.

includes/Admin.php
```diff
@@ -1380,9 +1609,86 @@ public function debug_enabled_warning( $wp_admin_bar ) {
...
+
+	/**
+	 * Process the order document form data and return an array with the data to be saved.
+	 *
+	 * @param array $form_data The form data submitted via AJAX.
+	 * @param string|OrderDocument $document The document object. It accepted a document type string before 4.6.0.
+	 * @return array Processed data ready to be saved.
+	 */
+	public function process_order_document_form_data( array $form_data, $document ): array {
+		if ( ! $document instanceof OrderDocument ) {
```
MohamadNateqi, line 1623: Maybe you could introduce a new method, such as `is_valid_document_object()`, and move this if statement code there, as it appears unrelated to this method, and also we can reuse that method somewhere else if needed in the future.

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,159 @@
...
 	}
 
-	public function __toString() {
+	/**
+	 * Returns the formatted string representation of the object.
+	 *
+	 * @return string
+	 */
+	public function __toString(): string {
 		return (string) $this->get_formatted();
```
MohamadNateqi: Type cast can be removed now.

wpo-ips-functions.php
```diff
@@ -1604,3 +1604,117 @@ function ( $matches ) use ( $address ) {
...
+	} elseif ( ! empty( $padding ) ) {
+		$padding_string = (string) $padding;
+	}
+
+	if ( ! empty( $padding_string ) ) {
+		$plain_number = sprintf( '%0' . $padding_string . 'd', $plain_number );
+	}
+
+	// Add prefix & suffix
+	return $formats['prefix'] . $plain_number . $formats['suffix'];
```
MohamadNateqi, line 1719: Wouldn't it be better to apply `wpo_wcpdf_format_document_number` filter here? So, if this function is directly called anywhere else, the filter is also applied.

includes/Admin.php
```diff
@@ -805,21 +807,97 @@ public function data_input_box_content( $post_or_order_object ) {
...
-				'minute'    => $document->exists() && ! empty( $document->get_date() ) ? $document->get_date()->date_i18n( 'i' ) : date_i18n( 'i' ),
+	/**
+	 * Returns the current values for the document data.
+	 * 
+	 * @param OrderDocument $document The document instance.
+	 * @param array $data The data to be processed.
+	 * 
+	 * @return array The current values for the document data.
+	 */
+	public function get_current_values_for_document_data( OrderDocument $document, array $data ): array {
```
MohamadNateqi, line 818: Maybe this function could be improved and refactored slightly.
This is what I had in mind: You can first get the raw values for both number and date, based on the document's existence, and then loop over each raw values to create the `$current` array.
What do you think?

MohamadNateqi (changes requested):

includes/Admin.php
alexmigf: I don't think it's necessary, and I'm not sure why you think it's unrelated.

wpo-ips-functions.php
alexmigf: I didn’t want to change the filter parameters because the second parameter is the `DocumentNumber` instance.

includes/Admin.php
MohamadNateqi: I think it would be more in line with SRP and provide more clarity, which is why I suggested it; however, the current implementation is also fine.

includes/Admin.php
alexmigf: I get what you mean, but the document might not exist, and we still need to build the `$current` array in those cases. It’s working as is, so I’d prefer to leave it for now and revisit later if we want to improve it further.

includes/Admin.php
```diff
@@ -933,86 +1016,159 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+													esc_html__( '%s number', 'woocommerce-pdf-invoices-packing-slips' ),
+													esc_html( $document->get_title() )
+												);
+											?>
+										</label>
+										<input type="number" min="1" step="1" class="short" name="<?php echo esc_attr( $data['number']['plain']['name'] ); ?>" id="<?php echo esc_attr( $data['number']['plain']['name'] ); ?>" value="<?php echo absint( $data['number']['plain']['value'] ); ?>" disabled="disabled">
+									</div>
+									<div class="field-group">
+										<label><?php esc_html_e( 'Formatted number', 'woocommerce-pdf-invoices-packing-slips' ); ?></label>
+										<input type="text" class="formatted-number keep-disabled" data-current="<?php echo esc_html( $data['number']['formatted']['value'] ); ?>" value="<?php echo esc_html( $data['number']['formatted']['value'] ); ?>" disabled="disabled">
```
MohamadNateqi: It seems like it's always a non-interactive element, maybe SPAN would be better?

includes/Admin.php
```diff
@@ -933,86 +1016,159 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+									</div>
+									<div class="field-group placeholder"></div> <!-- Empty cell -->
+									<div class="row-note">
+										<?php echo wp_kses_post( sprintf(
+											/* translators: %1$s: open anchor tag, %2$s: close anchor tag */
+											__( 'Manually changing the document\'s plain number also requires updating the next document number in the %1$sdocument settings%2$s.' ),
```
MohamadNateqi, line 1142: Perhaps adding a `may` word would be good, so they understand it's not mandatory. I mean like this:
"...plain number **may** also **require** updating the next..."

includes/Documents/Invoice.php
```diff
@@ -402,8 +402,17 @@ public function get_pdf_settings_fields( $option_name ) {
...
+					'description' => sprintf(
+							/* translators: %1$s: open anchor tag, %2$s: close anchor tag */
+							__( 'For more information about setting up the number format and see the available placeholders for the prefix and suffix, check this article: %1$sNumber format explained%2$s', 'woocommerce-pdf-invoices-packing-slips' ),
+							'<a href="https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/number-format-explained/" target="_blank">',
+							'</a>'
+						) . '.<br><br>'. sprintf(
+							/* translators: %1$s: strong Note tag, %2$s: document type */
+							__( '%1$s: Changes made to the number format will only be reflected on new orders. Also, if you have already created a custom %2$s number format with a filter, the above settings will be ignored.', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi: Extracting the "Note" from the translatable string?

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,156 @@
...
+	public ?int $padding;
 
 	/**
-	 * Document number formatted for display
-	 * @var String
+	 * Document number constructor.
+	 *
+	 * @param mixed                                 $number   Raw number value or full number array.
+	 * @param array                                 $settings Additional settings used when building the number (e.g. prefix, suffix, padding).
+	 * @param \WPO\IPS\Documents\OrderDocument|null $document Optional related document object.
```
MohamadNateqi: The "\WPO\IPS\Documents\" part can be removed.

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,156 @@
...
+	 * @param \WC_Abstract_Order|null               $order    Optional related order object.
 	 */
-	public $formatted_number;
+	public function __construct( $number, array $settings = array(), ?\WPO\IPS\Documents\OrderDocument $document = null, ?\WC_Abstract_Order $order = null ) {
```
MohamadNateqi: Same about "\WPO\IPS\Documents\"

includes/Documents/DocumentNumber.php
```diff
@@ -9,197 +9,156 @@
...
+		);
 
-	public function __construct( $number, $settings = array(), $document = null, $order = null ) {
-		$number = apply_filters( 'wpo_wcpdf_raw_document_number', $number, $settings, $document, $order );
-		if ( !is_array( $number ) && !empty( $number ) ) {
-			// we're creating a new number with settings as passed
-			$this->number = $number;
+		foreach ( $data as $key => $value ) {
+			if ( in_array( $key, $numeric_properties, true ) ) {
+				$value = (int) $value;
```
MohamadNateqi: Would it be better to use `absint()`?

MohamadNateqi (changes requested):

includes/Admin.php
alexmigf: It’s not mandatory, but it’s recommended. When you change the number manually, the next number won’t update automatically.

MohamadNateqi: Yes, that's why I recommended including a "may" — to avoid implying it's mandatory.

alexmigf: I disagree; I believe we should encourage that change rather than making it optional.

MohamadNateqi (approved):
Great job!

Terminator-Barbapapa:
I ran into a few issues when testing the feature:

- After setting or editing a document number and/or date, the date picker no longer shows when trying to edit the document data for a second time. Only after reloading the page the date picker shows up again.
- When you try to edit existing document data and the 'Enable document data editing' option is disabled, the invoice number is removed if you click on 'Save changes'.
<br>Scherm­afbeelding 2025-07-01 om 12 35 43<br>
Scherm­afbeelding 2025-07-01 om 12 35 20<br><br>
- When changing the order status on the order edit page for an order that doesn't yet have an invoice, an invoice is generated without an invoice number or date. The behavior is not affected by the order status you switch to, nor by the 'Attach to' setting (i.e., it makes no difference whether emails are selected there or not).
<br>Scherm­afbeelding 2025-07-01 om 12 37 10<br><br>
- When using ‘Set invoice number & date’ in the PDF document data panel, it shows the next invoice number by default, which can cause problems. 
<br>Scherm­afbeelding 2025-07-01 om 12 45 28<br>
Users might not realize that manually setting an invoice number does not update the next invoice number, despite the warning shown below the field. If they leave the default value unchanged, it can result in duplicate invoice numbers, as the next automatically generated invoice will use the same number. That’s why I believe it’s better to leave this field empty (or set it to zero), as we did previously.
<br>Scherm­afbeelding 2025-07-01 om 12 44 38<br>
 This encourages users to consciously choose an appropriate invoice number when entering it manually.

includes/Admin.php
```diff
@@ -933,86 +1023,159 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+									<div class="field-group">
+										<label><?php esc_html_e( 'Formatted number', 'woocommerce-pdf-invoices-packing-slips' ); ?></label>
+										<input type="text" class="formatted-number" data-current="<?php echo esc_html( $data['number']['formatted']['value'] ); ?>" value="<?php echo esc_html( $data['number']['formatted']['value'] ); ?>" readonly>
+									</div>
+									<div class="field-group placeholder"></div> <!-- Empty cell -->
+									<div class="row-note">
+										<?php echo wp_kses_post( sprintf(
+											/* translators: %1$s: open anchor tag, %2$s: close anchor tag */
+											__( 'Manually changing the document\'s plain number also requires updating the next document number in the %1$sdocument settings%2$s.' ),
+											'<a href="' . esc_url( admin_url( 'admin.php?page=wpo_wcpdf_options_page&tab=documents&section=' . $document->get_type() ) ) . '#next_invoice_number" target="_blank">',
```
MohamadNateqi: It's not always invoice for other documents in `#next_invoice_number`

includes/Admin.php
```diff
@@ -805,48 +807,134 @@ public function data_input_box_content( $post_or_order_object ) {
...
+		if ( ! isset( $current['number'] ) ) {
+			$number_settings = $document->get_number_settings();
+			
+			$current['number'] = array(
+				'prefix' => array(
+					'value' => $number_settings['prefix'] ?? null,
+					'name'  => "{$name_prefix}number_prefix",	
+				),
+				'plain' => array(
+					'value' => 0,
```
MohamadNateqi, line 874: Should it be allowed to save 0 as the document plain number? Or should we have a backend validation to prevent that?

includes/Admin.php
alexmigf: Good catch!

includes/Admin.php
alexmigf: It should be fine since this is only stored at the order meta level.

includes/Admin.php
```diff
@@ -933,86 +1023,159 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+											<?php esc_html_e( 'Number prefix', 'woocommerce-pdf-invoices-packing-slips' ); ?>
+											<?php
+												$tip_text = sprintf(
+													'%s %s',
+													__( 'If set, this value will be used as number prefix.' , 'woocommerce-pdf-invoices-packing-slips' ),
+													sprintf(
+														/* translators: 1. document type, 2-3 placeholders */
+														__( 'You can use the %1$s year and/or month with the %2$s or %3$s placeholders respectively.', 'woocommerce-pdf-invoices-packing-slips' ),
+														__( 'invoice', 'woocommerce-pdf-invoices-packing-slips' ),
+														'<strong>[invoice_year]</strong>',
```
MohamadNateqi: This also needs to be updated to be based on the document slug.

includes/Admin.php
```diff
@@ -933,86 +1023,159 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
+										<label for="<?php echo esc_attr( $data['number']['prefix']['name'] ); ?>">
+											<?php esc_html_e( 'Number prefix', 'woocommerce-pdf-invoices-packing-slips' ); ?>
+											<?php
+												$tip_text = sprintf(
+													'%s %s',
+													__( 'If set, this value will be used as number prefix.' , 'woocommerce-pdf-invoices-packing-slips' ),
+													sprintf(
+														/* translators: 1. document type, 2-3 placeholders */
+														__( 'You can use the %1$s year and/or month with the %2$s or %3$s placeholders respectively.', 'woocommerce-pdf-invoices-packing-slips' ),
+														__( 'invoice', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi: Same. Also for other fields.

PR #1218, opened 2025-06-30
Proposal email attachment not working if PRO is not installed

### LINKED ISSUE #1217: Proposal email attachment not working if PRO is not installed
[HelpScout ticket](https://secure.helpscout.net/conversation/2964863985/72512)

Currently, users that purchased the order proposal plugin and only have the free PDF plugin are not able to attach proposals in emails because the free plugin blocks them.

The related code is located in `includes\Main.php`, in the `get_documents_for_email` function:
```
// Pro not activated, only attach Invoice
if ( ! function_exists( 'WPO_WCPDF_Pro' ) && 'invoice' !== $document->get_type() ) {
	continue;
};
```

We need to allow proposals as attachments in case the customer has order proposal plugin active.

BrunoPavlinic98:
closes #1217

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -285,10 +285,18 @@ class Main {
 		$attach_documents = array();
 
 		foreach ( $documents as $document ) {
-			// Pro not activated, only attach Invoice
-			if ( ! function_exists( 'WPO_WCPDF_Pro' ) && 'invoice' !== $document->get_type() ) {
+			/**
+			 * Filter documents for email attachment based on plugin availability.
+			 * 
+			 * - If Pro plugin is not active, only allow 'invoice' and 'proposal' document types
+			 * - 'proposal' document type is only available if the Order Proposal plugin is active
+			 * - Skip documents that don't meet these criteria
+			 */
+			if ( 
+				( ! function_exists( 'WPO_WCPDF_Pro' ) && ! in_array( $document->get_type(), array( 'invoice', 'proposal' ) ) ) ||
+				( 'proposal' === $document->get_type() && ! function_exists( 'wc_order_proposal' ) ) ) {
 				continue;
-			};
+			}
 
 			foreach ( $document->output_formats as $output_format ) {
 				if ( $document->is_enabled( $output_format ) ) {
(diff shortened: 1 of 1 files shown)


includes/Main.php
```diff
@@ -285,10 +285,18 @@ public function get_documents_for_email( $email_id, $order ) {
...
-			if ( ! function_exists( 'WPO_WCPDF_Pro' ) && 'invoice' !== $document->get_type() ) {
+			/**
+			 * Filter documents for email attachment based on plugin availability.
+			 * 
+			 * - If Pro plugin is not active, only allow 'invoice' and 'proposal' document types
+			 * - 'proposal' document type is only available if the Order Proposal plugin is active
+			 * - Skip documents that don't meet these criteria
+			 */
+			if ( 
+				( ! function_exists( 'WPO_WCPDF_Pro' ) && ! in_array( $document->get_type(), array( 'invoice', 'proposal' ) ) ) ||
```
alexmigf: This doesn't seem correct to me. The Proposal is not part of the `WPO_WCPDF_Pro`.

alexmigf (changes requested):

includes/Main.php
BrunoPavlinic98: Now that I think about it, I think the current logic will work, but it is overly complicated without good reason. I added new code that is much simpler, and works as well.

What do you think?

includes/Main.php
```diff
@@ -285,10 +285,10 @@ public function get_documents_for_email( $email_id, $order ) {
 		$attach_documents = array();
 
 		foreach ( $documents as $document ) {
-			// Pro not activated, only attach Invoice
-			if ( ! function_exists( 'WPO_WCPDF_Pro' ) && 'invoice' !== $document->get_type() ) {
+			// This condition skips document types that require the Pro plugin unless it's an invoice or proposal
+			if ( ! in_array( $document->get_type(), array( 'invoice', 'proposal' ) ) && ! function_exists( 'WPO_WCPDF_Pro' ) ) {
```
alexmigf: Just add `true` as the third parameter for `in_array` for strict comparison.

alexmigf (changes requested):

includes/Main.php
BrunoPavlinic98: True, I tend to forget that. Done and thanks!

PR #1220, opened 2025-07-02
Update states if country changed by sync

### LINKED ISSUE #1219: State will not synchronize in settings though shop country supports states
- Shop country = USA
- USA supports states
- Colorado is the current shop state
Image
- At WooCommerce > PDF Invoices > General, I cannot sync the shop's state entered via WooCommerce. It shows the message "No states available"
Image
- When clicking the sync button, the field remains greyed-out.
Image

MohamadNateqi:
close #1219


alexmigf (approved):
LGTM!

dpeyou:
It seems I have to click twice in order to get the correct state (Colorado). A user might not know this. 

- The first click loads the 1st state available alphabetically:
loads first state available alphabetically instead of correct state

- The 2nd click loads the correct shop state:
click 2 times to get correct

To counteract this, a double-click warning may need to be added.

alexmigf:
@MohamadNateqi Maybe we could sync the 2 fields at the same time?

MohamadNateqi:
@dpeyou, I could not replicate this behavior. Could you please explain each step you take or share a recording?

Here's how it should work: syncing the country should first load new states, and then, with a single click of the sync button for the state field, it should automatically select the correct state.

alexmigf:
@MohamadNateqi I think that’s what @dpeyou meant—he clicked Country and then State, which takes two clicks and might be unnecessary.

MohamadNateqi:
@dpeyou Can you try the last commit please? It should automatically select the state when the country is synced.

alexmigf (commented):
Something isn’t working correctly with the States when multilingual is enabled.

image

In this case, Portugal is showing the states for Spain.

alexmigf:
Doesn't seem related to this PR tho.

PR #1221, opened 2025-07-03
Minor tweaks and improvements

MohamadNateqi:
This PR applies the following changes:
1. The style has been updated to ensure the sub-tab on the settings page always displays correctly. Previously, it was only working for the debug tab.
2. The `validate()` method now accepts null values. If it doesn't accept nulls, it could raise an error when no settings are saved, such as when only unchecked checkboxes are present on the page.
3. Small improvement.


woocommerce-pdf-invoices-packingslips.php
```diff
@@ -4,7 +4,7 @@
  * Requires Plugins:     woocommerce
  * Plugin URI:           https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/
  * Description:          Create, print & email PDF or UBL Invoices & PDF Packing Slips for WooCommerce orders.
- * Version:              4.6.0-rc.5
+ * Version:              4.6.1-beta-1
```
alexmigf: Maybe this is not needed.

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -22,7 +22,7 @@
 
 class WPO_WCPDF {
 
-	public $version              = '4.6.0-rc.5';
+	public $version              = '4.6.1-beta-1';
```
alexmigf: Same.

woocommerce-pdf-invoices-packingslips.php
MohamadNateqi: I made this change to update the base plugin version in the Pro PR. Why do you recommend reverting it?

alexmigf: `4.6.0` hasn’t been released yet, so we can leave it as is for now.

alexmigf:
The style of this sub-menu changed to the WP default:

image

But we can use the previous style in all sub-menus?

MohamadNateqi:
> Maybe we can use the previous style in all sub-menus?

I like the idea.

alexmigf:
Here looks a bit odd, maybe because the page doesn't have a main title:

image

MohamadNateqi:
@alexmigf What do you think now?

image

alexmigf:
I don't see much of a difference either, but I was referring to adding a title between the two menus. What do you think?

@n-rusu Could you share your input as well?

MohamadNateqi:
I updated the sub menu style to this:

image

Still, I'm not sure if this is the best way to implement it. Also, I added a bottom border because it seemed like the distinction wasn't clear.

@n-rusu, can you take a look and share your input, please?

n-rusu:
Looks good👌

PR #1225, opened 2025-07-04
Fix country sync button issue on multilingual setup

### LINKED ISSUE #1224: Incorrect states update on multilingual setup
In a multilingual setup, clicking the sync button for the country fields updates the country fields across all tabs. 
Additionally, the state field in the first language tab will always be updated, instead of the state field in the corresponding tab.

MohamadNateqi:
close #1224


alexmigf (approved):
LGTM!

PR #1227, opened 2025-07-07
Apply a filter on `get_shop_address()` return value

### LINKED ISSUE #1226: `wpo_wcpdf_shop_address_settings_text` filter not applying for shop address
Since this [PR](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1163), the `get_shop_address()` method no longer retrieves the address field value from `get_settings_text()`. As a result, the `wpo_wcpdf_shop_address_settings_text` filter applied to it within that method is no longer effective.

MohamadNateqi:
close #1226

Since the `get_shop_address()` is no longer directly associated with a setting, naming the filter `wpo_wcpdf_shop_address_settings_text` is no longer accurate. I have chosen a more appropriate name for it.

If we are concerned about compatibility, we can apply that filter as a deprecated filter.

### DIFF
--- a/includes/Documents/OrderDocument.php
+++ b/includes/Documents/OrderDocument.php
@@ -425,7 +425,7 @@ abstract class OrderDocument {
 		if ( empty( $number ) ) {
 			$number           = $order->get_meta( "_wcpdf_{$this->slug}_number" );
 			$formatted_number = $order->get_meta( "_wcpdf_formatted_{$this->slug}_number" );
-			
+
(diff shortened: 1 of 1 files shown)


alexmigf (commented):
We also have the `wpo_wcpdf_format_address` filter—so I'm not sure if we really need to introduce a new one.

alexmigf:
@MohamadNateqi Sure, please add also `$address` as a filter parameter.

alexmigf (approved):
LGTM!

PR #1228, opened 2025-07-14
Tweaks: Improves `shop_address` text hook deprecation

alexmigf:


includes/Documents/OrderDocument.php
```diff
@@ -1332,7 +1332,19 @@ public function get_settings_text( $settings_key, $default = false, $autop = tru
...
 		if ( in_array( $settings_key, array( 'shop_name', 'footer', 'extra_1', 'extra_2', 'extra_3' ) ) ) {
-			$text = apply_filters_deprecated( "wpo_wcpdf_{$settings_key}", array( $text, $this ), '4.5.0', "wpo_wcpdf_{$settings_key}_settings_text" );
+			$text = apply_filters_deprecated(
+				"wpo_wcpdf_{$settings_key}",
+				array( $text, $this ),
+				'4.5.0',
+				"wpo_wcpdf_{$settings_key}_settings_text"
+			);
+		} elseif ( 'shop_address' === $settings_key ) {
+			$text = apply_filters_deprecated(
```
MohamadNateqi, line 1342: The issue with this approach is that the filter will only affect the return value if the historical settings are enabled or exist; otherwise, the filter is ineffective.

alexmigf: I understand, but I’m not entirely comfortable with the deprecation being placed there. What’s your take on it?

MohamadNateqi: Honestly, I prefer removing it entirely because I don't believe it’s necessary. 
However, I agree to move it here as well.

PR #1231, opened 2025-07-17
Fix argument mismatch in `add_document_link_to_email`

### LINKED ISSUE #1230: Handle argument mismatch in `add_document_link_to_email`
Currently, the add_document_link_to_email() method is attached to multiple WooCommerce email-related hooks.
However, the `woocommerce_email_customer_address_section` have a different parameter signature:
```php
do_action( 'woocommerce_email_customer_address_section', $type, $order, $sent_to_admin, $plain_text );
```

This leads to incorrect argument mapping and runtime errors when the method is invoked via this hook.

<hr>

A customer reported this error:

```php
CRITICAL Uncaught TypeError: WPO\IPS\Main::add_document_link_to_email(): 
Argument #4 ($email) must be of type WC_Email, bool given, called in /var/www/vhosts/example.com/httpdocs/wp-includes/class-wp-hook.php on line 324 
and defined in /var/www/vhosts/example.com/httpdocs/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Main.php:1864 
```

Some third-party plugins (e.g., Woo Cart Abandonment Recovery) improperly trigger WooCommerce email hooks with invalid `$email` arguments (e.g., `true` instead of a `WC_Email` object). Although we don't cause this issue, we should defensively handle these cases with `instanceof` checks before accessing email properties to prevent crashes.

MohamadNateqi:
close #1230

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1847,7 +1847,7 @@ class Main {
 		$email_hooks = apply_filters( 'wpo_wcpdf_add_document_link_to_email_hooks', $email_hooks );
 
 		foreach ( $email_hooks as $email_hook ) {
-			if ( $email_hook === 'woocommerce_email_customer_address_section' ) {
+			if ( 'woocommerce_email_customer_address_section' === $email_hook ) {
 				add_action( $email_hook, array( $this, 'add_document_link_to_address_section' ), 10, 4 );
 			} else {
 				add_action( $email_hook, array( $this, 'add_document_link_to_email' ), 10, 4 );
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1847,8 +1847,28 @@ class Main {
 		$email_hooks = apply_filters( 'wpo_wcpdf_add_document_link_to_email_hooks', $email_hooks );
 
 		foreach ( $email_hooks as $email_hook ) {
-			add_action( $email_hook, array( $this, 'add_document_link_to_email' ), 10, 4 );
+			if ( $email_hook === 'woocommerce_email_customer_address_section' ) {
+				add_action( $email_hook, array( $this, 'add_document_link_to_address_section' ), 10, 4 );
+			} else {
+				add_action( $email_hook, array( $this, 'add_document_link_to_email' ), 10, 4 );
+			}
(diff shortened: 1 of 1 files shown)


includes/Main.php
```diff
@@ -1847,21 +1847,41 @@ function handle_document_link_in_emails(): void {
 		$email_hooks = apply_filters( 'wpo_wcpdf_add_document_link_to_email_hooks', $email_hooks );
 
 		foreach ( $email_hooks as $email_hook ) {
-			add_action( $email_hook, array( $this, 'add_document_link_to_email' ), 10, 4 );
+			if ( $email_hook === 'woocommerce_email_customer_address_section' ) {
```
alexmigf:
```php
if ( 'woocommerce_email_customer_address_section' === $email_hook ) {
```

includes/Main.php
```diff
@@ -1847,21 +1847,41 @@ function handle_document_link_in_emails(): void {
...
+	}
+
 	/**
 	 * Add document download link to the email.
 	 *
 	 * @param \WC_Abstract_Order $order
 	 * @param bool $sent_to_admin
 	 * @param bool $plain_text
-	 * @param \WC_Email $email
+	 * @param mixed|\WC_Email $email Some third-parties might send different values.
```
alexmigf, line 1880: Maybe `null|\WC_Email` is enough?

includes/Main.php
```diff
@@ -1847,21 +1847,41 @@ function handle_document_link_in_emails(): void {
...
 	 *
 	 * @return void
 	 */
-	public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sent_to_admin, bool $plain_text, \WC_Email $email ): void {
+	public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sent_to_admin, bool $plain_text, $email ): void {
```
alexmigf, line 1884: `?\WC_Email`

alexmigf (changes requested):

alexmigf (approved):
LGTM!

PR #1234, opened 2025-07-18
Add Postcode Checker promotion message

### LINKED ISSUE #1174: Add Postcode Checker promotion message
We can include a promotional message for the shop address in the General settings, directing users to the Postcode Checker address validation tool.

Needs to be done after this task: https://github.com/wpovernight/wc-postcode-checker/issues/191.

MohamadNateqi:
close #1174

### DIFF
--- a/includes/Settings/SettingsGeneral.php
+++ b/includes/Settings/SettingsGeneral.php
@@ -346,7 +346,18 @@ class SettingsGeneral {
 					'option_name'   => $option_name,
 					'id'            => 'shop_address_postcode',
 					'translatable'  => true,
-					'description'   => __( 'The postal code, if any, in which your business is located.', 'woocommerce-pdf-invoices-packing-slips' ),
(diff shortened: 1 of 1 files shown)


alexmigf:
I know I originally proposed this, but I'm not sure it's strong enough to convince someone to buy the Postcode plugin—especially considering the international APIs are paid and could become costly. Maybe we could rephrase it to something more compelling, like:

> This address is used in your store's documents. To validate customer addresses at checkout, consider using the Address Lookup tool in the **Postcode Checker for WooCommerce** plugin.

Let me know what you think!

MohamadNateqi:
@alexmigf I had the same concern when working on this task. 

Your suggestion seems to promote a feature that isn't directly relevant to what the customer is doing or seeking. I mean, they're entering their store address when they encounter a feature promotion related to checkout for customers. 
Does promoting such a feature in that context make sense?

PR #1240, opened 2025-07-23
Update class attributes in templates

### LINKED ISSUE #1239: Update class attribute to use hyphen in template files
In templates, we're using this class attribute:

```HTML
class="shop_email_address"
```

It needs to be updated to use a hyphen instead of an underscore.

MohamadNateqi:
close #1239


alexmigf (commented):
I'm confused about this changes.

alexmigf (approved):

PR #1242, opened 2025-07-25
Improvements in `get_setting()`

### LINKED ISSUE #1241: Get array values from `SettingsGeneral::get_setting()`
The `get_setting()` method in the `SettingsGeneral` class only returns string values and does not support array values. It should be updated to support arrays as well.

MohamadNateqi:
close #1241

This PR will update the `get_setting()` method, so that:
1. It can return array values.
2. It can return arrays values even if the array is not a multilingual array.


includes/Settings/SettingsGeneral.php
```diff
@@ -726,13 +731,30 @@ private function get_setting( string $key, string $locale = '' ): string {
...
-			wptexturize( trim( $setting_text ) ),
+			is_string( $setting_text ) ? wptexturize( trim( $setting_text ) ) : $setting_text,
 			$key,
 			$locale,
 			$general_settings
 		);
 	}
 
+	/**
+	 * Check if an array is a multilingual array (all keys are 2-letter locale codes or 'default').
```
alexmigf: I'm not sure this is the best approach, as TranslatePress uses different language codes.

image

Even setting that aside, it appears to open because you're treating any two-letter key as valid.

MohamadNateqi: I agree it's not the best approach, but I couldn't find a better one to address this issue. Do you have any suggestions?

We can also split `init_settings()` into two methods: one for retrieving the settings array and another for registering settings with `add_settings_fields()`. This approach allows us to obtain each setting field definition from the first method and determine whether it is multilingual. What do you think?

alexmigf: We can obtain the language codes if a multilingual plugin is installed from `wpo_wcpdf_get_multilingual_languages()`. If it's not the case, then we need to migrate the settings to the default.

MohamadNateqi: This function will return the language codes if any translation plugin is activated.
If a customer had the translation plugin enabled, updated their settings, and then deactivated the plugin, the setting array in the database still contains translations.
In this scenario, using that function will fail.

alexmigf: That's why I suggested to migrate them, because without the multilingual plugin they are useless right?

MohamadNateqi: So you mean if no multilingual plugin is active, we just remove any other elements in the array and only keep the default element?

For example, in this array:
```php
"shop_address_line_1" => [
    "en" => "Address line 1 data",
    "nl" => "Address line 1 data",
    "default" => "Address line default",
],
```

We only keep this:
```php
"shop_address_line_1" => [
    "default" => "Address line default",
],
```

alexmigf: Yes.

MohamadNateqi: The array does not contain a default element to do this migration. The array is like this:

```php
"shop_address_line_1" => [
    "en" => "Address line 1 data",
    "nl" => "Address line 1 data",
],
```

alexmigf: Maybe because it wasn't filled before you installed the multilingual plugin?

MohamadNateqi: No, it was filled before activating a multilingual plugin, but it was overwritten afterward.

alexmigf: Hum, maybe we shouldn't do that?

MohamadNateqi: Yes, that makes sense — we should always have a default value.
Additionally, we should generate the default value by copying one of the existing language values if it's not already set. This covers cases like when the multilingual plugin is activated first, and the PDF Invoice plugin is installed later, requiring the default value to be filled automatically.

alexmigf: I agree.

MohamadNateqi: I considered modifying the options before saving to the database using the `pre_update_option_{$option}` filter.

However, I ran into a limitation: some fields are arrays but not translation arrays—for example, static files field. With WPML, we rely on its media translation and don't wrap such fields in language tabs. This again forces us to determine whether the array is a translation array or not.

An alternative is to add a hidden input for the "default," but I think that's not a good approach and it would be better to find a better one.

Do you have any suggestions?

MohamadNateqi: @alexmigf For this matter, I think we can create a new method that holds all translatable fields. Then we can refer to this method to check if a setting value is a translation array or just a normal array in the `get_settings()` method. I might be the simplest way to solve this.

alexmigf: I was thinking that we could always have the `default` key, and when the array has the number of keys superior to 1 it's a multilingual. This way we don't need to check for the key.

MohamadNateqi: Current general settings don't have any array, so I think we can simply add the defualt value using `pre_update_option_{$option}` filter.

For the static files array, I can update the migration to place the value under the `default` key.

includes/Settings/SettingsGeneral.php
```diff
@@ -9,25 +9,28 @@
...
 class SettingsGeneral {
 
-	protected $option_name = 'wpo_wcpdf_settings_general';
+	protected string $option_name = 'wpo_wcpdf_settings_general';
 
-	protected static $_instance = null;
+	protected static ?self $_instance = null;
 
-	public static function instance() {
+	public static function instance(): ?SettingsGeneral {
```
alexmigf: Can be `?self`

includes/Settings/SettingsGeneral.php
```diff
@@ -739,20 +738,32 @@ public function get_setting( string $key, string $locale = '' ) {
...
-			if ( ! preg_match( '/^[a-z]{2}$/i', strtolower( $key ) ) && 'default' !== $key ) {
-				return false;
+	public function add_default_element_to_options( $new_value ) {
+		if ( empty( $new_value ) || ! is_array( $new_value ) ) {
+			return $new_value;
+		}
+
+		foreach ( $new_value as $field_key => $field_value ) {
+			// Not a multilingual value.
+			if ( ! is_array( $field_value ) ) {
```
alexmigf, line 754: Maybe this can be problematic for the static files from Pro?

MohamadNateqi: I think this can be handled using the migration. Currently, it's not under General settings.
I can update the migration and put its value under the `default` key. This way, this code should also work fine, right?

alexmigf: I'll test it tomorrow — I just looked at the code.

MohamadNateqi: Of course, that's not the only issue. We might have arrays, although they won't be translatable, such as `static_file_attach_to_email_ids`.

Putting this under `default` doesn't make sense, and the current code also causes issues.

alexmigf:
> Putting this under default doesn't make sense

Could you explain your reasoning?

includes/Settings/SettingsGeneral.php
alexmigf: > Putting this under default doesn't make sense

Could you explain your reasoning?

MohamadNateqi: Since the `static_file_attach_to_email_ids` is not a translatable setting.
Of course, we can put its value under the `default` key if we don't want to keep this format only for translatable settings.

alexmigf: I was thinking on having the `default` permanent, with or without multilingual, this way you always know that those are the defaults.

MohamadNateqi: What about when we have an array for a setting value that is not a multilingual setting?

For example, for `static_file_attach_to_email_ids`, the submission request data for this setting is an array, maybe with several keys like this:

```php
"static_file_attach_to_email_ids" => [
    "cancelled_order" => "1",
    "customer_failed_order" => "1",
]
```

How can we determine if this is a multilingual array that requires adding a new `default` key, or if we need to move all data under the `default` key?

alexmigf: Maybe you could check if the key is an actual setting key or just 'default'?

alexmigf: This is starting to get a bit messy, to be honest. Maybe we should drop it for now and focus solely on the Pro plugin. We could consider using the property directly for the static files instead of relying on `get_setting()`.

MohamadNateqi: > Maybe you could check if the key is an actual setting key or just 'default'?

I may not have fully grasped your point. When submitting the setting form, there isn't a 'default' key in the data, so we can't check for it directly. To determine if a key is a setting key, I believe we need to compare it against the available options. This feels like a workaround and maybe not the most ideal approach, but again, I might not have fully understood your point.


> This is starting to get a bit messy, to be honest. Maybe we should drop it for now and focus solely on the Pro plugin. We could consider using the property directly for the static files instead of relying on get_setting().

I completely agree, especially since the documentation has also been updated. I'll revert the change involving `get_setting()` in the Pro PR so we can discuss this PR further later.

alexmigf: > I may not have fully grasped your point. When submitting the setting form, there isn't a 'default' key in the data, so we can't check for it directly. To determine if a key is a setting key, I believe we need to compare it against the available options. This feels like a workaround and maybe not the most ideal approach, but again, I might not have fully understood your point.

The idea was to use an array of setting keys, or even compare directly with the settings array, but I think that might be too complex to implement right now.

> I completely agree, especially since the documentation has also been updated. I'll revert the change involving get_setting() in the Pro PR so we can discuss this PR further later.

Thank you!

alexmigf:
@MohamadNateqi Do we still need this?

alexmigf:
@MohamadNateqi Maybe it's too risky for now—let’s focus on the other, more important PRs.

PR #1249, opened 2025-07-30
Fix: Restore full item meta display in PDF

### LINKED ISSUE #1248: Some item meta data missing from invoice PDF due to use of `get_formatted_meta_data()`
A customer reported that certain item meta fields were missing from the invoice PDF, specifically for bundled or composite products. This issue is caused by WooCommerce's internal change to the `wc_display_item_meta()` function, which now uses `get_formatted_meta_data()` instead of the previously used `get_all_formatted_meta_data()`.

https://github.com/woocommerce/woocommerce/commit/7332b64eeb63159b5c8b39390dec7df0a55142f8

Forums topic: https://wordpress.org/support/topic/item-contents-on-invoice-missing/

alexmigf:
closes #1248

### DIFF
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -1729,3 +1729,51 @@ function wpo_wcpdf_format_document_number( ?int $plain_number, ?string $prefix, 
 	// Add prefix & suffix
 	return $formats['prefix'] . $plain_number . $formats['suffix'];
 }
+
+/**
+ * Outputs item meta data.
+ *
+ * This is a customized version of the WooCommerce function `wc_display_item_meta()`,
+ * which uses the deprecated `get_all_formatted_meta_data()` method instead of 
+ * `get_formatted_meta_data()`.
+ *
+ * @param WC_Order_Item $item Order item object.
+ * @param array         $args Optional. Display arguments.
+ * 
+ * @return string|void Meta data HTML output or void if echoed directly.
+ */
+
+function wpo_ips_display_item_meta( \WC_Order_Item $item, array $args = array() ) {
+	$strings = array();
+	$html    = '';
+	$args    = wp_parse_args(
+		$args,
+		array(
+			'before'       => '<ul class="wc-item-meta"><li>',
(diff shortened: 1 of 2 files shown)


wpo-ips-functions.php
```diff
@@ -1729,3 +1729,51 @@ function wpo_wcpdf_format_document_number( ?int $plain_number, ?string $prefix,
 	// Add prefix & suffix
 	return $formats['prefix'] . $plain_number . $formats['suffix'];
 }
+
+/**
+ * Outputs item meta data.
+ *
+ * This is a customized version of the WooCommerce function `wc_display_item_meta()`,
+ * which uses the deprecated `get_all_formatted_meta_data()` method instead of
```
MohamadNateqi: I think the `get_all_formatted_meta_data` is not deprecated?

MohamadNateqi (approved):
LGTM!

wpo-ips-functions.php
```diff
@@ -1729,3 +1729,51 @@ function wpo_wcpdf_format_document_number( ?int $plain_number, ?string $prefix,
...
+	foreach ( $item->get_all_formatted_meta_data() as $meta_id => $meta ) {
+		$value     = $args['autop'] ? wp_kses_post( $meta->display_value ) : wp_kses_post( make_clickable( trim( $meta->display_value ) ) );
+		$strings[] = $args['label_before'] . wp_kses_post( $meta->display_key ) . $args['label_after'] . $value;
+	}
+
+	if ( $strings ) {
+		$html = $args['before'] . implode( $args['separator'], $strings ) . $args['after'];
+	}
+
+	$html = apply_filters( 'woocommerce_display_item_meta', $html, $item, $args );
```
MohamadNateqi: Would it be good to also apply our custom filter here?

wpo-ips-functions.php
alexmigf: By deprecated I mean replaced, but I agree it can be confusing.

PR #1264, opened 2025-08-15
Delivery Note document

BrunoPavlinic98:
This PR is part of [this task](https://github.com/wpovernight/woocommerce-pdf-ips-pro/issues/612).

### DIFF
--- a/includes/Documents/OrderDocumentMethods.php
+++ b/includes/Documents/OrderDocumentMethods.php
@@ -115,7 +115,7 @@ abstract class OrderDocumentMethods extends OrderDocument {
 	 * Check whether the billing address should be shown
 	 */
 	public function show_billing_address() {
-		if ( ! in_array( $this->get_type(), array( 'packing-slip', 'delivery-note' ) ) ) {
+		if ( ! in_array( $this->get_type(), array( 'packing-slip', 'delivery-note' ), true ) ) {
 			return true;
 		} else {
 			return ! empty( $this->settings['display_billing_address'] ) && ( $this->ships_to_different_address() || 'always' === $this->settings['display_billing_address'] );


templates/Simple/style.css
```diff
@@ -327,3 +327,13 @@ table.totals tr.payment_method {
 .pagenum,.pagecount {
 	font-family: sans-serif;
 }
+
+.signatures table {
```
alexmigf: I’m not sure these CSS rules belong here, since the template file isn’t included in this plugin.

BrunoPavlinic98: I agree. I have added it into the PRO using the `wpo_wcpdf_template_styles` hook.

includes/Documents/OrderDocumentMethods.php
```diff
@@ -115,7 +115,7 @@ public function billing_address() {
 	 * Check whether the billing address should be shown
 	 */
 	public function show_billing_address() {
-		if ( 'packing-slip' !== $this->get_type() ) {
+		if ( ! in_array( $this->get_type(), array( 'packing-slip', 'delivery-note' ) ) ) {
```
alexmigf: Strict comparison.

alexmigf (changes requested):

alexmigf (changes requested):
I think you forgot here too:

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/daee36d68acd748c1c1a03b7b9267d4ebf56aa82/includes/Main.php#L289

https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/daee36d68acd748c1c1a03b7b9267d4ebf56aa82/views/advanced-status.php#L165

alexmigf:
Ah, of course, nice catch!

includes/Documents/OrderDocumentMethods.php
```diff
@@ -115,7 +115,7 @@ public function billing_address() {
 	 * Check whether the billing address should be shown
 	 */
 	public function show_billing_address() {
-		if ( 'packing-slip' !== $this->get_type() ) {
+		if ( ! in_array( $this->get_type(), array( 'packing-slip', 'delivery-note' ), true ) ) {
```
MohamadNateqi, line 118: @BrunoPavlinic98 @alexmigf 
I just came across this PR and was wondering if introducing a filter would be a better option, as it would give third parties the flexibility to modify this value rather than hard-coding the document name that will be introduced in a third-party plugin?

BrunoPavlinic98: @MohamadNateqi Both the proposed and current implementation make sense to me since third-party plugins that create their own documents can always override the `show_billing_address` function. I personally don't have a preference.

MohamadNateqi: I think, generally, we should create flexibility for third parties to modify values rather than hard-coding those values directly.

alexmigf:
I’ve reverted this PR. This should be handled in the Pro plugin, since the document doesn’t exist in this one.

PR #1265, opened 2025-08-25
Update WC order api hook

### LINKED ISSUE #1254: Deprecated hook `woocommerce_api_order_response` usage
The `woocommerce_api_order_response` is part of the Woo legacy API, which was deprecated and removed in version `9.0.0`: https://github.com/woocommerce/woocommerce/pull/40627.

Currently, the plugin relies on this hook to display the invoice number in the API response; however, it is no longer functional.

MohamadNateqi:
close #1254

### DIFF
--- a/includes/Frontend.php
+++ b/includes/Frontend.php
@@ -20,7 +20,8 @@ class Frontend {
 
 	public function __construct()	{
 		add_filter( 'woocommerce_my_account_my_orders_actions', array( $this, 'my_account_invoice_pdf_link' ), 999, 2 ); // needs to be triggered later because of Jetpack query string: https://github.com/Automattic/jetpack/blob/1a062c5388083c7f15b9a3e82e61fde838e83047/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-my-account.php#L235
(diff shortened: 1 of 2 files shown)


includes/Frontend.php
```diff
@@ -100,22 +101,57 @@ public function open_my_account_pdf_link_on_new_tab() {
...
+		$data['wpo_wcpdf_invoice_number'] = $this->get_invoice_number( $order );
+
+		return $data;
+	}
+
+	/**
+	 * Add invoice number to WC REST API.
+	 *
+	 * @param \WP_REST_Response $response
+	 * @param \WC_Order $order
```
alexmigf: `WC_Data` instead?

includes/Frontend.php
```diff
@@ -100,22 +101,57 @@ public function open_my_account_pdf_link_on_new_tab() {
...
+	 *
+	 * @return \WP_REST_Response
+	 */
+	public function add_invoice_number_to_wc_order_api( \WP_REST_Response $response, \WC_Order $order ): \WP_REST_Response {
```
alexmigf: I believe it’s always better to display all supported arguments, since the code may be extended in the future and you wouldn’t notice any missing ones unless you check the source directly.

alexmigf (approved):

PR #1276, opened 2025-08-28
Set the next invoice number when the number field is set to `0`

### LINKED ISSUE #1274: A setting to set the default value of invoice number
Following the [conversation here](https://wpovernight.slack.com/archives/C0121KYKGN6/p1755889488282119), we can introduce a new setting under the Advanced tab to allow customers to set the default value of the invoice number in the "PDF document data" meta box when manually setting invoice details. They should be able to either set `0` as the default value or the next invoice number.

Image

MohamadNateqi:
close #1274


MohamadNateqi:
Sure, no problem. I'll implement the first idea.

PR #1279, opened 2025-08-28
New setting to allow choosing default manual invoice number

### LINKED ISSUE #1274: A setting to set the default value of invoice number

MohamadNateqi:
close #1274

This PR introduces a new setting under the Advanced settings to allow choosing the default invoice number value for the "PDF document data" editing form.

image

### DIFF
--- a/includes/Admin.php
+++ b/includes/Admin.php
@@ -868,9 +868,8 @@ class Admin {
 			$default_number  = 0;
 
 			if (
-				'invoice' === $document->get_type() &&
-				! empty( \WPO_WCPDF()->settings->debug_settings['default_manual_invoice_number'] ) &&
-				'next_invoice_number' === \WPO_WCPDF()->settings->debug_settings['default_manual_invoice_number']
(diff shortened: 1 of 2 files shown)


includes/Admin.php
```diff
@@ -865,14 +865,23 @@ public function get_current_values_for_document_data( OrderDocument $document, a
 		// Default number data
 		if ( ! isset( $current['number'] ) ) {
 			$number_settings = $document->get_number_settings();
+			$default_number  = 0;
+
+			if (
+				'invoice' === $document->get_type() &&
```
alexmigf: Maybe this can be applied to other documents as well?

PR #1283, opened 2025-08-29
Hide "PDF document data" metabox when no document is enabled

### LINKED ISSUE #1282: Hide "PDF document data" metabox when no document is enabled
Currently, the "PDF document data" metabox will be displayed even if no documents are enabled.

MohamadNateqi:
close #1282


MohamadNateqi:
Missed that. Thank you for mentioning.

PR #1284, opened 2025-08-29
New: Add advanced setting to configure document lock cleanup frequency

### LINKED ISSUE #1132: Add a new setting to select the interval for clearing the log of semaphore locks
A customer asked if they could delete the log of semaphore lock, like this:
- `wpo_ips_semaphore_lock_get_invoice_document_pdf_attachment_for_order`

Although we have an action to clear those specific locks, that are used to prevent [race conditions](https://en.wikipedia.org/wiki/Race_condition), @alexmigf suggested adding a new setting to modify this interval through a new setting.

alexmigf:
closes #1132

### DIFF
--- a/includes/Semaphore.php
+++ b/includes/Semaphore.php
@@ -420,7 +420,7 @@ class Semaphore {
 	 */
 	public static function schedule_semaphore_cleanup(): void {
 		if ( ! self::is_cleanup_scheduled() ) {
-			$frequency     = apply_filters( self::get_cleanup_hook_name() . '_frequency', ( WPO_WCPDF()->settings->debug_settings['semaphore_cleanup_frequency'] ?? 15 ) * DAY_IN_SECONDS ); // default: every 15 days
+			$frequency     = apply_filters( self::get_cleanup_hook_name() . '_frequency', ( (int) ( WPO_WCPDF()->settings->debug_settings['cleanup_days'] ?? 7 ) ) * DAY_IN_SECONDS );
 			$frequency     = apply_filters_deprecated( self::get_cleanup_hook_name() . '_interval', array( $frequency ), '4.7.0', self::get_cleanup_hook_name() . '_frequency' );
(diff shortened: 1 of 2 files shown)


includes/Semaphore.php
```diff
@@ -420,7 +420,7 @@ public static function get_cleanup_action(): ?object {
 	 */
 	public static function schedule_semaphore_cleanup(): void {
 		if ( ! self::is_cleanup_scheduled() ) {
-			$interval      = apply_filters( self::get_cleanup_hook_name() . '_interval', 30 * DAY_IN_SECONDS ); // default: every 30 days
+			$interval      = apply_filters( self::get_cleanup_hook_name() . '_frequency', ( WPO_WCPDF()->settings->debug_settings['semaphore_cleanup_frequency'] ?? 15 ) * DAY_IN_SECONDS ); // default: every 15 days
```
MohamadNateqi: Maybe applying a deprecated filter since the name has been changed?

includes/Settings/SettingsDebug.php
```diff
@@ -1054,6 +1054,33 @@ public function init_settings() {
 					'description' => __( 'Our plugin uses a semaphore class that prevents race conditions in multiple places in the code. Enable this setting only if you are having issues with document numbers, yearly reset or documents being assigned to the wrong order.', 'woocommerce-pdf-invoices-packing-slips' ),
 				)
 			),
+			array(
+				'type'     => 'setting',
+				'id'       => 'semaphore_cleanup_frequency',
+				'title'    => __( 'Semaphore cleanup frequency', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'select',
```
MohamadNateqi: Why not use a number input field for it?

MohamadNateqi (changes requested):

includes/Settings/SettingsDebug.php
alexmigf: We don't have a callback for it, but I had another idea.

MohamadNateqi (approved):
Great!

PR #1285, opened 2025-09-01
New: Include deleted document numbers in Number tools

### LINKED ISSUE #1250: "Numbers" tool: Avoid the invoice data log removal when the order is deleted permanently
A user reported that some invoices number were missing. After reviewing the "Numbers" logs, I realized that some rows were empty:

Image

Normally, we track the order related with the invoice number until it is moved to the trash:

Image

But if the order is deleted permanently, the logs are clear:

Image

alexmigf:
closes #1250

### DIFF
--- a/includes/Tables/NumberStoreListTable.php
+++ b/includes/Tables/NumberStoreListTable.php
@@ -88,12 +88,10 @@ class NumberStoreListTable extends \WP_List_Table {
 
 					// not invoice document but using invoice number system
 					if ( 'invoice' !== $document_type && in_array( $document_type, $invoice_number_store_doc_types, true ) ) {
-						$document_type = 'invoice';
-
 						// using invoice number
 						if ( empty( $number_data ) ) {
-							$number_data   = $order->get_meta( "_wcpdf_invoice_number_data", true );
-							$saved_number  = (int) ( $number_data['number'] ?? 0 );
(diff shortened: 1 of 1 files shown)


includes/Tables/NumberStoreListTable.php
```diff
@@ -58,87 +58,95 @@ public function column_default( $item, $column_name ) {
...
-						$saved_number  = isset( $number_data['number'] ) ? $number_data['number'] : null;
-						$order_id      = $order->get_id() ?? 0;
-						$item_order_id = $item->order_id ?? 0;
-						$item_id       = $item->id;
+				$document_type = reset( $document_types );
+
+				if ( ! empty( $document_type ) && $order ) {
+					$document_slug = str_replace( '-', '_', $document_type );
+					$number_data   = $order->get_meta( "_wcpdf_{$document_slug}_number_data", true );
+					$saved_number  = (int) $number_data['number'] ?? 0;
```
MohamadNateqi: The cast executes before the ternary. Same for others.

YordanSoares (commented):
Now the numbers keep logged, even if the orders are deleted permanently! 🥳 

Just a remark: When the order is deleted, the document type switch from `'invoice'` to `'credit-note'` (see number `128` below):
image

This only happen with the invoices logs, other documents works correctly.

YordanSoares (approved):
In that case, LGTM!

PR #1317, opened 2025-10-16
Fix "Include document link in emails" feature bug

### LINKED ISSUE #1316: Bug in "Include document link in emails" feature
<p class="editor-paragraph-block">Currently, only the Invoice document includes this feature. However, suppose we enable this feature for other documents as well, and each document places the URL in a different location within the same email type. In that case, the link will be displayed in all designated positions for each document.</p><p class="editor-paragraph-block"></p><p class="editor-paragraph-block">For example, consider the "Order on-hold" email type. Suppose the Invoice document adds its link under the "Order details" section, and the Packing Slip document adds its link under the "Customer details" section. In that case, the download link will appear in both sections for each document, creating a duplication.</p>

MohamadNateqi:
closes #1316

The issue was that the link for each document was being added on any email hook (placement) collectively from all document settings, rather than checking the document's specific placement with the action.

Since only the Invoice document has this feature, we didn't see this bug so far, but a customer reported that since they introduced customer documents to the plugin.

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1907,7 +1907,7 @@ class Main {
 			$is_allowed =
 				in_array( $document->get_type(), $allowed_document_types, true ) &&
 				in_array( $email_id, $selected_emails, true ) &&
-				'woocommerce_email_' . $email_placement === current_filter();
+				( 'woocommerce_email_' . $email_placement ) === current_filter();
 
 			if ( ! apply_filters( 'wpo_wcpdf_add_document_link_to_email_is_allowed', $is_allowed, $order, $sent_to_admin, $plain_text, $email ) ) {
 				continue;


MohamadNateqi:
@alexmigf Sorry, I forgot that, you're right.

includes/Main.php
```diff
@@ -1896,9 +1896,18 @@ public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sen
...
+			if ( empty( $email_placement ) ) {
+				continue;
+			}
 
 			$email_id   = ( $email instanceof \WC_Email ) ? $email->id : '';
-			$is_allowed = in_array( $document->get_type(), $allowed_document_types, true ) && in_array( $email_id, $selected_emails, true );
+			$is_allowed =
+				in_array( $document->get_type(), $allowed_document_types, true ) &&
+				in_array( $email_id, $selected_emails, true ) &&
+				'woocommerce_email_' . $email_placement === current_filter();
```
alexmigf: Shouldn't be `('woocommerce_email_' . $email_placement) === current_filter()` instead?

MohamadNateqi: Both can work; however, I updated the code to improve it. Thanks for the suggestion.

alexmigf (approved):
LGTM!

PR #1318, opened 2025-10-16
Enhance Semaphore class: Implement transient caching for cleanup status

ignacyhologa:
# Performance improvement for scheduled cleanup in Semaphore

## Problem

The current implementation of `Semaphore::is_cleanup_scheduled()` runs a database query on every request to check if the Action Scheduler job is scheduled. This can cause unnecessary load, especially on high-traffic sites.

## Solution

Cache the result of the scheduled action check using a WordPress transient. This reduces database queries to at most once per hour (or whatever interval you set), while still ensuring the scheduled job is created if missing.

### Implementation

- Added a transient (`wpo_ips_semaphore_cleanup_scheduled`) to cache the result of the scheduled action check for 1 hour.
- On every check, if the transient exists, use its value instead of querying the DB.
- When a new recurring action is scheduled, update the transient immediately.

## Why this is better

- Reduces DB queries from every request to once per hour.
- Works with or without object cache (transients fallback to DB if no cache).
- No risk of missing scheduled jobs: if the job is missing, it will be scheduled and the cache updated.
- No change to external API or plugin usage.

Below images show the **3 sql queries** that are being skipped due to this caching:

Screenshot 2025-10-16 at 19 09 32
Screenshot 2025-10-16 at 19 09 18

### DIFF
--- a/includes/Semaphore.php
+++ b/includes/Semaphore.php
@@ -9,6 +9,11 @@ if ( ! defined( 'ABSPATH' ) ) {
 if ( ! class_exists( '\\WPO\\IPS\\Semaphore' ) ) :
 
 class Semaphore {
+	/**
+	 * Transient key for caching scheduled cleanup status
+	 * @var string
+	 */
+	public const CLEANUP_TRANSIENT_KEY = 'wpo_ips_semaphore_cleanup_scheduled';
 
 	/**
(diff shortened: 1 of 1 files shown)


alexmigf (approved):
Thank you for taking the time to contribute and open this PR, we really appreciate your help.

PR #1319, opened 2025-10-17
Improvement: Update cleanup scheduling logic to support new Action Scheduler functions

ignacyhologa:
Update action scheduler register actions as advised in the action scheduler documentation.


alexmigf (approved):
LGTM!

PR #1321, opened 2025-10-17
New: Add `missing_template_files_notice()` function for custom template files missing

### LINKED ISSUE #1320: Add missing_template_files_notice() function
<p class="editor-paragraph-block">We should move the <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">missing_template_files_notice()</code> function from the Pro plugin to the free plugin, as displaying missing template file notices is a core feature and makes more sense to be available by default.</p>

alexmigf:
closes #1320

### DIFF
--- a/includes/Settings/SettingsGeneral.php
+++ b/includes/Settings/SettingsGeneral.php
@@ -856,7 +856,7 @@ class SettingsGeneral {
 		}
 
 		$missing_list = wp_sprintf_l( '%l', array_map( 'esc_html', $missing_titles ) );
-		$file_label   = _n( 'file', 'files', count( $missing_titles ), 'woocommerce-pdf-invoices-packing-slips' );
+		$file_label   = _n( 'a file', 'files', count( $missing_titles ), 'woocommerce-pdf-invoices-packing-slips' );
 
 		$notice  = '<div class="notice notice-info inline notice-wpo"><p>';
 		$notice .= sprintf(


includes/Settings/SettingsGeneral.php
```diff
@@ -84,12 +84,12 @@ public function init_settings() {
 					'option_name'      => $option_name,
 					'id'               => 'template_path',
 					'options_callback' => array( $this, 'get_installed_templates_list' ),
-					/* translators: 1,2. template paths */
-					'description'      => apply_filters( 'wpo_wcpdf_template_settings_path_description', sprintf(
+					'description'      => sprintf(
+						/* translators: 1,2. template paths */
 						__( 'Want to use your own template? Copy all the files from %1$s to your (child) theme in %2$s to customize them' , 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi: Maybe making it `file(s)`.

See this:

Image

includes/Settings/SettingsGeneral.php
```diff
@@ -807,6 +813,95 @@ private function get_setting( string $key, string $locale = '' ): string {
...
+	 * @param string[] $missing_titles
+	 * @return string
+	 */
+	private function render_missing_template_files_notice( array $missing_titles ): string {
+		if ( empty( $missing_titles ) ) {
+			return '';
+		}
+
+		$missing_list = wp_sprintf_l( '%l', array_map( 'esc_html', $missing_titles ) );
+		$file_label   = _n( 'file', 'files', count( $missing_titles ), 'woocommerce-pdf-invoices-packing-slips' );
```
MohamadNateqi: For singular, it should be "a file"

includes/Settings/SettingsGeneral.php
```diff
@@ -807,6 +813,95 @@ private function get_setting( string $key, string $locale = '' ): string {
...
+			'</strong>'
+		);
+
+		// Premium Templates guidance (only if bundle license missing/invalid)
+		if ( function_exists( 'WPO_WCPDF_Templates' ) ) {
+			$license_info = WPO_WCPDF()->settings->upgrade->get_extension_license_infos();
+			$info         = $license_info['bundle'] ?? null;
+
+			if ( empty( $info['status'] ) || 'valid' !== $info['status'] ) {
+				$template_folder = str_replace( wp_normalize_path( ABSPATH ), '', wp_normalize_path( WPO_WCPDF_Templates()->plugin_path() ) . '/templates/...' );
```
MohamadNateqi, line 882: Maybe using `{choosen_theme}` instead of `...`, since it doesn't look nice and clear.

image

alexmigf: I don't think that works because the chosen template is a custom one.

MohamadNateqi: I feel like the sentence that starts with "Otherwise" can be improved and is a little confusing, don't you think?

Also, shouldn't it be "copy the template files from the newest version **to**(instead of in)"?

alexmigf: I think the intention is to tell the user to copy the latest version from the plugin’s templates.

PR #1323, opened 2025-10-20
Fix: Prevent fatal error when `SettingsCallbacks::validate()` receives `false`

### LINKED ISSUE #1322: Fatal error when SettingsCallbacks::validate() receives false
<p class="editor-paragraph-block">A fatal error occurs during plugin upgrades because <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">SettingsCallbacks::validate()</code> is sometimes called with <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">false</code> instead of an array, causing a type mismatch.</p><p class="editor-paragraph-block">Forum topic: <a target="_blank" rel="noopener noreferrer nofollow" class="text-custom-primary-300 underline underline-offset-[3px] hover:text-custom-primary-500 transition-colors cursor-pointer" href="https://wordpress.org/support/topic/pdf-invoices-not-working-latest-update/">https://wordpress.org/support/topic/pdf-invoices-not-working-latest-update/</a></p>

alexmigf:
closes #1322


includes/Settings/SettingsCallbacks.php
```diff
@@ -734,11 +734,14 @@ public function normalize_settings_args( $args ) {
...
-	 *
-	 * @return array|null validated options.
+	 * @param array|false|null $input Options to validate.
+	 * @return array|null Validated options.
 	 */
-	public function validate( ?array $input ): ?array {
+	public function validate( $input ): ?array {
+		if ( $input === false ) {
+			$input = null;
+		}
```
MohamadNateqi: I think this would be unnecessary because we check the `empty( $input )` in the next lines.

MohamadNateqi (changes requested):

includes/Settings/SettingsCallbacks.php
alexmigf: That way the function returns false.

MohamadNateqi: Ah, you're right.
So, maybe we can change the return value to `null` for that if-statement?

alexmigf: I pushed some changes.

MohamadNateqi (approved):
LGTM!

PR #1329, opened 2025-10-29
Improve the notice for unavailable documents in the previewer

### LINKED ISSUE #1326: Improve the notice for unavailable documents in the previewer
<p class="editor-paragraph-block">Currently, if a document is not yet available, for instance because the order has not reached an allowed status (configured within the "Disable for" setting) or based on any different condition, we always display the same message:</p><blockquote><p class="editor-paragraph-block">Document not available for order #1234, try selecting a different order.</p></blockquote><p class="editor-paragraph-block">But this causes confusion in some users who interpret the message as a type of error, perhaps because the message lacks additional context or an explanation of the cause. This message also uses the <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">.notice-error</code>, but I think it should be highlighted as a warning or an info notice instead.</p><h3 class="editor-heading-block">Improvement proposal</h3><p class="editor-paragraph-block">We could add more context to the message, so the users understand better why the document can't be created yet, for instance:</p><blockquote><p class="editor-paragraph-block">Invoice not yet available for order #1234, because it is disabled for the current order status. Try selecting a different order to preview your changes.</p></blockquote><p class="editor-paragraph-block">For the credit note case, we can also check whether an invoice or a refund has not been created yet (this is a common cause of confusion):</p><blockquote><p class="editor-paragraph-block">Credit Note not yet available for order #1234, because an invoice does not exist. Try selecting a different order to preview your changes.</p></blockquote><p class="editor-paragraph-block">Or</p><blockquote><p class="editor-paragraph-block">Credit Note not yet available for order #1234, because a refund has not been generated yet. Try selecting a different order to preview your changes.</p></blockquote><p class="editor-paragraph-block">Finally, modify the notice styles, to display it as a <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">notice-warning</code>or <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">notice-info</code></p>

BrunoPavlinic98:
closes #1326

### DIFF
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -361,7 +361,7 @@ class Settings {
 								/* translators: 1. order ID, 2. documentation page link, 3. documentation page link closing tag */
 								esc_html__( 'The PDF preview for order #%1$d is not available. This can happen if some settings prevent the document from being generated. Please review your configuration or check the %2$sdocumentation%3$s for more details.', 'woocommerce-pdf-invoices-packing-slips' ),
 								$order_id,
-								'<a href="https://todo.com" target="_blank">',
+								'<a href="https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/troubleshooting-pdf-preview-unavailability/" target="_blank">',
 								'</a>'
 							)
 						)


includes/Settings.php
```diff
@@ -386,6 +382,63 @@ public function ajax_preview() {
...
+			} else {
+				return sprintf(
+					/* translators: order ID */
+					esc_html__( 'Credit Note not yet available for order #%d, because a refund has not been generated yet. Try selecting a different order to preview your changes.', 'woocommerce-pdf-invoices-packing-slips' ),
+					$order_id
+				);
+			}
+		} else {
+			$document_title = __( 'Document', 'woocommerce-pdf-invoices-packing-slips' );
+			foreach ( WPO_WCPDF()->documents->documents as $document ) {
```
alexmigf: `$document_titles = WPO_WCPDF()->documents->get_document_titles();`

includes/Settings.php
```diff
@@ -386,6 +382,63 @@ public function ajax_preview() {
...
+	}
+
+	/**
+	 * Get order not found error message.
+	 *
+	 * @param string $document_type
+	 * 
+	 * @return string
+	 */
+	public function get_order_not_found_error( string $document_type ): string {
```
alexmigf: Maybe this could be `null` as well, or have a default empty value, in case we don't want to pass the document type.

includes/Settings.php
BrunoPavlinic98: @alexmigf `WPO_WCPDF()->documents->get_document_titles();` returns document titles only for enabled documents. That is the reason why I went with `WPO_WCPDF()->documents->documents`, because then, we can display the exact document title for a particular document that is disabled and therefore, not available.

BrunoPavlinic98:
> We chose a more general message since there are many situations where a document cannot be generated, such as this one:
> 
>     * Document is disabled
> 
>     * Message refers to the order status
> 
> 
> Image
> 
> Maybe we could keep a general message, but with a link to a documentation page?

@alexmigf I tweaked the message a bit, making it cover both scenarios. Personally, I think it is great and informative now. Of course, we could add a documentation link, I agree with it. To which documentation article should this point?

And of course, please let me know what you think about the tweaked message.

includes/Settings.php
alexmigf: You could add a parameter to the function to make it more flexible.

alexmigf:
@BrunoPavlinic98 I still believe this isn’t the most effective way to handle it, as mentioned earlier.

Screenshot from 2025-11-04 09-32-25

includes/Settings.php
```diff
@@ -358,9 +358,11 @@ public function ajax_preview() {
 					wp_send_json_error(
 						array(
 							'error' => sprintf(
-								/* translators: order ID */
-								esc_html__( 'Document not available for order #%s, try selecting a different order.', 'woocommerce-pdf-invoices-packing-slips' ),
-								$order_id
+								/* translators: 1. order ID, 2. documentation page link, 3. documentation page link closing tag */
+								esc_html__( 'Document not available for order #%1$d, try selecting a different order. More information about why this might be happening can be found in the %2$sdocumentation page%3$s.', 'woocommerce-pdf-invoices-packing-slips' ),
```
alexmigf: Maybe this is better?

>The PDF preview for order #%1$d is not available. This can happen if some settings prevent the document from being generated. Please review your configuration or check the %2$sdocumentation%3$s for more details.

BrunoPavlinic98: I agree, this is better.

dpeyou:
I noticed I made a mistake with the very last image. It should be of the "choose a template" setting.

choose a template

dpeyou:
## Draft 2
I went ahead and posted this draft 2, so that we could have the link ready.

### Link
https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/troubleshooting-pdf-preview-unavailability/

### Changelog
- New section: **Missing Credit Note requirements**
  - It contains a link to the Credit Note documentation.
- New title: **Troubleshooting PDF Preview unavailability** based on [Bruno's comment](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1329#issuecomment-3504611706).
- Updated the PDF Preview unavailability message text and image.

PR #1330, opened 2025-10-29
Fix: Improve display of Shipping cost and location details in PDF totals

### LINKED ISSUE #1171: Shipping not cost printed with Pickup Location
Following the [thread here](https://wordpress.org/support/topic/price-of-local-pickup/#post-18451186), when the "Pickup Locations" is set under Woocommerce > Shipping > Local Pickup, the cost for the pickup is not printed, only the pickup location address is printed. This doesn't happen in our Premium Templates, though.

alexmigf:
closes #1171

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -1051,8 +1051,13 @@ class Main {
 	}
 
 	public function html_currency_filters( $filters ) {
-		// only apply these fixes if the bundled dompdf version is used!
-		if ( wcpdf_pdf_maker_is_default() ) {
+		// Apply currency font when Extended currency symbol support is enabled
+		if ( isset( WPO_WCPDF()->settings->general_settings['currency_font'] ) ) {
+			$filters[] = array( 'woocommerce_currency_symbol', array( $this, 'use_currency_font' ), 10001, 2 );
+			// 'wpo_wcpdf_custom_styles' is actually an action, but WP handles them with the same functions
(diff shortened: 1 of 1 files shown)


includes/Documents/OrderDocumentMethods.php
```diff
@@ -959,6 +959,20 @@ public function get_woocommerce_totals() {
...
+			
+			// Local pickup specific
+			if (
+				'shipping' === $key &&
+				\wpo_ips_order_has_local_pickup_method( $this->order ) &&
+				apply_filters( 'wpo_ips_show_pickup_location_details', true, $this )
+			) {
+				$order_shipping          = $this->get_order_shipping();
+				$totals[ $key ]['value'] = sprintf(
+					'%s<p class="pickup-location-details">%s</p>',
```
MohamadNateqi, line 971: What do you think about wrapping all around a P tag, and having a SPAN for "pickup-location-details"?

MohamadNateqi (approved):
LGTM!

includes/Documents/OrderDocumentMethods.php
alexmigf: The first placeholder shows the default pricing, and I’d like the pickup details to appear below it. That would require adding a line break or changing the span to display as a block element, I don't see that need.

MohamadNateqi: Good then.

PR #1364, opened 2025-11-27
New: Add ink saving mode setting for the Simple template

### LINKED ISSUE #99: Create a new "Light" PDF template
@Spreeuw said:

> I also think it's time for an additional 'Light' template in PDF Invoices. It should have a few improvements over the old 'Simple' template such as:

1. Out of the box cross-compatibility with mpdf
2. Maybe using floats instead of columns for the addresses so that they can be moved to the other side more easily
3. No black header to save ink

alexmigf:
closes #99

image

image

image


templates/Simple/template-functions.php
```diff
@@ -4,3 +4,31 @@
  * Requires PDF Invoices & Packing Slips for WooCommerce 1.4.13 or higher
  */
 if ( ! defined( 'ABSPATH' ) ) exit; // Exit if accessed directly
+
+add_filter( 'wpo_ips_ink_saving_supported_templates', function( $templates ) {
```
MohamadNateqi, line 8: Is this necessary?

alexmigf: Can you explain your reasoning?

alexmigf: Although the default value is already passed through `apply_filters`, these are code snippets that should be included in every template. That is why I prefer to keep them here as well, as it makes it easier to simply copy and paste them into other templates.

MohamadNateqi: That was the reason I mentioned it, as we had it as the default value. Thanks for explaining it, all good then.

MohamadNateqi (approved):
LGTM!

YordanSoares (approved):
✅ LGTM!

dwalkerpriv (approved):
Works well on my end 👍

PR #1369, opened 2025-12-02
Fix: Fatal error caused by invalid option type when loading debug settings

### LINKED ISSUE #1368: Fatal error when loading plugin due to invalid option data type
<p class="editor-paragraph-block">A fatal <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">TypeError</code> occurs on some installations where an option is stored as a string instead of an array, causing the plugin to crash during initialization.</p>

alexmigf:
closes #1368

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -5,7 +5,7 @@ Tags: woocommerce, pdf, ubl, invoices, packing slips
 Requires at least: 4.4
 Tested up to: 6.9
 Requires PHP: 7.4
-Stable tag: 5.2.0-rc.1
+Stable tag: 5.2.0-1368.1
 License: GPLv2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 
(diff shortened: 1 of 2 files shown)


includes/Settings.php
```diff
@@ -1240,6 +1236,21 @@ public function apply_setting_categories( array $settings_fields, array $setting
...
 		return $modified_settings_fields;
 	}
+	
+	/**
+	 * Initializes settings by loading them from the database.
+	 *
+	 * @return void
+	 */
+	private function load_settings(): void {
+		$general_settings = get_option( 'wpo_wcpdf_settings_general', array() );
```
MohamadNateqi, line 1246: It might be good to get them only if the class properties are empty.

alexmigf: I think this can be used to reload those settings, right?

MohamadNateqi: Yes, good point. I was thinking of lazy-loading in case it's useful.

PR #1373, opened 2025-12-04
New: Debug tool to generate full plugin settings report for support

### LINKED ISSUE #1372: Add a new option to get the plugin's info to improve support assistance
<p class="editor-paragraph-block">Currently, we use screenshots or the WooCommerce system report to get the users' setup to try to find a solution more quickly. However, sometimes we need to get more details from our plugin, e.g. which options are selected for a specific document, and we often rely on screenshots, which is not always ideal.</p><p class="editor-paragraph-block">However, at the meeting of October 22, Michael proposed to include a new setting to get a report of all the plugin's settings, so that we can get a text file with all our plugin's details in a single interaction.</p>

alexmigf:
closes #1372

image


includes/Settings/SettingsDebug.php
```diff
@@ -1398,30 +1424,26 @@ public function display_server_requirement_notice(): void {
...
-			if ( ! empty( $plugin_data ) ) {
-				$plugins[ $premium_plugin ] = array(
-					'name'      => $plugin_data['Name'],
-					'version'   => $plugin_data['Version'],
-					'is_active' => is_plugin_active( $premium_plugin ),
-				);
-			}
+		foreach ( $plugins as $plugin_file => &$data ) {
+			$license_slug        = $premium_plugins[ $plugin_file ] ?? '';
+			$data['license_key'] = $license_slug && isset( $licenses[ $license_slug ] ) ? $licenses[ $license_slug ] : '';
```
MohamadNateqi, line 1444: Would it be better to mask the license key? Like showing only a few characters. I say this to ensure that if customers share the report with others or the public, then they don't risk exposing their license data.

views/advanced-tools.php
```diff
@@ -3,6 +3,26 @@
...
+				'<a href="mailto:support@wpovernight.com">support@wpovernight.com</a>'
+			); ?></p>
+
+			<p><?php printf(
+				/* translators: %s: plugin link */
+				wp_kses_post( __( 'If you prefer, you can also create a temporary login link for our support team using the free plugin %s.', 'woocommerce-pdf-invoices-packing-slips' ) ),
+				'<a href="https://wordpress.org/plugins/temporary-login-without-password/" target="_blank" rel="noopener noreferrer"><strong>Temporary Login Without Password</strong></a>'
+			); ?></p>
+
+			<a href="<?php echo esc_url( wp_nonce_url( admin_url( 'admin-ajax.php?action=wpo_ips_plugin_report' ), 'wpo_ips_plugin_report', 'nonce' ) ); ?>" class="button button-primary">
```
MohamadNateqi: I think it would be good to have the export button in the Status page as well.

views/advanced-tools.php
```diff
@@ -3,6 +3,26 @@
...
+			<p><?php printf(
+				/* translators: support email link */
+				esc_html__( 'Download a report with plugin, environment, and store information that you can attach to support requests made to %s.', 'woocommerce-pdf-invoices-packing-slips' ),
+				'<a href="mailto:support@wpovernight.com">support@wpovernight.com</a>'
+			); ?></p>
+
+			<p><?php printf(
+				/* translators: %s: plugin link */
+				wp_kses_post( __( 'If you prefer, you can also create a temporary login link for our support team using the free plugin %s.', 'woocommerce-pdf-invoices-packing-slips' ) ),
+				'<a href="https://wordpress.org/plugins/temporary-login-without-password/" target="_blank" rel="noopener noreferrer"><strong>Temporary Login Without Password</strong></a>'
```
MohamadNateqi: Is it OK to mention another plugin here? I mean it's like promoting them.

views/plugin-report.php
```diff
@@ -0,0 +1,258 @@
...
+					<td class="report-meta-value"><?php echo esc_html( $store_url ); ?></td>
+				</tr>
+				<tr>
+					<td class="report-meta-label"><?php echo esc_html( 'Report Date:' ); ?></td>
+					<td class="report-meta-value"><?php echo esc_html( $report_date ); ?></td>
+				</tr>
+				<tr>
+					<td class="report-meta-label"><?php echo esc_html( 'Generated By:' ); ?></td>
+					<td class="report-meta-value">
+						<?php echo esc_html( $report_user->display_name . ' (' . $report_user->user_email . ')' ); ?>
```
MohamadNateqi: This is also exposing the customer's email, which might not be good.

views/plugin-report.php
```diff
@@ -0,0 +1,258 @@
...
+		</div>
+		
+		<div class="section-title"><?php echo esc_html( 'Directory Permissions' ); ?></div>
+
+		<div class="system-info">
+			<table class="config-table">
+				<?php foreach ( $dir_permissions as $key => $value ) : ?>
+					<tr>
+						<td><?php echo esc_attr( $key ); ?></td>
+						<td><?php echo wp_kses_post( wpo_ips_format_report_setting_value( $value ) ); ?></td>
```
MohamadNateqi: This is also exposing the customer directory, which is critical. Maybe the path should be completely masked, or we should warn customers about this data whenever they try to download a report.

views/plugin-report.php
```diff
@@ -0,0 +1,258 @@
...
+		</div>
+		
+		<!-- Debug Settings Group -->
+		<div class="settings-group">
+			<div class="group-header"><?php echo esc_html( 'Debug Settings' ); ?></div>
+			<table class="settings-table">
+				<?php foreach ( $debug_settings as $key => $value ) : ?>
+					<tr>
+						<td class="setting-name"><?php echo esc_attr( $key ); ?></td>
+						<td class="setting-value"><?php echo wp_kses_post( wpo_ips_format_report_setting_value( $value ) ); ?></td>
```
MohamadNateqi, line 174: After reaching this point in my review, I can see more data included in the report that might be sensitive and should not be exposed/leaked, either by purpose or by mistake.

Maybe we can have an option to include this kind of data in the reported data, like a checkbox or another button that downloads reports including private data.
Or issue a warning and rely on customers to be careful.

What do you think?

includes/Settings/SettingsDebug.php
```diff
@@ -1951,6 +1973,190 @@ public function maybe_schedule_unstable_version_check(): void {
...
+		return apply_filters( 'wpo_ips_multilingual_plugins_data', $plugins_data );
+	}
+	
+	/**
+	 * Get recent log excerpts for key handles.
+	 *
+	 * @param int $limit Number of lines per log file.
+	 * @return array
+	 */
+	protected function get_recent_logs( int $limit = 200 ): array {
```
MohamadNateqi: Isn't 200 lines by default a lot?

views/plugin-report.php
```diff
@@ -0,0 +1,258 @@
...
+							</tr>
+						<?php endforeach; ?>
+					</table>
+				</div>
+			<?php endforeach; ?>
+		<?php endif; ?>
+
+	</div>
+	
+	<?php if ( ! empty( $logs_data ) ) : ?>
```
MohamadNateqi, line 227: Same concern about the log files, which might contain sensitive data.

includes/Settings/SettingsDebug.php
```diff
@@ -1951,6 +1973,190 @@ public function maybe_schedule_unstable_version_check(): void {
...
+			}
+		}
+		
+		// Logs
+		$logs_data = $this->get_recent_logs();
+		
+		// Extensions Settings
+		$extensions_settings = apply_filters( 'wpo_ips_get_extensions_settings_for_report', array() );
+		
+		ob_start();
```
MohamadNateqi, line 2037: Is it necessary to generate a PDF file for it? I think a simple text file will also do the job and reduce the size and trouble of report generation.

views/plugin-report.php
alexmigf: This can only be generated by admins and is also protected by a nonce. Could you clarify in what situation this could be exposed or leaked? Adding a checkbox seems equivalent to clicking the Download button. We can certainly include a warning if needed, but I don’t see the issue you’re describing. Otherwise, the same concern would apply to all functions on the status page.

includes/Settings/SettingsDebug.php
alexmigf: The issue with that approach is its limited flexibility in styling and structure, and considering we’re only talking about a ~100 KB PDF, the difference is negligible to me.

views/plugin-report.php
alexmigf: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1373#discussion_r2599538874

views/plugin-report.php
alexmigf: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1373#discussion_r2599538874

views/plugin-report.php
MohamadNateqi: > This can only be generated by admins and is also protected by a nonce. Could you clarify in what situation this could be exposed or leaked?

My concern isn’t about non-admins downloading the file. The concern is what happens after the file is downloaded and potentially shared. For example, we can’t expect users to post this report in the WordPress forum, because even without license information, it may still contain other sensitive details. The same applies if they want to share the file with developers from different plugins when seeking support.

What I’m suggesting is an option that lets users choose what to include in the report—either through a checkbox that toggles sensitive data on or off, or by offering two separate download buttons: one for general data only, and another for general data plus sensitive information.

includes/Settings/SettingsDebug.php
MohamadNateqi: Do we actually need any styling or structure for this? Since the report is sent to us, and I assumed we could review the data without additional formatting. But I agree the size difference is negligible.

views/plugin-report.php
alexmigf: I understand your concerns, but even if we add checkboxes, the issue you’re describing would still occur after clicking **Download**.

That said, I’ll think it over and see if there’s a way we can improve it.

includes/Settings/SettingsDebug.php
alexmigf: I get your point, but the PDF’s structure makes the data clearer to read, and the size difference is minimal.

views/plugin-report.php
MohamadNateqi: Maybe we can remove the sensitive data for now and, if we see they're needed, improve it in the future?

views/advanced-tools.php
alexmigf: I think it’s fine to keep it in one place for now. We can revisit it later if needed.

views/plugin-report.php
```diff
@@ -0,0 +1,260 @@
...
+		</div>
+		
+		<div class="section-title"><?php echo esc_html( 'Directory Permissions' ); ?></div>
+
+		<div class="system-info">
+			<table class="config-table">
+				<?php foreach ( $dir_permissions as $key => $value ) : ?>
+					<tr>
+						<td><?php echo esc_attr( $key ); ?></td>
+						<td><?php echo wp_kses_post( wpo_ips_format_report_setting_value( $value ) ); ?></td>
```
MohamadNateqi: I think the directory path could also be sensitive for many customers.

includes/Settings/SettingsDebug.php
```diff
@@ -1182,6 +1184,11 @@ public function get_server_config(): array {
 		$woo_version       = defined( 'WC_VERSION' ) ? WC_VERSION : null;
 		$woo_hpos_enabled  = WPO_WCPDF()->order_util->custom_orders_table_usage_is_enabled();
 
+		// WP + Woo
```
MohamadNateqi, line 1182: Repeated lines

includes/Settings/SettingsDebug.php
```diff
@@ -1975,6 +1978,194 @@ public function maybe_schedule_unstable_version_check(): void {
...
+		$all_documents = \WPO_WCPDF()->documents->get_documents( 'all' );
+		if ( ! empty( $all_documents ) ) {
+			foreach ( $all_documents as $document ) {
+				$document_type                        = $document->get_type();
+				$documents_settings[ $document_type ] = \WPO_WCPDF()->settings->get_document_settings( $document_type, 'pdf' );
+			}
+		}
+		
+		// Logs
+		$logs_data = $this->get_recent_logs();
```
MohamadNateqi: Would it be better to prevent calling this function when `$include_sensitive` is `true`? To prevent unnecessary file reading process.
Same about getting user data and the rest.

includes/Settings/SettingsDebug.php
```diff
@@ -1975,6 +1978,194 @@ public function maybe_schedule_unstable_version_check(): void {
...
+		
+		// Extensions Settings
+		$extensions_settings = apply_filters( 'wpo_ips_get_extensions_settings_for_report', array() );
```
MohamadNateqi: Maybe include `$include_sensitive` to arguments.

wpo-ips-functions.php
```diff
@@ -1874,3 +1874,147 @@ function wpo_ips_get_refund_ids( $order_or_ids ) {
...
+
+			return $html;
+		}
+
+		// Server config array (required/value/result[/fallback])
+		if ( isset( $value['required'] ) || isset( $value['value'] ) || isset( $value['result'] ) ) {
+			$html = '<div class="config-item">';
+
+			if ( ! empty( $value['required'] ) ) {
+				$html .= '<div class="config-required"><strong>Required:</strong> ' . $value['required'] . '</div>';
```
MohamadNateqi, line 1946: `esc_html()`?

wpo-ips-functions.php
```diff
@@ -1874,3 +1874,147 @@ function wpo_ips_get_refund_ids( $order_or_ids ) {
...
+				if ( $result ) {
+					$html .= '<span class="badge badge-enabled">OK</span>';
+				} else {
+					$html .= '<span class="badge badge-warning">Not OK</span>';
+				}
+				$html .= '</div>';
+			}
+
+			if ( ! empty( $value['fallback'] ) && empty( $value['result'] ) ) {
+				$html .= '<div class="config-fallback"><em>' . $value['fallback'] . '</em></div>';
```
MohamadNateqi, line 1966: `esc_html()`?

wpo-ips-functions.php
```diff
@@ -1874,3 +1874,147 @@ function wpo_ips_get_refund_ids( $order_or_ids ) {
...
+
+		// Server config array (required/value/result[/fallback])
+		if ( isset( $value['required'] ) || isset( $value['value'] ) || isset( $value['result'] ) ) {
+			$html = '<div class="config-item">';
+
+			if ( ! empty( $value['required'] ) ) {
+				$html .= '<div class="config-required"><strong>Required:</strong> ' . $value['required'] . '</div>';
+			}
+
+			if ( isset( $value['value'] ) && '' !== $value['value'] && null !== $value['value'] ) {
```
MohamadNateqi: Checking null can be removed (3rd condition).

wpo-ips-functions.php
```diff
@@ -1874,3 +1874,147 @@ function wpo_ips_get_refund_ids( $order_or_ids ) {
...
+		if ( in_array( $normalized, array( 'enabled', 'yes', 'true', 'on' ), true ) ) {
+			return '<span class="badge badge-enabled">Enabled</span>';
+		}
+
+		if ( in_array( $normalized, array( 'disabled', 'no', 'false', 'off' ), true ) ) {
+			return '<span class="badge badge-disabled">Disabled</span>';
+		}
+
+		if ( in_array( $normalized, array( 'restricted', 'limited', 'partial', 'deprecated', 'experimental', 'warning' ), true ) ) {
+			return '<span class="badge badge-warning">' . esc_html( ucfirst( $value ) ) . '</span>';
```
MohamadNateqi, line 1910: `$normalized`?

MohamadNateqi (changes requested):

includes/Settings/SettingsDebug.php
```diff
@@ -1975,6 +1978,194 @@ public function maybe_schedule_unstable_version_check(): void {
...
+		ob_start();
+		include \WPO_WCPDF()->plugin_path() . '/views/plugin-report.php';
+		$report_html = ob_get_clean();
+
+		$pdf_settings = array(
+			'paper_size'		=> 'A4',
+			'paper_orientation'	=> 'portrait',
+			'font_subsetting'	=> false,
+		);
+		$pdf_maker = \wcpdf_get_pdf_maker( $report_html, $pdf_settings, $this );
```
YordanSoares, line 2051: The PDF worked perfectly in the 2 testing environments I tried, but I wonder what happen if `DOMDocument` is not available (I tried to disable it in Hostinger and Local to check, but it seems it comes enabled in the core). I think this is an edge case, but last week we have received a case like this ([see](https://secure.helpscout.net/conversation/3167153414/78008)).

Therefore, I was thinking that we could offer two options, like "Download report (PDF)" and "Download report (HTML)", just in case. What do you think?

views/advanced-tools.php
```diff
@@ -3,6 +3,28 @@
 <div id="debug-tools">
 	<div class="wrapper">
 		<?php do_action( 'wpo_wcpdf_before_debug_tools', $this ); ?>
```
YordanSoares, line 5: I wonder if it won't be better to move this tool to the "Status" page instead.

wpo-ips-functions.php
alexmigf: That's only for echoing.

wpo-ips-functions.php
alexmigf: I think it's fine?

MohamadNateqi: I thought you wanted to use `$normalized` instead of the `$value`, as you already trimmed and lowercased it.

alexmigf: That is for the comparison.

YordanSoares (approved):
@alexmigf LGTM! Great job! 👏

MohamadNateqi (approved):
LGTM!

PR #1388, opened 2026-01-05
Fix: Backend saving of Peppol Endpoint ID for guest orders

### LINKED ISSUE #1387: Peppol Endpoint ID is not saved on the order edit page for guest customers
<p class="editor-paragraph-block">When an order is placed by a guest user, the customer Peppol Endpoint ID entered or edited on the order edit page is not persisted due to a bug in the current implementation.</p>

alexmigf:
closes #1387


alexmigf:
Customer approved the changes.

PR #1389, opened 2026-01-05
Display a warning if the document number is already in used

### LINKED ISSUE #1386: Add warning: Document number already in use
<p class="editor-paragraph-block"><span>When manually editing document numbers, it is possible to set the invoice number to an already used document number. Can this be avoided? A user is asking for a warning to be shown<span>&nbsp;</span>-<span>&nbsp;</span></span><a target="_blank" rel="noopener noreferrer" class="text-custom-primary-300 underline underline-offset-[3px] hover:text-custom-primary-500 transition-colors cursor-pointer c-link" href="https://secure.helpscout.net/conversation/3162295786/77853?viewId=221726">Helpscout</a></p><image-component src="0c765433-5862-4c8f-8d45-6c85d5e2794c" width="300px" height="338px" id="d1b50001-5272-4128-bf33-2abbd05af9a6" aspectratio="0.8892900120336944"></image-component><image-component src="7368772a-fccd-465b-9604-723e2da9d006" width="300px" height="155px" id="d96d84ac-f02d-47a5-a45a-4855a3e865ee" aspectratio="1.9427860696517414"></image-component><p class="editor-paragraph-block"></p>

MohamadNateqi:
close #1386

This PR will ensure a warning appears beneath the invoice number field if the document number is already in use. No additional settings are required for this feature.

image


assets/js/order-script.js
```diff
@@ -258,7 +258,7 @@ jQuery( function( $ ) {
 			url:    wpo_wcpdf_ajax.ajaxurl,
 			method: 'POST',
 			data: {
-				action:   'wpo_wcpdf_preview_formatted_number',
+				action:   'wpo_wcpdf_invoice_number_changed',
```
alexmigf, line 261: Was this required?

includes/Admin.php
```diff
@@ -1741,11 +1741,11 @@ public function debug_enabled_warning( $wp_admin_bar ) {
 	}
 
 	/**
-	 * AJAX handler to preview a formatted number
+	 * AJAX handler for invoice number change.
 	 *
 	 * @return void
 	 */
-	public function ajax_preview_formatted_number(): void {
+	public function ajax_invoice_number_changed(): void {
```
alexmigf, line 1748: Also this. I’m not sure why this needs to be changed.

wpo-ips-functions.php
```diff
@@ -109,6 +109,32 @@ function wcpdf_get_document( string $document_type, $order, bool $init = false )
...
+function wcpdf_get_document_by_number( int $number, string $document_type ) {
+	if ( empty ( $document_type ) ) {
+		return null;
+	}
+
+	global $wpdb;
+
+	$query = wpo_wcpdf_prepare_identifier_query(
+		"SELECT * FROM %i WHERE id = %d LIMIT 1",
+		array( $wpdb->prefix . 'wcpdf_' . $document_type . '_number' ),
```
alexmigf, line 129: I don’t think this is a reliable method. The table can be empty. Using order meta instead would be preferable, for example `_wcpdf_invoice_number_data`.

alexmigf (changes requested):

assets/js/order-script.js
MohamadNateqi: I changed the name of this action because it now serves two purposes: displaying the formatted number and displaying a warning if the number is already in use. So I thought a general name would be better.
What do you think?

includes/Admin.php
MohamadNateqi: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1389#discussion_r2664775054

wpo-ips-functions.php
MohamadNateqi: You mean an order might have an invoice while it's not stored in the DB?
Isn't the DB our source of truth?

alexmigf: Not necessarily. The invoice is stored as order meta. In scenarios like site migrations or resets via the number tools, the table may be empty while the order still has an invoice.

assets/js/order-script.js
alexmigf: Isn’t it still just a preview of the formatted number?

MohamadNateqi: Well, previewing the formatted number and also the warning if it should be displayed.

wpo-ips-functions.php
MohamadNateqi: So the meta is the source of truth!
In this case, I think implementing this isn't ideal, as it could require processing hundreds or thousands of orders to check their metadata and identify duplicates, which could impact performance.

assets/js/order-script.js
alexmigf: I see it as part of the preview.

wpo-ips-functions.php
alexmigf: You could use the data from the number tools, but ideally this should be triggered by a button rather than being applied automatically. That said, I agree it’s probably not worth the extra effort for something that isn’t recommended in the first place.

MohamadNateqi: Then, if you agree, we close this PR and the issue, and re-evaluate this in the future if we receive more customer requests.

alexmigf: I agree.

PR #1394, opened 2026-01-06
New: Added configurable custom field to checkout

### LINKED ISSUE #1093: Add configurable checkout field
In a team meeting, we discussed adding a new configurable field at checkout that can be used for collecting a local NIF (e.g., in Spain) but can also serve other purposes. Merchants could use it to collect tax identification numbers from other countries (e.g., VAT ID, CPF, RFC), company registration numbers for B2B transactions, purchase order numbers, or internal customer codes like membership or loyalty IDs. This field should be optional, enabled in the plugin settings, and allow merchants to set a custom label.

Open to discussion on placement, handling in order details, validation rules, and any refinements needed.

alexmigf:
closes #1093

image

image

image

### DIFF
--- a/includes/Frontend.php
+++ b/includes/Frontend.php
@@ -609,10 +609,6 @@ class Frontend {
 	 * @return void
 	 */
 	public function checkout_field_display_admin_billing( \WC_Order $order ): void {
-		if ( ! $order instanceof \WC_Order ) {
-			return;
-		}
-
 		// If your setting disables the field, don't show it.
 		if ( ! $this->checkout_field_is_enabled() ) {
 			return;
@@ -762,7 +758,7 @@ class Frontend {
 		}
 
 		$general_settings = WPO_WCPDF()->settings->general;
-		return ! empty( $general_settings->get_setting( 'checkout_field_enable_my_account' ) ?? '' );
+		return ! empty( $general_settings->get_setting( 'checkout_field_enable_my_account' ) );
 	}
 
 	/**
@@ -773,14 +769,7 @@ class Frontend {
 	private function checkout_field_get_label(): string {
 		$default          = __( 'Customer identification', 'woocommerce-pdf-invoices-packing-slips' );
 		$general_settings = WPO_WCPDF()->settings->general;
-		$label            = $general_settings->get_setting( 'checkout_field_label' ) ?? '';
-
-		// If the value is stored as an array, use the default key.
-		if ( is_array( $label ) ) {
-			$label = $label['default'] ?? '';
-		}
-
-		$label = trim( (string) $label );
+		$label            = trim( $general_settings->get_setting( 'checkout_field_label' ) );
 
 		if ( '' === $label ) {
 			$label = $default;
@@ -796,7 +785,7 @@ class Frontend {
 	 */
 	private function checkout_field_is_vat_number(): bool {
 		$general_settings = WPO_WCPDF()->settings->general;
-		$enabled          = ! empty( $general_settings->get_setting( 'checkout_field_as_vat_number' ) ?? '' );
+		$enabled          = ! empty( $general_settings->get_setting( 'checkout_field_as_vat_number' ) );
 
 		if ( ! $enabled ) {
 			return false;


wpo-ips-functions.php
```diff
@@ -2020,3 +2020,48 @@ function wpo_ips_get_plugins_data( array $plugin_files ): array {
...
+			},
+		)
+	);
+
+	if ( ! is_array( $detectors ) ) {
+		return false;
+	}
+
+	foreach ( $detectors as $detector ) {
+		if ( $detector() ) {
```
MohamadNateqi: Maybe checking if it is callable?

edi/Peppol.php
```diff
@@ -0,0 +1,1073 @@
...
+		}
+	}
+	
+
+	/**
+	 * Callback function to return false, used to disable storing document settings.
+	 *
+	 * @return bool
+	 */
+	public function return_false(): bool {
```
MohamadNateqi: Maybe you can use `__return_false` instead?

edi/Peppol.php
```diff
@@ -0,0 +1,1073 @@
...
+		if ( ! wpo_ips_edi_peppol_enabled_for_location( 'checkout' ) ) {
+			return;
+		}
+
+		$visibility_mode = $this->peppol_checkout_visibility_mode();
+
+		$has_company = ! empty( $data['billing_company'] );
+		$has_id      = ! empty( $data['peppol_endpoint_id'] );
+
+		$wants_peppol = false;
```
MohamadNateqi: Not needed, the value is overwritten immediately.

includes/Settings/SettingsGeneral.php
```diff
@@ -461,6 +469,49 @@ public function init_settings() {
 					'translatable' => true,
 				)
 			),
+			array(
```
MohamadNateqi, line 472: Is it better to put this settings under the Advanced settings?

edi/Peppol.php
```diff
@@ -0,0 +1,1073 @@
+<?php
+namespace WPO\IPS\EDI;
```
MohamadNateqi, line 2: Should it be "WPO\IPS\Peppol"?

edi/Peppol.php
```diff
@@ -0,0 +1,1073 @@
...
+						echo wp_kses_post( $description );
+					?>
+				</small>
+			</p>
+
+			<?php if ( 'select' === $input_mode ) : ?>
+				<p class="woocommerce-form-row woocommerce-form-row--wide form-row form-row-wide">
+					<label for="peppol_endpoint_eas"><?php esc_html_e( 'Peppol Endpoint Scheme (EAS)', 'woocommerce-pdf-invoices-packing-slips' ); ?></label>
+					<select name="peppol_endpoint_eas" id="peppol_endpoint_eas" class="woocommerce-Input input-select">
+						<option value=""><?php esc_html_e( 'Select', 'woocommerce-pdf-invoices-packing-slips' ) . '...'; ?></option>
```
MohamadNateqi: The `esc_html_e` echoes the string directly, and the `...` will not be displayed.

edi/Peppol.php
```diff
@@ -0,0 +1,1073 @@
...
+			return (string) $value;
+		}
+
+		$key      = str_replace( 'woocommerce_get_default_value_for_', '', current_filter() );
+		$meta_key = str_replace( '-', '_', substr( $key, strlen( 'wpo-ips-edi/' ) ) );
+
+		// Auto-enable toggle if we already have a stored identifier.
+		if ( 'peppol_invoice' === $meta_key ) {
+			$id  = (string) get_user_meta( $user_id, 'peppol_endpoint_id', true );
+			$eas = (string) get_user_meta( $user_id, 'peppol_endpoint_eas', true );
```
MohamadNateqi: Variable not being used.

includes/Frontend.php
```diff
@@ -318,465 +318,179 @@ public function restore_storing_document_settings(): void {
...
+				return (string) apply_filters( 'wpo_ips_checkout_field_sanitize', $val );
 			},
 			'validate_callback' => function ( $val ) {
-				$result = $this->peppol_validate_identifier_value( (string) $val );
-				return is_wp_error( $result ) ? $result : true;
+				$val = (string) $val;
+
+				// If not treated as VAT, keep the existing flexible hook.
+				if ( ! $this->checkout_field_is_vat_number() ) {
+					$result = apply_filters( 'wpo_ips_checkout_field_validate', true, $val );
```
MohamadNateqi, line 346: Would it be better to apply this filter for both situations?

includes/Frontend.php
```diff
@@ -785,546 +499,239 @@ public function edi_peppol_set_classic_checkout_fields_value( $value, string $in
...
-			}
+	/**
+	 * Get the checkout field label from settings.
+	 *
+	 * @return string
+	 */
+	private function checkout_field_get_label(): string {
+		$default  = __( 'Customer identification', 'woocommerce-pdf-invoices-packing-slips' );
+		$settings = WPO_WCPDF()->settings->general_settings;
+		$label    = $settings['checkout_field_label'] ?? '';
```
MohamadNateqi: Maybe using `get_setting()`? Same for other use cases.

includes/Frontend.php
```diff
@@ -41,25 +39,27 @@ public function __construct() {
...
-			add_filter( 'woocommerce_checkout_fields', array( $this, 'edi_peppol_display_classic_checkout_fields' ), 10, 1 );
-			add_filter( 'woocommerce_checkout_get_value', array( $this, 'edi_peppol_set_classic_checkout_fields_value' ), 10, 2 );
-			add_action( 'woocommerce_after_checkout_validation', array( $this, 'edi_peppol_validate_classic_checkout_field_values' ), 10, 2 );
-			add_action( 'woocommerce_checkout_update_order_meta', array( $this, 'edi_peppol_save_classic_checkout_fields' ), 10, 2 );
-			add_action( 'wp_enqueue_scripts', array( $this, 'edi_peppol_enqueue_classic_checkout_script' ), 20 );
+		
+		// Optional Checkout field (General Settings).
+		if ( $this->checkout_field_is_enabled() ) {
+			if ( wpo_wcpdf_checkout_is_block() ) {
+				$this->checkout_field_display_checkout_block_field();
```
MohamadNateqi, line 46: Should this field be also displayed in the My Account page?

includes/Frontend.php
```diff
@@ -785,546 +499,239 @@ public function edi_peppol_set_classic_checkout_fields_value( $value, string $in
...
-			if ( 'peppol_empty_endpoint' === $result->get_error_code() ) {
-				return new \WP_Error(
-					'peppol_empty_endpoint',
-					__( 'Peppol Endpoint ID is empty.', 'woocommerce-pdf-invoices-packing-slips' )
-				);
-			}
-
-			// Network/response errors: do not block checkout.
-			return true;
+	public function checkout_field_display_admin_billing( $order ): void {
```
MohamadNateqi: Type hint

MohamadNateqi (changes requested):

includes/Settings/SettingsGeneral.php
alexmigf: No, I don’t think it belongs there.

This is a general setting intended for all users, not just advanced ones.

edi/Peppol.php
alexmigf: No, it's inside the `edi` directory now.

edi/Peppol.php
alexmigf: Good catch!

edi/Peppol.php
MohamadNateqi: Then the `class_exists( '\\WPO\\IPS\\Peppol' )` should be updated accordingly.

alexmigf: Ahh, good point.

includes/Frontend.php
alexmigf: I hadn’t thought about that, but it makes sense.

includes/Frontend.php
```diff
@@ -785,546 +497,366 @@ public function edi_peppol_set_classic_checkout_fields_value( $value, string $in
...
+	 * @return void
 	 */
-	private function peppol_validate_identifier_value( string $raw_value ) {
-		$val = preg_replace( '/\s+/', '', trim( (string) $raw_value ) );
-
-		// Let "required" or other validation handle this elsewhere.
-		if ( '' === $val ) {
-			return true;
+	public function checkout_field_display_admin_billing( \WC_Order $order ): void {
+		if ( ! $order instanceof \WC_Order ) {
```
MohamadNateqi: Can be removed.

includes/Frontend.php
```diff
@@ -785,546 +497,366 @@ public function edi_peppol_set_classic_checkout_fields_value( $value, string $in
...
-	 * @param string $query
-	 * @return array|\WP_Error
+	 * @return string
 	 */
-	private function peppol_directory_request( string $query ) {
-		$base_url = 'https://directory.peppol.eu/search/1.0/json';
+	private function checkout_field_get_label(): string {
+		$default          = __( 'Customer identification', 'woocommerce-pdf-invoices-packing-slips' );
+		$general_settings = WPO_WCPDF()->settings->general;
+		$label            = $general_settings->get_setting( 'checkout_field_label' ) ?? '';
```
MohamadNateqi: Maybe `?? ''` can be removed.

includes/Frontend.php
```diff
@@ -785,546 +497,366 @@ public function edi_peppol_set_classic_checkout_fields_value( $value, string $in
...
+		if ( ! $user_id ) {
+			return;
 		}
 
-		/**
-		 * Scheme + value provided.
-		 */
+		$key   = 'wpo_ips_checkout_field';
+		$value = (string) get_user_meta( $user_id, $key, true );
+		$value = (string) apply_filters( 'wpo_ips_checkout_field_default_value', $value, $value, 'my-account', null );
```
MohamadNateqi, line 646: Sending 2 `$values` is needed?

alexmigf: I'm reusing an existing hook. What do you suggest?

MohamadNateqi: Sorry for the vague comment. I meant the `$value` has been repeated in the `apply_filters`. I meant, isn't this enough:

`apply_filters( 'wpo_ips_checkout_field_default_value', $value, 'my-account', null );`

alexmigf: I understand, but I think there may be a misunderstanding. This filter is already used in other places within the same class, and I’m simply following the existing convention.

MohamadNateqi: I see, understood.

MohamadNateqi (approved):

PR #1407, opened 2026-01-15
New: Add global template color setting

### LINKED ISSUE #1401: Modern template color setting displayed in wrong location
<p class="editor-paragraph-block">The Modern template color setting appears in the wrong section of the settings. A global template color setting may be needed to properly hide/show options via the settings API, similar to how the Ink Save setting is handled.</p>

alexmigf:
closes #1401 

image


templates/Simple/template-functions.php
```diff
@@ -8,27 +8,85 @@
...
 	}
 
-	$css .= "
-		.order-details thead th {
-			color: black;
-			background-color: white;
-			border-width: 0 0 0.8pt 0;
-			border-style: solid;
-			border-color: black;
+	$ink_saving_enabled = ! empty( $features['ink_saving']['enabled'] ) && ! empty( $features['ink_saving']['value'] );
```
MohamadNateqi: What is the "value" part for?

templates/Simple/template-functions.php
```diff
@@ -8,27 +8,85 @@
...
+			$text = $color;
+		} else {
+			$text = 'black';
 		}
-		.notes-totals .totals tfoot tr.order_total th,
-		.notes-totals .totals tfoot tr.order_total td {
-			border-top: .8pt solid black;
-			border-bottom: .8pt solid black;
+
+		$css .= "
```
MohamadNateqi, line 43: It might be good to introduce these values as a custom properties in the `:root`.

includes/Main.php
```diff
@@ -2028,38 +2028,92 @@ public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sen
...
+					$document,
+					$current_template,
+					$feature_key,
+					$feature_value,
+					$settings
+				);
+
+				if ( empty( $supported_templates ) || ! in_array( $current_template, $supported_templates, true ) ) {
+					$feature['enabled'] = false;
+					continue;
```
MohamadNateqi: Can be removed.

includes/Main.php
```diff
@@ -2028,38 +2028,92 @@ public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sen
...
 
-		$ink_saving_enabled = ! empty( $settings['template_ink_saving'] );
-		$current_template   = $settings['template_path'] ?? '';
+		$features = apply_filters(
+			'wpo_ips_template_style_features',
+			array(
+				'ink_saving' => array(
+					'enabled'                    => ! empty( $settings['template_ink_saving'] ),
+					'value'                      => ! empty( $settings['template_ink_saving'] ),
+					'supported_templates_filter' => 'wpo_ips_ink_saving_supported_templates',
```
MohamadNateqi: Would it be necessary to have a separate filter for each feature? Maybe the `wpo_ips_template_style_features` filter which already applied to the whole array is enough to modify `supported_templates`?

MohamadNateqi (changes requested):

templates/Simple/template-functions.php
alexmigf: Dompdf does not support CSS custom properties.

MohamadNateqi: I tested it on my end and could see the custom properties. Maybe it has been supported recently? I just saw this: https://github.com/dompdf/dompdf/pull/3314

alexmigf: It doesn’t seem to work reliably with the current totals implementation, and mPDF also appears to have inconsistent support for it.

includes/Main.php
```diff
@@ -2028,38 +2028,66 @@ public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sen
...
+		}
 
-		$supported_templates = apply_filters(
-			'wpo_ips_ink_saving_supported_templates',
-			array( 'default/Simple' )
+		$features = apply_filters(
+			'wpo_ips_template_style_features',
+			array(
+				'ink_saving' => array(
+					'setting_id' => 'template_ink_saving',
```
MohamadNateqi: Maybe the array key can be used as the setting ID?

MohamadNateqi (approved):
LGTM!

MohamadNateqi:
I think the "modern_color" for the modern template in the Premium template should be updated as well.

templates/Simple/template-functions.php
```diff
@@ -5,30 +5,85 @@
...
 	}
 
-	$css .= "
-		.order-details thead th {
-			color: black;
-			background-color: white;
-			border-width: 0 0 0.8pt 0;
-			border-style: solid;
-			border-color: black;
+	$ink_saving_enabled = ! empty( $features['template_ink_saving']['enabled'] );
```
MohamadNateqi, line 18: Just asking: do you think it would be good to have a global function to get the template color palette, and so move the logic here to that function? 
A function to return different colors for the template, like bg, border, etc. Or is it an unnecessary complication?

alexmigf:
Implemented here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1543

PR #1444, opened 2026-02-27
Applying a filter to due date in payment terms

### LINKED ISSUE #1439: Apply a filter to the due date for XML
<p class="editor-paragraph-block">We need to apply a new filter to the XML output's due date to enable validation in the Professional plugin, as we do for the PDF version.</p>

MohamadNateqi:
close #1439

### DIFF
--- a/edi/Syntaxes/Ubl/Handlers/PaymentTermsHandler.php
+++ b/edi/Syntaxes/Ubl/Handlers/PaymentTermsHandler.php
@@ -18,13 +18,7 @@ class PaymentTermsHandler extends AbstractUblHandler {
 	 */
 	public function handle( array $data, array $options = array() ): array {
 		$payment_terms = array();
-		$due_date_days = apply_filters(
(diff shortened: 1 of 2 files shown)


edi/Syntaxes/Ubl/Handlers/PaymentTermsHandler.php
```diff
@@ -18,7 +18,13 @@ class PaymentTermsHandler extends AbstractUblHandler {
 	 */
 	public function handle( array $data, array $options = array() ): array {
 		$payment_terms = array();
-		$due_date_days = $this->get_due_date_days();
+		$due_date_days = apply_filters(
```
alexmigf: Wouldn't be better to place the filter here? https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/47f5e03e2e52cc492aa49f76abf50bca02ff0c4e/edi/Abstracts/AbstractHandler.php#L85

PR #1468, opened 2026-03-13
New: Add search settings feature

### LINKED ISSUE #1461: Add a search field to settings
<p class="editor-paragraph-block">We can add a search field top of the settings pages: General settings, Documents settings, and Advanced settings.</p><p class="editor-paragraph-block">We have many settings and also many groups, sometimes it is hard to find one there.</p>

MohamadNateqi:
close #1461

image


n-rusu (approved):
@MohamadNateqi looks good👌

assets/js/admin-script.js
```diff
@@ -825,7 +825,199 @@ jQuery( function( $ ) {
 	// Initialize accordion
 	settingsAccordion();
 	//----------> /Settings Accordion <----------//
-	
+
+	//----------> Settings Search <----------//
+
+	function settingsSearch() {
+		const $input = $( '#settings-search' );
```
alexmigf: Maybe we could use `wpo` to make it more unique, just in case.

assets/js/admin-script.js
```diff
@@ -825,7 +825,199 @@ jQuery( function( $ ) {
...
+		}
+
+		// Create dropdown container.
+		const $dropdown = $( '<ul class="settings-search-dropdown"></ul>' );
+		$searchContainer.append( $dropdown );
+
+		let activeIndex = -1;
+		let matches     = [];
+
+
```
alexmigf, line 862: Maybe we don't need 2 empty lines.

assets/js/admin-script.js
```diff
@@ -825,7 +825,199 @@ jQuery( function( $ ) {
...
+			activeIndex = -1;
+			matches     = [];
+
+			if ( ! query ) {
+				$dropdown.hide();
+				return;
+			}
+
+			matches = settingItems.filter( function ( item ) {
+				return item.label.toLowerCase().indexOf( query.toLowerCase().trim() ) !== -1;
```
alexmigf, line 874: Wouldn't it be better to use IDs for navigation? We could pass a list of IDs and labels from the server side, use the labels for matching, and then navigate directly to the corresponding setting by ID.

assets/js/admin-script.js
```diff
@@ -825,7 +825,199 @@ jQuery( function( $ ) {
...
+			} );
+
+			if ( ! matches.length ) {
+				$dropdown.hide();
+				return;
+			}
+
+			matches.forEach( function ( item, index ) {
```
alexmigf: You could use `$.each()`

assets/js/admin-script.js
```diff
@@ -825,7 +825,199 @@ jQuery( function( $ ) {
...
+			} );
+		} );
+
+		if ( ! settingItems.length ) {
+			return;
+		}
+
+		// Create dropdown container.
+		const $dropdown = $( '<ul class="settings-search-dropdown"></ul>' );
+		$searchContainer.append( $dropdown );
```
alexmigf, line 848: Maybe you should first check whether this has already been created, just in case, or remove it before creating it.

assets/js/admin-script.js
```diff
@@ -825,7 +825,199 @@ jQuery( function( $ ) {
...
+				if ( activeIndex >= 0 && matches[ activeIndex ] ) {
+					navigateToSetting( matches[ activeIndex ] );
+				}
+			} else if ( 'Escape' === e.key ) {
+				$dropdown.hide();
+				$input.val( '' );
+			}
+		} );
+
+		$input.on( 'blur', function () {
```
alexmigf: Maybe this can introduce issues with the other input events?

alexmigf (changes requested):
The input looks slightly larger in my setup compared to the other fields. I’m not sure that’s ideal from a design consistency point of view.

Image

assets/js/admin-script.js
MohamadNateqi: How do you think it can cause issues for other input events?

alexmigf: It could hide the dropdown before it's clicked. It works now because `mousedown` runs before. It's not that is currently causing issues, but can be a bit fragile. Maybe this could be safer?

```js
$( document ).on( 'mousedown', function ( e ) {
	if ( ! $( e.target ).closest( '.settings-search' ).length ) {
		$dropdown.hide();
	}
} );
```

MohamadNateqi: I updated the code, but I couldn't understand how exactly the scenario you explained can happen. Can you please explain?

alexmigf: Yes, you just needed to comment this:

```js
$.each( matches, function ( index, item ) {
	const $item = $( '<li class="settings-search-item"></li>' )
		.text( item.label )
		.attr( 'data-index', index );
		// .on( 'mousedown', function ( e ) {
		// 	e.preventDefault();
		// 	navigateToSetting( item );
		// } );
	$dropdown.append( $item );
} );
```

alexmigf: I mean, there is no issue with your code right now, because `mousedown` runs before `blur`, so the navigation works as expected. It is just something we could improve to make it a bit more robust.

includes/Assets.php
```diff
@@ -374,10 +377,46 @@ public function backend_scripts_styles( $hook ) {
...
+	 * @param string $tab
+	 *
+	 * @return array
+	 */
+	private function get_settings_search_index( string $tab ): array {
+		$page = '';
+
+		switch ( $tab ) {
+			case 'general':
+			case '':
```
alexmigf, line 393: Can this be `default`?

alexmigf (changes requested):
Some of the dropdown options do not seem to navigate correctly, for example Extra Fields and Shop Fields.

includes/Assets.php
MohamadNateqi: It could work that way, but the default would return general settings for any value. I wanted it to return general settings only when `$tab` is empty, and to return an empty array for any invalid value instead.

Do you think the `default` would be better?

alexmigf: Makes sense to me.

alexmigf (approved):
Maybe we should have a label like "Choose document" for the document selector here:

Image

alexmigf (approved):
Love this feature. Nice work!

PR #1481, opened 2026-04-01
Fix: Handling of zero product/VAT categories in EN16931 UBL output

### LINKED ISSUE #1480: VAT category K is not fully supported in EN16931 UBL output
<p class="editor-paragraph-block">Invoices using VAT category <strong>K</strong> for intra-community supply are not being generated with all required EN16931 / Peppol BIS data.</p>

alexmigf:
closes #1480


alexmigf:
Passed both local testing and customer testing.

PR #1483, opened 2026-04-02
New V6: Performance optimizations and several improvements (PHP 8.1+)

### LINKED ISSUE #1482: New V6: Performance optimizations and several improvements (PHP 8.1+)
[Plane work item: WPOVE-325](https://work.teamupdraft.com/team-updraft/browse/WPOVE-325)

Use cachegrind profiles to identify performance bottlenecks in the plugin and highlight opportunities for improvement.

alexmigf:
closes #1482

## Summary of changes

- Raise minimum support to PHP 8.1, WordPress 5.3, and WooCommerce 4.0 for v6
- Update CI and tooling for newer PHP versions
- Refactor plugin bootstrapping and dependency loading to be more lazy and typed
- Introduce a dedicated `Notices` class and move admin/dependency notices into it
- Add request-level caching and refactor document loading/settings handling
- Add page-detection helpers and use them to conditionally initialize document-related functionality
- Reduce always-on compatibility loading by restructuring third-party integration hooks
- Clean up legacy/deprecated paths and move some notice/utility logic out of the main plugin class


wpo-ips-functions.php
```diff
@@ -103,29 +103,55 @@ function wcpdf_get_document( string $document_type, $order, bool $init = false )
...
 
-function wcpdf_get_invoice( $order, $init = false ) {
+/**
+ * Deprecated functions to get the document object for an order, kept for backward compatibility.
+ *
+ * @param mixed $order
+ * @param bool $init
+ * @return object|false
+ */
+function wcpdf_get_invoice( mixed $order, bool $init = false ): object|false {
```
MohamadNateqi: Maybe it's time to remove these deprecated functions?

wpo-ips-functions.php
```diff
@@ -211,12 +238,12 @@ function wcpdf_pdf_headers( string $filename, string $mode = 'inline', ?string $
...
+ * @param  \WPO\IPS\Documents\OrderDocument $document
  * @param  string $output_format
  * @param  string $error_handling
  * @return string|false
  */
-function wcpdf_get_document_file( object $document, string $output_format = 'pdf', string $error_handling = 'exception' ) {
+function wcpdf_get_document_file( \WPO\IPS\Documents\OrderDocument $document, string $output_format = 'pdf', string $error_handling = 'exception' ): string|false {
 	$default_output_format = 'pdf';
 
 	if ( ! $document ) {
```
MohamadNateqi: This is not needed anymore.

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -81,22 +106,94 @@ public function __construct() {
...
+	/**
+	 * Get a plugin class instance by slug.
+	 *
+	 * @param string $property
+	 *
+	 * @return object|null
+	 */
+	public function get_instance( string $property ): ?object {
+		$map = array(
+			'third_party_plugins' => ThirdPartyPlugins::class,
```
MohamadNateqi: Wouldn't it be better to use the class name as the key, so `get_instance()` can also be called like this: `get_instance( SomeClass::class );`?

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -81,22 +106,94 @@ public function __construct() {
...
+			'install'             => Install::class,
+			'font_synchronizer'   => FontSynchronizer::class,
+			'peppol'              => Peppol::class,
+			'notices'             => Notices::class,
+			'setup_wizard'        => SetupWizard::class,
+		);
 
-	public function is_dependency_version_supported( $dependency ) {
+		if ( ! isset( $map[ $property ] ) ) {
+			return null;
```
MohamadNateqi, line 285: What do you think about registering a log here?

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -81,22 +106,94 @@ public function __construct() {
...
+	
+	/**
+	 * Get a plugin class instance by slug.
+	 *
+	 * @param string $property
+	 *
+	 * @return object|null
+	 */
+	public function get_instance( string $property ): ?object {
+		$map = array(
```
MohamadNateqi: What about applying a filter to this array?

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -110,84 +207,41 @@ public function is_dependency_version_supported( $dependency ) {
...
-	 *
-	 * Note: the first-loaded translation file overrides any following ones if the same translation is present
+	 * 
+	 * @return void
 	 */
-	public function translations() {
-		$locale = $this->determine_locale();
-		$dir    = trailingslashit( WP_LANG_DIR );
+	public function translations(): void {
+		static $loaded = false;
```
MohamadNateqi, line 323: This static guard prevents reloading the translation that is being called in the `attach_document_to_email()` method in the `Main` class.

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -271,128 +310,228 @@ public function woocommerce_hpos_compatible(): void {
...
+	public function plugin_path(): string {
+		return untrailingslashit( plugin_dir_path( __FILE__ ) );
+	}
+	
+	/**
+	 * Check if the current admin page is the plugin settings page.
+	 *
+	 * @return bool
+	 */
+	public function is_settings_page(): bool {
```
MohamadNateqi: Would it be better to create a new Utility class and move some of these functions there?

includes/Admin.php
```diff
@@ -2258,6 +2357,65 @@ private function document_data_editing_disabled_notice( OrderDocument $document
...
+	}
+	
+	/**
+	 * Get cached document for the order list to avoid multiple instances of the same document during the same request.
+	 *
+	 * @param string $document_type
+	 * @param \WC_Abstract_Order $order
+	 * @return OrderDocument|null
+	 */
+	private function get_cached_order_document( string $document_type, \WC_Abstract_Order $order ): ?OrderDocument {
```
MohamadNateqi, line 2449: Can we have this caching in the Order Document class itself, in the original method?

includes/Admin.php
```diff
@@ -380,10 +263,16 @@ public function add_invoice_columns( $columns ) {
...
-	 * @param  string $column                 column slug
-	 * @param  string $post_or_order_object   object
+	 * 
+	 * @param  string $column              column slug
+	 * @param  \WC_Abstract_Order|\WP_Post $post_or_order_object
+	 * @return void
 	 */
-	public function invoice_columns_data( $column, $post_or_order_object ) {
+	public function invoice_columns_data( string $column, object $post_or_order_object ): void {
+		if ( ! $this->invoice_columns_enabled() ) {
```
MohamadNateqi, line 272: Wouldn't it be better to keep this check in the constructor as before? Because it'll be called N (number of orders in the page) times in the Orders page, which is sort of unnecessary when disabled.

includes/Main.php
```diff
@@ -12,92 +13,81 @@
...
 	 * Attach document to WooCommerce email
+	 * 
+	 * @param array              $attachments
+	 * @param string             $email_id
+	 * @param \WC_Abstract_Order $order
+	 * @param \WC_Email|null     $email
+	 * @return array
 	 */
-	public function attach_document_to_email( $attachments, $email_id, $order, $email = null ) {
+	public function attach_document_to_email( array $attachments, string $email_id, \WC_Abstract_Order $order, ?\WC_Email $email = null ): array {
```
MohamadNateqi: The third parameter object might be something else than `WC_Abstract_Order`, it might be better to remove it?

includes/Main.php
```diff
@@ -390,7 +403,7 @@ public function generate_document_ajax(): void {
 
 		// debug enabled by URL
 		if ( isset( $request['debug'] ) && ! ( is_user_logged_in() || isset( $request['my-account'] ) ) ) {
-			$this->enable_debug();
+			$this->maybe_enable_debug( true );
```
MohamadNateqi: I think this will show debug errors for anonymous visitors in the my account page.

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -81,22 +106,94 @@ public function __construct() {
...
+		);
 
-	public function is_dependency_version_supported( $dependency ) {
+		if ( ! isset( $map[ $property ] ) ) {
+			return null;
+		}
+
+		if ( null === $this->{$property} ) {
+			$class_name = $map[ $property ];
+			$this->{$property} = $class_name::instance();
```
MohamadNateqi, line 290: There is no property defined for the setup wizard.

includes/SetupWizard.php
```diff
@@ -285,27 +299,44 @@ public function setup_wizard_footer() {
...
-
-	public function get_step( $delta ) {
-		$step_keys = array_keys( $this->steps );
+	/**
+	 * Get the next or previous step key.
+	 *
+	 * @param int $delta Direction to move in the steps array (1 for next, -1 for previous).
+	 * @return string|false The key of the next/previous step, or false if there is no such step.
+	 */
+	public function get_step( int $delta ) {
```
MohamadNateqi: Return type.

includes/Settings.php
```diff
@@ -17,14 +18,14 @@
...
 
-	/** @var string|false */
+	public ?SettingsCallbacks $callbacks = null;
+	public ?SettingsGeneral $general     = null;
+	public ?SettingsDocuments $documents = null;
+	public ?SettingsDebug $debug         = null;
+	public ?SettingsUpgrade $upgrade     = null;
+	public ?SettingsEDI $edi             = null;
+	
 	public $options_page_hook;
```
MohamadNateqi: Type hint.

includes/Settings.php
```diff
@@ -33,74 +34,93 @@ class Settings {
...
 	}
+	
+	/**
+	 * Get a settings instance by slug.
+	 *
+	 * @param string $setting
+	 * @return object|null
+	 */
+	public function get_instance( string $setting ): ?object {
+		$map = array(
```
MohamadNateqi: Would it be better to put the `$map` out of this method? Since there is no need to construct this array each time this method is called.

includes/Settings.php
```diff
@@ -546,7 +604,14 @@ public function get_common_document_settings(): array {
...
 
-	public function get_document_settings( $document_type, $output_format = 'pdf' ) {
+	/**
+	 * Get document settings by document type and output format.
+	 *
+	 * @param string $document_type
+	 * @param string $output_format
+	 * @return array|false
+	 */
+	public function get_document_settings( string $document_type, string $output_format = 'pdf' ) {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -278,12 +276,26 @@ public function get_document_pdf_attachment( $document, $tmp_path ) {
...
 
-	public function get_document_xml_attachment( $document, $tmp_path ) {
+	/**
+	 * Get the document XML attachment path, write the file if it doesn't exist and return the path
+	 *
+	 * @param OrderDocument $document
+	 * @param string        $tmp_path
+	 * @return string|false
+	 */
+	public function get_document_xml_attachment( OrderDocument $document, string $tmp_path ) {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -717,25 +740,36 @@ public function get_tmp_base ( $append_random_string = true ) {
...
 	}
 
 	/**
-	 * Get WordPress uploads folder base
+	 * Get WordPress uploads folder base.
+	 *
+	 * @return string|false
 	 */
-	public function get_wp_upload_base () {
+	public function get_wp_upload_base() {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -854,171 +916,138 @@ public function get_random_string () {
...
 
-		$tmp_base = $this->get_tmp_base(); // get tmp base
+		$tmp_base             = $this->get_tmp_base();
+		$file_system_instance = WPO_WCPDF()->get_instance( 'file_system' );
 
-		// create plugin base temp folder
-		if ( ! WPO_WCPDF()->file_system->is_dir( $tmp_base ) ) {
-			$dir = WPO_WCPDF()->file_system->mkdir( $tmp_base );
+		if ( false === $tmp_base ) {
+			return false;
```
MohamadNateqi, line 1031: I think it might be good to keep the logs here.

includes/Settings.php
```diff
@@ -750,36 +849,62 @@ public function get_template_list_cache() {
...
+	 * Callback for when debug settings are updated, to check if we need to flush rewrite rules for pretty document links.
+	 *
+	 * @param array $old_settings
+	 * @param array $settings
+	 * @param string $option
+	 * @return void
+	 */
+	public function debug_settings_updated( array $old_settings, array $settings, string $option ): void {
+		if ( empty( $old_settings['pretty_document_links'] ) && ! empty ( $settings['pretty_document_links'] ) ) {
 			flush_rewrite_rules();
```
MohamadNateqi: Doesn't it create a timing issue flushing the rewrite rule here, instead of deferring it using transient?

includes/Documents.php
```diff
@@ -9,40 +9,37 @@
...
 	}
 
 	/**
-	 * Constructor for the document class hooks in all documents that can be created.
+	 * Constructor.
 	 *
 	 */
 	public function __construct() {
-		add_action( 'init', array( $this, 'init' ), 15 ); // after regular 10 actions but before most 'follow-up' actions (usually 20+)
+		if ( WPO_WCPDF()->is_order_page() || WPO_WCPDF()->is_settings_page() || WPO_WCPDF()->is_account_page() ) {
```
MohamadNateqi: This constructor runs on `init`, and the `is_order_page()` function relies on the `get_current_screen()`.  Also, `is_account_page()` relies on `is_account_page().
Are these functions defined and have a value at that time?

includes/Settings/SettingsDebug.php
MohamadNateqi, line 372: Maybe you can add `@see ...` at the top of this line, since it directly links the function to the location they get called.

includes/Documents/Invoice.php
```diff
@@ -7,16 +7,12 @@
...
-
 class Invoice extends OrderDocumentMethods {
 
 	/**
 	 * Init/load the order object.
 	 *
-	 * @param  int|object|WC_Order $order Order to init.
+	 * @param int|object|null $order Order to init.
 	 */
 	public function __construct( $order = 0 ) {
```
MohamadNateqi, line 17: Type hint

includes/Documents/OrderDocument.php
```diff
@@ -9,97 +9,29 @@
...
+	protected array $data                    = array();
+	protected array $resolved_settings_cache = array();
 
 	/**
 	 * Init/load the order object.
 	 *
-	 * @param  int|object|WC_Order $order Order to init.
+	 * @param int|object|null $order Order to init.
 	 */
 	public function __construct( $order = 0 ) {
```
MohamadNateqi, line 36: Type hint

includes/Compatibility/ThirdPartyPlugins.php
```diff
@@ -160,293 +113,118 @@ public function wcs_renewal_order_meta( $meta, $to_order, $from_order ) {
...
-			if ( absint( $order_item_id ) === $item_id ) {
-				$product    = $order_item->get_product();
-				$bundled_by = $order_item->get_meta( '_bundled_by', true );
-				break;
-			}
-		}
-
-		if ( empty( $product ) || ! is_a( $product, 'WC_Product' ) ) {
-			return $classes;
+		if ( $this->has_wc_bundles ) {
```
MohamadNateqi: Where is this property declared and assigned? Same for the rest.

wpo-ips-functions.php
```diff
@@ -1030,10 +1056,10 @@ function wpo_wcpdf_get_simple_template_default_table_headers( $document ): array
 /**
  * Get the WP_Filesystem instance
  *
- * @return WP_Filesystem|false
+ * @return \WP_Filesystem|false
  * @throws RuntimeException
  */
-function wpo_wcpdf_get_wp_filesystem() {
+function wpo_wcpdf_get_wp_filesystem(): \WP_Filesystem|false {
```
MohamadNateqi: Does the `WP_Filesystem` class exist in WP? Or is it only a function?

MohamadNateqi (changes requested):

woocommerce-pdf-invoices-packingslips.php
alexmigf: I see the point, but in this case the key represents the legacy property name where the instance is stored. So the map is not only resolving classes, it is also preserving compatibility with properties like settings, documents, etc. We could support `SomeClass::class` as an alias later, but I’d keep the property map as the source of truth for now.

woocommerce-pdf-invoices-packingslips.php
alexmigf: Good catch!

includes/Admin.php
alexmigf: Thanks for the suggestion. In this case I’d prefer to leave it as-is for now. The check is already cached, so the impact should be very limited.

includes/Documents/Invoice.php
alexmigf: This one is a bit tricky, so I’d prefer not to tackle it right now, for the reasons I explained on Slack.

includes/Documents/OrderDocument.php
alexmigf: This one is a bit tricky, so I’d prefer not to tackle it right now, for the reasons I explained on Slack.

wpo-ips-functions.php
alexmigf: Function, my bad.

wpo-ips-functions.php
alexmigf: I don’t think we need to remove it. I’ve moved it to the dedicated deprecated functions file instead.

includes/Admin.php
alexmigf: I’d keep this outside `OrderDocument`, because it is request-level lookup caching rather than behavior of an individual document instance.

includes/Settings.php
alexmigf: We’re doing the same in the main plugin file, and honestly I don’t see a significant performance downside here. I think this is fine for now.

YordanSoares (commented):
## Test results

### Document rendering

I have tested all our documents, including the premium ones from [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) & [Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/), generated with the "Simple" template as well as the premium templates included in the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/) (Simple Premium, Business, and Modern), and all of them were correctly generated, although I had issues when activating the mPDF extension (see next section).

| Document Types ↓ / Templates →     | Simple | Simple Premium<sup>1</sup> | Business<sup>1</sup> | Modern<sup>1</sup> |
|---------------------------------|:------:|:----------------:|:---------:|:-------:|
| Invoice                         | ✅     | ✅               | ✅        | ✅      |
| Packing Slip                    | ✅     | ✅               | ✅        | ✅      |
| Credit Note<sup>2</sup>      | ✅     | ✅               | ✅        | ✅      |
| Proforma Invoice<sup>2</sup> | ✅     | ✅               | ✅        | ✅      |
| Receipt<sup>2</sup>          | ✅     | ✅               | ✅        | ✅      |
| Delivery Note<sup>2</sup>          | ✅     | ✅               | ✅        | ✅      |
| Proposal<sup>3</sup>                        | ✅     | ✅               | ✅        | ✅      |
| Order Confirmation<sup>3</sup>              | ✅     | ✅               | ✅        | ✅      |

<sup>1</sup> Included in the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/).
<sup>2</sup> Included in the [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) extension and the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/).
<sup>3</sup> Included in the [Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/) extension and the [Ultimate Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-ultimate-bundle/).

### Issues found

#### mPDF extension

While activating the mPDF extension, the images in the document were not printed, and also the styles changes compared with the latest release. 

Here is animated screenshot to compare the output between `5.13.0` and `6.0.0-i1482.2` while using the Modern template (in both environment I have installed IPS mPDF 3.1.0 + IPS Premium Templates 2.27.0):

Image

#### Fatal error on manually update

I was able to update the plugin manually while using PHP 8.0.30:

Image

And this caused a fatal error in the site:

```php
Fatal error: Uncaught RuntimeException: Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.1.0". You are running 8.0.30. in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\vendor\composer\platform_check.php:22 
Stack trace:
#0 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\vendor\composer\autoload_real.php(25): require()
#1 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\vendor\autoload.php(22): ComposerAutoloaderInit799089b7d5c8c7fc7a692559c75d6491::getLoader()
#2 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\woocommerce-pdf-invoices-packingslips.php(89): require('C:\\Users\\Yordan...')
#3 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\woocommerce-pdf-invoices-packingslips.php(80): WPO_WCPDF->__construct()
#4 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\woocommerce-pdf-invoices-packingslips.php(555): WPO_WCPDF::instance()
#5 \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\woocommerce-pdf-invoices-packingslips.php(558): WPO_WCPDF()
#6 \wp-settings.php(589): include_once('C:\\Users\\Yordan...')
#7 \wp-config.php(109): require_once('C:\\Users\\Yordan...')
#8 \wp-load.php(50): require_once('C:\\Users\\Yordan...')
#9 \wp-admin\admin.php(35): require_once('C:\\Users\\Yordan...')
#10 {main} thrown in \wp-content\plugins\woocommerce-pdf-invoices-packing-slips\vendor\composer\platform_check.php on line 22
```

Probably we need to add an extra check before trying to manually update the plugin to avoid completing the process and break the site?

## Test outputs
- Generated documents: [6.0.0-i1482.2.zip](https://github.com/user-attachments/files/28518638/6.0.0-i1482.2.zip)

alexmigf:
@YordanSoares Could you test with this version of the mPDF instead please? https://github.com/wpovernight/woocommerce-pdf-ips-mpdf/pull/58

YordanSoares (commented):
@alexmigf I have tested with https://github.com/wpovernight/woocommerce-pdf-ips-mpdf/pull/58, and the documents were correctly generated now:

Image

I'd say that everything is working as it should. The only thing left to do, as far as user-facing functionality goes, is look into the fatal error I mentioned when updating manually.

alexmigf:
@YordanSoares Brilliant, thank you!

includes/Documents/OrderDocumentMethods.php
```diff
@@ -791,52 +943,55 @@ public function get_tax_rate_by_id( $rate_id, $order = null ) {
...
-					// so we use a fallback if the rate_percent = 0 and the amount != 0
-					$rate_percent = $tax_item->get_rate_percent();
-					$tax_amount = $tax_item->get_tax_total() + $tax_item->get_shipping_tax_total();
-					if ( $tax_amount > 0 && $rate_percent > 0 ) {
-						$tax_rates[ $tax_item->get_rate_id() ] = $rate_percent;
-					} else {
-						continue; // not setting the rate will let the plugin fall back to the rate from the settings
-					}
+		foreach ( $tax_items as $tax_item_key => $tax_item ) {
+			if ( is_callable( array( $order, 'get_created_via' ) ) && 'subscription' === $order->get_created_via() ) {
```
MohamadNateqi: Maybe we can remove this `if`, and only keep the statement of the `else`?

MohamadNateqi:
@alexmigf Can you please merge main?

includes/Admin.php
```diff
@@ -1519,23 +1482,41 @@ public function save_invoice_number_date( $order_id, $order ) {
...
 
 	/**
 	 * Send emails manually
+	 * 
+	 * @param int $post_or_order_id The order ID or post ID.
+	 * @param \WC_Abstract_Order|\WP_Post $post_or_order_object The order object or order ID.
+	 * @return void
 	 */
-	public function send_emails( $post_or_order_object_id, $post_or_order_object ) {
+	public function send_emails( int $post_or_order_id, object $post_or_order_object ): void {
```
MohamadNateqi, line 1553: Variable names changed, but were not reflected in the code.

includes/Settings/SettingsDebug.php
```diff
@@ -297,89 +330,23 @@ public function get_additional_invoice_number_store_document_types() {
...
-		wp_send_json_success( compact( 'message' ) );
-	}
-
 	/**
 	 * AJAX handler for processing settings debug tools.
 	 *
+	 * Dispatches to one of the internal debug tool handlers.
+	 *
 	 * @return void
+	 * @see self::export_settings()
```
MohamadNateqi, line 339: Maybe `@uses` is more proper here?

wpo-ips-deprecated-functions.php
```diff
@@ -127,9 +127,57 @@ function wpo_ips_edi_maybe_save_order_customer_peppol_data( \WC_Abstract_Order $
  *
  * @return bool
  * 
- * @deprecated 5.8.0 Use \WPO_WCPDF()->vat_plugins->has_active() instead.
+ * @deprecated 5.8.0 Use \WPO_WCPDF()->get_instance( 'vat_plugins' )->has_active() instead.
  */
```
MohamadNateqi, line 131: Maybe we need to consider a date/version at which this method won't be available anymore, and then we remove it from here, otherwise we end up collecting all depracted functions here?

alexmigf:
@MohamadNateqi Done.

alexmigf:
@YordanSoares The issue you experienced should be fixed now. Please test it with [this version](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/releases/download/v6.0.0-i1482.3/woocommerce-pdf-invoices-packing-slips.6.0.0-i1482.3.zip).

includes/Main.php
```diff
@@ -651,39 +668,44 @@ private function load_template_functions() {
...
 		}
 
 		if ( empty( $type ) ) {
-			return $tmp_base;
+			$this->tmp_path_cache[ $cache_key ] = $tmp_base;
+			return $this->tmp_path_cache[ $cache_key ];
 		}
 
 		switch ( $type ) {
```
MohamadNateqi: Maybe using `match()`?

wpo-ips-functions.php
```diff
@@ -242,9 +248,12 @@ function wcpdf_get_document_file( object $document, string $output_format = 'pdf
...
 	}
 
-	$tmp_path = WPO_WCPDF()->main->get_tmp_path( 'attachments' );
+	$main_instance        = WPO_WCPDF()->get_instance( 'main' );
+	$file_system_instance = WPO_WCPDF()->get_instance( 'file_system' );
+	
+	$tmp_path             = $main_instance->get_tmp_path( 'attachments' );
 
-	if ( ! WPO_WCPDF()->file_system->is_dir( $tmp_path ) || ! WPO_WCPDF()->file_system->is_writable( $tmp_path ) ) {
+	if ( ! $file_system_instance->is_dir( $tmp_path ) || ! $file_system_instance->is_writable( $tmp_path ) ) {
```
MohamadNateqi: Since `get_tmp_path()` no longer creates the directory if it doesn't exist, should we consider creating it here to avoid silent failures of email attachments?

includes/Main.php
```diff
@@ -2095,8 +2271,68 @@ public function apply_ink_saving_styles( string $css, OrderDocument $document ):
...
+	 */
+	private function ensure_tmp_path_exists( string $path ): bool {
+		$file_system_instance = WPO_WCPDF()->get_instance( 'file_system' );
+
+		if ( empty( $path ) ) {
+			return false;
+		}
+
+		if ( ! $file_system_instance->is_dir( $path ) ) {
+			$dir = $file_system_instance->mkdir( $path );
```
MohamadNateqi: This method creates the dir, but compared to `init_tmp()`, it doesn't create the `.htaccess` and `index.php`.

includes/Documents/OrderDocumentMethods.php
```diff
@@ -791,52 +943,55 @@ public function get_tax_rate_by_id( $rate_id, $order = null ) {
...
-			$tax_items = $order->get_items( array('tax') );
+	/**
+	 * Get tax rates from the order's tax items
+	 * 
+	 * @param \WC_Abstract_Order|null $order
+	 * @return array|bool array of rate_id => rate_percent, or false if not available
+	 */
+	public function get_tax_rates_from_order( ?\WC_Abstract_Order $order = null ): array|bool {
+		$tax_rates = array();
+		$tax_items = $order->get_items( array( 'tax' ) );
```
MohamadNateqi, line 1021: `$order` can be `null`.

includes/Documents/OrderDocument.php
```diff
@@ -1653,34 +1946,45 @@ public function get_pdf() {
 
 		return apply_filters( 'wpo_wcpdf_get_pdf', $pdf, $this );
```
MohamadNateqi: Given that all methods now have a strict return type, maybe we should consider casting for those that return values coming from a filter.

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -111,83 +225,45 @@ public function is_dependency_version_supported( $dependency ) {
 	}
 
 	/**
-	 * Define constant if not already set
-	 * @param  string $name
-	 * @param  string|bool $value
+	 * Load the translation / textdomain files
+	 * 
+	 * @param bool $$force_reload
```
MohamadNateqi: Double `$`

YordanSoares (approved):
@alexmigf I confirmed that the update process is interrupted if the PHP version is below 8.1:

image

includes/Settings/SettingsUpgrade.php
```diff
@@ -191,10 +212,10 @@ public function extension_overview( $tab, $section ) {
...
-	 * @param  string  $extension  can be 'pro' or 'templates'
-	 * @return boolean
+	 * @param string $extension  can be 'pro' or 'templates'
+	 * @return bool
 	 */
-	public function extension_is_enabled( $extension ) {
+	public function extension_is_enabled( string $extension ): bool {
 		$is_enabled = false;
 
 		if ( ! empty( $extension ) || ! in_array( $extension, $this->extensions ) ) {
```
MohamadNateqi: Should this be `&&`?

includes/Documents/BulkDocument.php
```diff
@@ -116,7 +102,13 @@ public function get_pdf() {
...
 	}
 
-	public function get_html() {
+	/**
+	 * Get the HTML content.
+	 *
+	 * @param array $args
+	 * @return string
+	 */
+	public function get_html( array $args = array() ): string {
```
MohamadNateqi, line 111: The `$args` is unused and can be removed.

edi/Peppol.php
```diff
@@ -29,7 +29,7 @@ public static function instance(): self {
 	 * Constructor
 	 */
 	public function __construct() {
-		if (  ! wpo_ips_edi_peppol_is_available() ) {
+		if ( ! \wpo_ips_is_frontend_page_request() ) {
```
MohamadNateqi, line 45: What if a request is coming from a REST API?

wpo-ips-functions.php
```diff
@@ -242,9 +248,12 @@ function wcpdf_get_document_file( object $document, string $output_format = 'pdf
 		return wcpdf_error_handling( $error_message, $error_handling, true, 'critical' );
```
MohamadNateqi, line 266: Maybe we should include throw tags in the docblock as well.

wpo-ips-functions.php
```diff
@@ -2345,3 +2330,267 @@ function wpo_ips_get_bacs_account_options(): array {
...
+}
+
+/**
+ * Check if this is a frontend page request (not admin, ajax, cron, rest or wp-cli)
+ *
+ * @return bool
+ */
+function wpo_ips_is_frontend_page_request(): bool {
+	return ! is_admin()
+		&& ! wp_doing_ajax()
```
MohamadNateqi, line 2596: What if, in the future, we use this for AJAX requests from the frontend? This will return `false`, and I thought it might be confusing.

woocommerce-pdf-invoices-packingslips.php
```diff
@@ -271,128 +332,140 @@ public function woocommerce_hpos_compatible(): void {
...
+	 *
+	 * @param array  $unstable The unstable version data.
+	 * @param string $owner    GitHub repo owner.
+	 * @param string $repo     GitHub repo name.
+	 * @return void
 	 */
-	public function in_plugin_update_message( $args ) {
-		$transient_name = 'wpo_wcpdf_upgrade_notice_' . $args['Version'];
+	public function set_new_unstable_version_available_option( array $unstable, string $owner, string $repo ): void {
+		$debug_settings = $this->settings->debug_settings;
```
MohamadNateqi: Maybe it's better to use `get_instance()` here as well, to have a single source for getting classes?

includes/Main.php
```diff
@@ -651,39 +668,44 @@ private function load_template_functions() {
...
 	/**
-	 * Return tmp path for different plugin processes
+	 * Return tmp path for different plugin processes.
+	 *
+	 * @param string $type
+	 * @return string|false
 	 */
-	public function get_tmp_path( $type = '' ) {
-		$tmp_base = $this->get_tmp_base();
+	public function get_tmp_path( string $type = '' ) {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -705,41 +727,42 @@ public function get_tmp_path( $type = '' ) {
...
+	
 	/**
-	 * return the base tmp folder (usually uploads)
+	 * Return the base tmp folder path.
+	 *
+	 * @param bool $append_random_string Whether to append the random string to the temp folder name.
+	 * @return string|false
 	 */
-	public function get_tmp_base ( $append_random_string = true ) {
+	public function get_tmp_base( bool $append_random_string = true ) {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -860,19 +893,48 @@ public function maybe_reinstall_fonts( bool $force = false ): void {
...
+		$this->clear_tmp_path_caches();
+	}
+	
+	/**
+	 * Regenerate random string and copy contents to new tmp folder
+	 *
+	 * @param bool $reinstall_fonts
+	 * @return string|false
+	 */
+	public function regenerate_random_string( bool $reinstall_fonts = true ) {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -860,19 +893,48 @@ public function maybe_reinstall_fonts( bool $force = false ): void {
...
+		
+		return $new_path;
 	}
 
 	/**
 	 * Get random string
+	 * 
+	 * @return string|false
 	 */
 	public function get_random_string () {
```
MohamadNateqi: Return type.

includes/Main.php
```diff
@@ -2095,8 +2271,68 @@ public function apply_ink_saving_styles( string $css, OrderDocument $document ):
...
 		return $css;
 	}
+	
+	/**
+	 * Ensure a temporary path exists and is writable.
+	 *
+	 * @param string $path
+	 * @return bool
+	 */
+	private function ensure_tmp_path_exists( string $path ): bool {
```
MohamadNateqi: Maybe `dir` instead of `path` in the method name?

includes/Frontend.php
```diff
@@ -393,7 +393,7 @@ static function ( $value, string $group, \WC_Data $wc_object ) {
 	 * @param string $key Field key.
 	 * @param mixed $value Field value.
 	 * @param string $group Group name.
-	 * @param object $wc_object WC object (e.g. order).
+	 * @param \WC_Customer|\WC_Order $wc_object
 	 * @return void
 	 */
 	public function checkout_field_save_checkout_block_field( string $key, $value, string $group, object $wc_object ): void {
```
MohamadNateqi: Maybe `mixed` type hint for `$value`?

includes/Documents.php
```diff
@@ -101,7 +105,14 @@ public function get_documents( $filter = 'enabled', $output_format = 'pdf' ) {
...
 
-	public function get_document( $document_type, $order ) {
+	/**
+	 * Return an instance of the document class for a given document type and order.
+	 *
+	 * @param string $document_type
+	 * @param int|object|\WC_Order $order
+	 * @return OrderDocument|false
+	 */
+	public function get_document( string $document_type, $order ) {
```
MohamadNateqi: Type hint and return type missing.

includes/Documents.php
```diff
@@ -9,40 +9,45 @@
...
 	}
 
 	/**
-	 * Constructor for the document class hooks in all documents that can be created.
+	 * Constructor.
 	 *
 	 */
 	public function __construct() {
-		add_action( 'init', array( $this, 'init' ), 15 ); // after regular 10 actions but before most 'follow-up' actions (usually 20+)
+		if (
```
MohamadNateqi: Can these strict conditions cause issues for third parties?
Would it be better to register the hook all the time, as before, or to add a filter that allows third parties to modify it?

MohamadNateqi (changes requested):

includes/Admin.php
alexmigf: Could you please point me to where exactly you mean?

includes/Settings/SettingsDebug.php
alexmigf: Because the relationship is indirect and depends on `$_POST['debug_tool']`, `@see` feels more accurate.

wpo-ips-deprecated-functions.php
alexmigf: I don’t see this as a problem right now, but we can definitely discuss it again in the future.

includes/Admin.php
MohamadNateqi: Sorry, wrong line. It's in the `save_invoice_number_date()` method, here: https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1483/changes/BASE..8d13209f6ffe2f996a6ac500c8e9d14f1af07119#diff-a3d4b2d1929fa32a08a5eb4b357e9edd3ad86794fbff72f5b1fe7110a7551dfaR1415

assets/css/settings-styles.css
```diff
@@ -200,88 +200,93 @@ table.wcpdf_documents_settings_list td.title {
...
-	margin-bottom: 0.6em;
-	display: inline-block;
+	color: #3c434a;
 }
 
 .wcpdf_document_settings_sections > h2 {
-	cursor: pointer;
-	padding: 1em 0.8em;
+	margin: 0;
+	padding: 14px 16px;
```
MohamadNateqi: Keeping `em` as the unit is generally better compared to pixel, as it scales relative to the font size, making the design more flexible.

includes/Documents/BulkDocument.php
alexmigf: I know, but I kept it aligned with the `OrderDocument` method for consistency.

edi/Peppol.php
alexmigf: I think this should be handled by the same function.

MohamadNateqi: I meant we need to allow Peppol to work with REST APIs, right? If so, then the `wpo_ips_is_frontend_page_request` prevents that.

wpo-ips-functions.php
alexmigf: Good point. This helper is only for normal frontend page loads, so AJAX returning `false` is expected.

edi/Peppol.php
alexmigf: Good catch!

wpo-ips-functions.php
MohamadNateqi: Would it be helpful to include that in the docBlock?

alexmigf: I think I did.

MohamadNateqi: You're right. My bad, I missed it.

includes/Documents/OrderDocument.php
alexmigf: Great idea!

assets/css/settings-styles.css
```diff
@@ -200,88 +200,93 @@ table.wcpdf_documents_settings_list td.title {
...
-	color: white;
+	border-bottom: none;
 }
 
 .wcpdf_document_settings_sections ul li a {
-	color: black;
-	text-decoration: none;
-	padding: 1.2em 1.6em;
 	display: block;
+	padding: 10px 16px;
```
MohamadNateqi: There are some more cases that `em` replaced with `px`. It would be good to change them as well.

edi/Peppol.php
```diff
@@ -29,34 +29,32 @@ public static function instance(): self {
...
-			return;
-		}
-		
-		// Peppol My Account
-		add_filter( 'woocommerce_account_menu_items', array( $this, 'peppol_account_menu_item' ), 10, 2 );
+		// REST / Store API / checkout processing hooks.
 		add_action( 'rest_api_init', array( $this, 'peppol_register_checkout_autofill_endpoint_route' ) );
+
+		$this->peppol_display_checkout_block_fields();
+		$this->peppol_set_checkout_block_fields_value();
```
MohamadNateqi, line 36: Maybe the `wpo_ips_edi_peppol_is_available()` is better to be called early in the `peppol_set_checkout_block_fields_value()` method, since `wpo_ips_edi_peppol_is_available()` is removed from the `__construct`?

edi/Peppol.php
```diff
@@ -29,34 +29,32 @@ public static function instance(): self {
...
+
+		add_action( 'woocommerce_set_additional_field_value', array( $this, 'peppol_save_checkout_block_fields' ), 10, 4 );
+		add_action( 'woocommerce_store_api_checkout_order_processed', array( $this, 'peppol_remove_order_checkout_block_fields_meta' ), 10, 1 );
```
MohamadNateqi, line 39: Maybe the same about this?

edi/Peppol.php
```diff
@@ -29,34 +29,32 @@ public static function instance(): self {
...
+		if ( ! \wpo_ips_is_frontend_page_request() ) {
+			return;
+		}
+
+		// Pure frontend page hooks only.
+		add_filter( 'woocommerce_account_menu_items', array( $this, 'peppol_account_menu_item' ), 10, 2 );
 		add_action( 'template_redirect', array( $this, 'save_peppol_settings' ) );
 		add_action( 'woocommerce_account_peppol_endpoint', array( $this, 'peppol_settings_account_page' ) );
 
 		add_rewrite_endpoint( 'peppol', EP_PAGES );
```
MohamadNateqi, line 54: Should it come after the REST hook?

includes/Documents/OrderDocumentMethods.php
```diff
@@ -7,39 +7,54 @@
...
 		}
 
 		$parent_order_id = $this->get_refund_parent_id( $order );
-		$order = wc_get_order( $parent_order_id );
-		return $order;
+		if ( ! $parent_order_id ) {
+			return null;
+		}
+		
+		return wc_get_order( $parent_order_id );
```
MohamadNateqi: The `wc_get_order()` can return `false`.

includes/Documents/OrderDocumentMethods.php
MohamadNateqi: Maybe this block can be removed?

MohamadNateqi: Same.

includes/Documents/OrderDocumentMethods.php
```diff
@@ -264,9 +346,22 @@ public function get_custom_field( $field_name ) {
 			}
 		}
 
-		return apply_filters( 'wpo_wcpdf_billing_custom_field', $custom_field, $this );
+		return apply_filters(
+			'wpo_wcpdf_billing_custom_field',
+			$custom_field,
```
MohamadNateqi, line 366: Can this variable be undefined when reaching here?

includes/Documents/OrderDocumentMethods.php
MohamadNateqi, line 1003: Should we consider removing this? Since it seems it was for WC before 3.7.

includes/Documents/OrderDocumentMethods.php
```diff
@@ -1055,56 +1281,92 @@ public function get_order_shipping( $tax = 'excl' ) { // set $tax to 'incl' to i
...
 		}
 
-		$discount = array (
-			'label'		=> __( 'Discount', 'woocommerce-pdf-invoices-packing-slips' ),
-			'value'		=> $this->format_price( $discount_value ),
-			'raw_value'	=> $discount_value,
-		);
+		$discount = null;
 
 		if ( round( $discount_value, 3 ) != 0 ) {
```
MohamadNateqi, line 1345: Would it be better to replace `3` with `wc_get_price_decimals()`?

includes/Admin.php
```diff
@@ -380,11 +263,17 @@ public function add_invoice_columns( $columns ) {
...
-	 * @param  string $column                 column slug
-	 * @param  string $post_or_order_object   object
+	 * 
+	 * @param  string $column              column slug
+	 * @param  \WC_Abstract_Order|\WP_Post $post_or_order_object
+	 * @return void
 	 */
-	public function invoice_columns_data( $column, $post_or_order_object ) {
-		if ( ! in_array( $column, array( 'invoice_number_column', 'invoice_date_column' ) ) ) {
+	public function invoice_columns_data( string $column, object $post_or_order_object ): void {
```
MohamadNateqi: For non-HPOS, the second arg is `int`:

```php
Uncaught Error: WPO\IPS\Admin::invoice_columns_data(): Argument #2 ($post_or_order_object) must be of type object, int given, called in ...\wp-includes\class-wp-hook.php on line 341
in ...\wp-content\plugins\woocommerce-pdf-invoices-packing-slips\includes\Admin.php on line 271
```

includes/Admin.php
```diff
@@ -1431,111 +1378,159 @@ public function output_number_date_edit_fields( $document, $data ): void {
...
 	 * but we don't want to store the settings for uninitialized documents.
 	 * Only use in frontend/backed (page requests), otherwise settings will never be stored!
+	 * 
+	 * @return void
 	 */
-	public function disable_storing_document_settings() {
-		add_filter( 'wpo_wcpdf_document_store_settings', array( $this, 'return_false' ), 9999 );
-	}
+	public function disable_storing_document_settings(): void {
+		if ( $this->storing_document_settings_disabled ) {
```
MohamadNateqi, line 1514: Could the cached flag get out of sync here?
Both `Admin` and `Frontend` toggle the same filter. If `Admin` disables storing, then `Frontend` re-enables it, `Admin`'s flag still reads "disabled". 
So the next `disable_storing_document_settings()` call early returns and never re-adds the filter, leaving it inactive when Admin expects it to be active. 
Is it a possible scenario?

includes/Main.php
```diff
@@ -12,120 +13,109 @@
...
+	 * @param string          $email_id
+	 * @param mixed           $order
+	 * @param \WC_Email|null  $email
+	 * @return array
 	 */
-	public function attach_document_to_email( $attachments, $email_id, $order, $email = null ) {
+	public function attach_document_to_email( array $attachments, string $email_id, mixed $order, ?\WC_Email $email = null ): array {
 		// check if all variables properly set
-		if ( ! is_object( $order ) || ! isset( $email_id ) ) {
+		if ( ! $order instanceof \WC_Abstract_Order || ! isset( $email_id ) ) {
```
MohamadNateqi: This change in the guard breaks the `WooCommerce Booking compatibility`. On line `111` we have compatibility for `wc_booking`.

includes/Main.php
```diff
@@ -826,55 +883,100 @@ public function tmp_subfolder_has_files( string $subfolder ): bool {
...
+	 */
+	public function regenerate_random_string( bool $reinstall_fonts = true ): string|false {
+		if ( ! empty( $this->get_random_string() ) ) {
+			$old_path = $this->get_tmp_base();
+		} else {
+			$old_path = $this->get_tmp_base( false );
+		}
+
+		$this->generate_random_string();
+		$new_path = $this->get_tmp_base();
```
MohamadNateqi, line 980: Maybe you need to check the value of `$old_path` and `$new_path`, since the return value can be `false` as well?

includes/Main.php
```diff
@@ -1992,23 +2219,24 @@ public function add_document_link_to_address_section( string $type, \WC_Abstract
 	 * @param bool $sent_to_admin
 	 * @param bool $plain_text
 	 * @param mixed|\WC_Email $email Some third-parties might send different values.
-	 *
 	 * @return void
 	 */
 	public function add_document_link_to_email( \WC_Abstract_Order $order, bool $sent_to_admin, bool $plain_text, $email ): void {
```
MohamadNateqi, line 2252: Type hint for `$email`.

includes/Main.php
```diff
@@ -1992,23 +2219,24 @@ public function add_document_link_to_address_section( string $type, \WC_Abstract
 	 * @param bool $sent_to_admin
 	 * @param bool $plain_text
 	 * @param mixed|\WC_Email $email Some third-parties might send different values.
```
MohamadNateqi, line 2249: Do you think it's better to remove the `?\WC_Email` type hint from `attach_document_to_email()` for the same reason explained here?

MohamadNateqi (changes requested):

edi/Peppol.php
alexmigf: The REST route and the account endpoint use different hooks, so the order here should not affect when they run.

includes/Documents/OrderDocumentMethods.php
alexmigf: I’d prefer to leave this unchanged for now.

includes/Main.php
alexmigf: I’d rather not, just to be on the safe side.

includes/Main.php
alexmigf: Yes, done.

MohamadNateqi (approved):
Amazing job!

alexmigf:
@MohamadNateqi Really appreciated

includes/Settings.php
```diff
@@ -580,19 +661,35 @@ public function get_document_settings( $document_type, $output_format = 'pdf' )
...
 
-	public function get_output_format( $document = null, $request = null ) {
+	/**
+	 * Get output format for document generation.
+	 *
+	 * @param OrderDocument|null $document
+	 * @param array|null $request
+	 * @return string
+	 */
+	public function get_output_format( ?OrderDocument $document = null, ?array $request = null ): string {
```
YordanSoares: I'm getting this error while using the bulk actions to create either PDF invoices or packing slips:
```
Fatal error: WPO\IPS\Settings::get_output_format(): Argument #1 ($document) must be of type ?WPO\IPS\Documents\OrderDocument, WPO\IPS\Documents\BulkDocument given, called in /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Main.php on line 572
/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Settings.php (671)
#0 /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Main.php(572): WPO\IPS\Settings->get_output_format()
#1 /wp-includes/class-wp-hook.php(341): WPO\IPS\Main->generate_document_ajax()
#2 /wp-includes/class-wp-hook.php(365): WP_Hook->apply_filters()
#3 /wp-includes/plugin.php(522): WP_Hook->do_action()
#4 /wp-admin/admin-ajax.php(192): do_action()
#5 {main}
```

YordanSoares (changes requested):

includes/Settings.php
alexmigf: Fixed in the last commit.

wpo-ips-functions.php
```diff
@@ -107,29 +122,31 @@ function wcpdf_get_document( string $document_type, $order, bool $init = false )
...
-	return wcpdf_get_document( 'packing-slip', $order, $init );
-}
-
-function wcpdf_get_bulk_actions() {
+/**
+ * Get the available bulk actions for documents, which can be used in the admin order list or other places where bulk actions are needed.
+ *
+ * @return array
+ */
+function wcpdf_get_bulk_actions(): array {
```
YordanSoares, line 147: Another issue related with the bulk actions is that when you try to export several XML invoices, only the first one is exported and the rest of the selection is ignored.

alexmigf: Bulk actions work differently for XML files, they open as new tabs. Saying that, does the issue persist?

YordanSoares: Thanks for the clarification Alex! I remember now that you mentioned this before. The solution consists in enabling the pop-ups and redirects in the browser for the current URL/website (I tested this on Chrome):

image

includes/Settings.php
YordanSoares: I'm getting a new error:

```
Fatal error: wcpdf_get_pdf_maker(): Argument #3 ($document) must be of type ?WPO\IPS\Documents\OrderDocument, WPO\IPS\Documents\BulkDocument given, called in /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Documents/BulkDocument.php on line 94
/wp-content/plugins/woocommerce-pdf-invoices-packing-slips/wpo-ips-functions.php (185)
#0 /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Documents/BulkDocument.php(94): wcpdf_get_pdf_maker()
#1 /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Documents/BulkDocument.php(174): WPO\IPS\Documents\BulkDocument->get_pdf()
#2 /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Main.php(588): WPO\IPS\Documents\BulkDocument->output_pdf()
#3 /wp-includes/class-wp-hook.php(341): WPO\IPS\Main->generate_document_ajax()
#4 /wp-includes/class-wp-hook.php(365): WP_Hook->apply_filters()
#5 /wp-includes/plugin.php(522): WP_Hook->do_action()
#6 /wp-admin/admin-ajax.php(192): do_action()
#7 {main}
```

alexmigf: Fixed in the last commit.

YordanSoares (approved):
## Tests results - Free features:

| Feature tested | Result |
| ------- | ----------- |
| Setup Wizard | ✅ Passed | 
| Admin: Creation of PDF invoice [single document] | ✅ Passed | 
| Admin: Creation of PDF packing slip [single document] | ✅ Passed | 
| Admin: Creation of electronic invoice (XML) [single document] | ✅ Passed | 
| Admin: Creation of PDF invoices [bulk actions] | ✅ Passed | 
| Admin: Creation of PDF packing slips [bulk actions] | ✅ Passed | 
| Admin: Creation of electronic invoices (XML) [bulk actions] | ✅ Passed | 
| Customer (My Account): Download of PDF invoice [single document] | ✅ Passed | 
| Customer (My Account): Download of electronic invoice (XML) [single document] | ✅ Passed | 
| Customer (Guest): Download of PDF invoice [single document] | ✅ Passed | 
| Customer (Guest): Download of electronic invoice (XML) [single document] | ✅ Passed | 
| Automatic creation via "Attach to" setting + email attachment | ✅ Passed | 
| Layout of PDF invoice | ✅ Passed | 
| Layout of PDF packing slip | ✅ Passed | 
| Document Internationalization (default translations delivered by WP.org) | ✅ Passed | 
| Sequential numeration [default] | ✅ Passed | 
| Sequential numeration [custom/filtered] | ✅ Passed | 
| Number format (prefix, suffix, padding) + placeholders [default] | ✅ Passed | 
| Number format (prefix, suffix, padding) + placeholders [custom/filtered] | ✅ Passed | 
| Historical settings on a per-order basis | ✅ Passed |

YordanSoares (approved):
# Extensions compatibility tests
##  [Premium Templates](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/) + [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) and [Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/) documents

| Document Types ↓ / Templates →     | Simple Premium<sup>1</sup> | Business<sup>1</sup> | Modern<sup>1</sup> |
|---------------------------------|:----------------:|:---------:|:-------:|
| Invoice                         | ✅               | ✅        | ✅      |
| Packing Slip                    | ✅               | ✅        | ✅      |
| Credit Note<sup>2</sup>      | ✅               | ✅        | ✅      |
| Proforma Invoice<sup>2</sup> | ✅               | ✅        | ✅      |
| Receipt<sup>2</sup>          | ✅               | ✅        | ✅      |
| Delivery Note<sup>2</sup>          | ✅               | ✅        | ✅      |
| Proposal<sup>3</sup>                        | ✅               | ✅        | ✅      |
| Order Confirmation<sup>3</sup>              | ✅               | ✅        | ✅      |

<sup>1</sup> Included in the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/).
<sup>2</sup> Included in the [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) extension and the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/).
<sup>3</sup> Included in the [Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/) extension and the [Ultimate Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-ultimate-bundle/).

---

##  [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) features

| Feature tested | Result |
| ------- | ----------- |
| Admin: Creation of PDF credit note, proforma, receipt, delivery note [single document] | ✅ Passed |
| Admin: Creation of PDF credit note, proforma, receipt, delivery note [bulk actions] | ✅ Passed |
| Admin: Creation of PDF credit note, proforma, receipt, delivery note [bulk export tool] | ✅ Passed |
| Admin: Creation of electronic credit note (XML) [single document] | ✅ Passed |
| Admin: Creation of electronic credit notes (XML) [bulk actions] | ✅ Passed |
| Customer (My Account): Download of PDF credit note [single document] | ✅ Passed |
| Customer (My Account): Download of XML credit note [single document] | ✅ Passed |
| Automatic creation via "Attach to" setting + email attachment | ✅ Passed |
| Layout of PDF credit note, proforma, receipt, delivery note (Simple template) | ✅ Passed |
| Multilingual: WPML | ✅ Passed |
| Multilingual: Polylang | ✅ Passed |

Just a remark regarding the multilingual testing with Polylang: Although the document is correctly translated in the attachments, when an admin user opens the document it uses the language set in their profile, without respecting "Order/customer language" setting.

---

##  [mPDF library](https://github.com/wpovernight/woocommerce-pdf-ips-mpdf)

| Feature tested | Result |
| ------- | ----------- |
|  PDF document creation (all documents)  | ✅ Passed |
|  RTL support (all documents)  | ✅ Passed |

---

##  [Custom Font extension](https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/available-extensions/#custom-font-extension)

| Feature tested | Result |
| ------- | ----------- |
|  Custom font applied on generated PDF documents  | ✅ Passed |

---

##  [Shipping and Fees item](https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/available-extensions/#pdf-invoices-&-packing-slips-for-woocommerce-%E2%80%93-shipping-and-fees-item)

| Feature tested | Result |
| ------- | ----------- |
| Shipping and fees moved to the order table | ✅ Passed |

---

##  [Credit Notes for Cancelled Orders](https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/available-extensions/#cancelled-credit-notes-extension)

| Feature tested | Result |
| ------- | ----------- |
| Creation of PDF credit notes for cancelled orders | ✅ Passed |

---

##  [Quotation Document](https://docs.wpovernight.com/woocommerce-pdf-invoices-packing-slips/available-extensions/#pdf-invoices-&-packing-slips-for-woocommerce-%E2%80%93-quotation)

When I tested this extension alongside Professional 2.26.0, the following error appeared immediately after I activated it:

```
Fatal error:  Declaration of WPO_WCPDF_Quotation_Document::init_settings() must be compatible with WPO\IPS\Documents\OrderDocument::init_settings(): void in /wp-content/plugins/wcpdf-quotation/class-wcpdf-quotation-document.php on line 70

Stack trace:
#0 /wp-content/plugins/wcpdf-quotation/wcpdf-quotation.php(33): include()
#1 /wp-includes/class-wp-hook.php(341): WooCommerce_PDF_IPS_Quotation->register_documents()
#2 /wp-includes/plugin.php(205): WP_Hook->apply_filters()
#3 /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Documents.php(53): apply_filters()
#4 /wp-content/plugins/woocommerce-pdf-invoices-packing-slips/includes/Documents.php(67): WPO\IPS\Documents->init()
#5 /wp-content/plugins/woocommerce-pdf-ips-pro/includes/wcpdf-pro-functions.php(1093): WPO\IPS\Documents->get_documents()
#6 /wp-includes/class-wp-hook.php(341): WPO\WC\PDF_Invoices_Pro\Functions->filter_document_titles()
#7 /wp-includes/class-wp-hook.php(365): WP_Hook->apply_filters()
#8 /wp-includes/plugin.php(522): WP_Hook->do_action()
#9 /wp-settings.php(774): do_action()
#10 /wp-config.php(50): require_once('...')
#11 /wp-load.php(50): require_once('...')
#12 /wp-admin/admin.php(35): require_once('...')
#13 /wp-admin/plugins.php(10): require_once('...')
#14 {main}
```

That said, as this probably need to be fixed in the Quotation extension directly, I'll mark this PR as approved for now.

alexmigf:
@YordanSoares I’ve applied some changes to prevent the fatal error. Since the Quotation extension will be deprecated, we should direct users to the new extension available here, which is compatible with v6 only:

https://github.com/wpovernight/wpo-ips-custom-document

YordanSoares (commented):
@alexmigf the plugin is correctly deactivated when upgrading:

Image

However, the Quotation plugin can be reactivated by accident after updating IPS, producing the fatal error again. Can you also block the extension from activating at a later stage?

alexmigf:
@YordanSoares Could you try again with the last commit?

YordanSoares (approved):
@alexmigf I think we've got all the bases covered now! After your last comment, I realized that the Quotation extension is not only disabled after updating to v6, but that the "Activate" button is also completely disabled to prevent it from being accidentally enabled later on.

Image

YordanSoares (approved):
## Test results:

I tested all of our default document types, including the ones from [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) and [Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/), using both the **Simple** template and the premium templates included in the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/). 

I generated the documents using the following versions:

* `v6.0.0-i1482.4` ([beta release](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/releases/tag/v6.0.0-i1482.4)) → dompdf `3.1.5`
* `1482-cachegrind-performance-review` (this PR branch) → dompdf `3.1.6`

All documents were generated identically, with no observable differences.

| Document Types ↓ / Templates →     | Simple | Simple Premium<sup>1</sup> | Business<sup>1</sup> | Modern<sup>1</sup> |
|---------------------------------|:------:|:----------------:|:---------:|:-------:|
| Invoice                         | ✅     | ✅               | ✅        | ✅      |
| Packing Slip                    | ✅     | ✅               | ✅        | ✅      |
| Credit Note<sup>2</sup>      | ✅     | ✅               | ✅        | ✅      |
| Proforma Invoice<sup>2</sup> | ✅     | ✅               | ✅        | ✅      |
| Receipt<sup>2</sup>          | ✅     | ✅               | ✅        | ✅      |
| Delivery Note<sup>2</sup> NEW         | ✅     | ✅               | ✅        | ✅      |
| Proposal<sup>3</sup>                        | ✅     | ✅               | ✅        | ✅      |
| Order Confirmation<sup>3</sup>              | ✅     | ✅               | ✅        | ✅      |

<sup>1</sup> Included in the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/).
<sup>2</sup> Included in the [Professional](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-professional/) extension and the [Plus Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-bundle/).
<sup>3</sup> Included in the [Order Proposal](https://wpovernight.com/downloads/woocommerce-order-proposal/) extension and the [Ultimate Bundle](https://wpovernight.com/downloads/woocommerce-pdf-invoices-packing-slips-ultimate-bundle/).

## Resources:
- Generated documents:  [pdf-tests-v6-dompdf-3.1.5-3.1.6.zip](https://github.com/user-attachments/files/30548403/pdf-tests-v6-dompdf-3.1.5-3.1.6.zip)
- Checker tool: https://github.com/vslavik/diff-pdf

MohamadNateqi:
@alexmigf Can you please merge main into this?

alexmigf:
@MohamadNateqi Not yet. As I mentioned in the meeting, it should be merged next week.

PR #1485, opened 2026-04-08
Fix: Fatal error during Peppol checkout field initialization

### LINKED ISSUE #1484: Peppol class is instantiated too early and triggers fatal error on checkout field registration
<p class="editor-paragraph-block">A fatal error occurs because the Peppol class is being instantiated before the required WooCommerce checkout field function is available.</p>

alexmigf:
closes #1484

### DIFF
--- a/edi/Peppol.php
+++ b/edi/Peppol.php
@@ -219,13 +219,6 @@ class Peppol {
 		if ( ! wpo_ips_edi_peppol_enabled_for_location( 'checkout' ) ) {
 			return;
 		}
-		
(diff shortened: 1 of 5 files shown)


MohamadNateqi:
@alexmigf How can I reproduce the fatal error? I'm not having it.

alexmigf:
This issue comes from [this report](https://wordpress.org/support/topic/undefined-function-wpoipsediwoocommerce_register_additional_checkout_field/). I couldn’t reproduce the error, but I’m assuming it happens because we call the function in our class constructor, which may run before that function is available.

MohamadNateqi (approved):
LGTM!

PR #1495, opened 2026-04-22
Refactore and improvement in Checkout Field

### LINKED ISSUE #1486: Option to print the "Checkout field" values on the invoice
<p class="editor-paragraph-block">The "Checkout field" feature in the free plugin currently saves the value received at checkout to the user data. This value isn't printed on the documents and there is no way to do this. The value is only printed on the back end order details page. It would be good to add the option to print this on the invoice in the same way we have the option to print the Shop VAT/COC numbers.</p>

### LINKED ISSUE #1550: Make Checkout Field values editable from the order page
[Plane work item: WPOVE-516](https://work.teamupdraft.com/team-updraft/browse/WPOVE-516)

Allow Checkout Field value to be edited directly from the order page, preferably below the addresses, similar to how VAT number plugins handle this.

MohamadNateqi:
close #1486
close #1550 

Changes in this PR:

- Update the labels of the two general settings related to the checkout field to enhance clarity in search results.
- Move 4 methods related to the checkout field to the general functions file to improve reusability.
- Add a new setting to display checkout field value on the invoice document.
- Make Checkout Field values editable from the order page

Settings page:
image

Invoice document:
image


wpo-ips-functions.php
```diff
@@ -1248,6 +1238,67 @@ function wpo_wcpdf_get_order_customer_vat_number( \WC_Abstract_Order $order ): ?
 	return apply_filters( 'wpo_wcpdf_order_customer_vat_number', $vat_number, $order, $meta_key ?? null );
 }
 
+/**
```
alexmigf: To be honest, I’m not a big fan of this change. It doesn’t really feel like it belongs here. Wouldn’t static methods make more sense in this case?

includes/Documents/OrderDocument.php
```diff
@@ -2214,6 +2235,35 @@ public function show_due_date(): bool {
 		return $this->get_due_date() > 0;
 	}
 
+	/**
+	 * Return the value of the checkout field.
+	 *
+	 * @return string
+	 */
+	public function get_checkout_field(): string {
```
alexmigf: I think `get_checkout_field_value()` would be a more appropriate name here.

includes/Documents/OrderDocument.php
```diff
@@ -2214,6 +2235,35 @@ public function show_due_date(): bool {
...
+
+		return apply_filters( 'wpo_wcpdf_checkout_field_value', $checkout_field_value, $this );
+	}
+
+	/**
+	 * Print the checkout field value.
+	 *
+	 * @return void
+	 */
+	public function checkout_field(): void {
```
alexmigf: Same here: `checkout_field_value()`

alexmigf (changes requested):

wpo-ips-functions.php
MohamadNateqi: Keeping them in `Frontend` would mean those classes need to reach into a `Frontend` class just to read settings, which seemed a misleading coupling to me.

Also, they're for reading settings, not a front-end behavior, so I thought putting them in the functions file, along with the rest of similar functions, might be better.

What do you think?

alexmigf: I understand your point of view, and it makes sense to me as well. Maybe we could create a separate class to handle everything related to the Checkout Field. What do you think?

Since this is a group of functions related to the same feature, rather than a single isolated function, it might be cleaner to keep them together in their own class.

MohamadNateqi: It seems a good idea, I'll implement that and see how it goes.

alexmigf:
@MohamadNateqi I’ll have to postpone this to a future version (after 5.10.0), as there isn’t enough time to properly test it now and it would introduce significant changes to the code structure.

alexmigf:
@MohamadNateqi It would be better if you could migrate the base branch to [this one](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1495). Sorry for the trouble. If you think creating a new branch on your side is cleaner, that’s also fine with me.

MohamadNateqi:
@alexmigf This is ready for review.

alexmigf:
@MohamadNateqi Could you please resolve the conflicts when you have a chance?

MohamadNateqi:
@alexmigf Done.

PR #1497, opened 2026-04-28
Fix: Escaping context in PDF templates

### LINKED ISSUE #1496: Fix incorrect escaping in PDF templates
<p class="editor-paragraph-block">Some PDF template values are escaped with the wrong function for their output context.</p>

alexmigf:
closes #1496

### DIFF
--- a/templates/Simple/invoice.php
+++ b/templates/Simple/invoice.php
@@ -118,16 +118,16 @@
 	</thead>
 	<tbody>
 		<?php foreach ( $this->get_order_items() as $item_id => $item ) : ?>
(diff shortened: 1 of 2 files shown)


templates/Simple/invoice.php
MohamadNateqi, line 141: Should this one be `wp_kses_post()`? Because the value of this comes from the `get_formatted_line_subtotal()` which return a formatted value.

alexmigf: Maybe not? We don’t really know what side effects could result from injecting their classes, do we?

MohamadNateqi (approved):
LGTM!

PR #1501, opened 2026-05-04
New: My Account display helper to order documents

### LINKED ISSUE #1500: Add My Account display helper to Documents
<p class="editor-paragraph-block">Add an <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">is_my_account_allowed()</code> method to the abstract <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">OrderDocument</code> class to centralize the logic that determines whether a document can be displayed on the My Account page.</p>

alexmigf:
closes #1500

### DIFF
--- a/includes/Documents/OrderDocument.php
+++ b/includes/Documents/OrderDocument.php
@@ -582,6 +582,37 @@ abstract class OrderDocument {
 		return apply_filters( 'wpo_wcpdf_document_is_allowed', $allowed, $this );
 	}
 
+	/**
+	 * Check if document is allowed to be displayed on the My Account page.
+	 *
+	 * @param string $default
+	 * @param string $output_format
+	 * @return bool
+	 */
+	public function is_my_account_allowed( string $default = '', string $output_format = 'pdf' ): bool {
+		if ( ! $this->is_enabled() ) {
+			return false;
+		}
+
+		$button_setting = $this->get_setting( 'my_account_buttons', $default, $output_format );
+		$allowed        = false;
+
+		switch ( $button_setting ) {
+			case 'available':
+				$allowed = $this->exists();
+				break;
+			case 'always':
+				$allowed = true;
+				break;
+			case 'custom':
+				$allowed_statuses = $this->get_setting( 'my_account_restrict', array(), $output_format );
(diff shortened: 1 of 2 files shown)


includes/Documents/OrderDocument.php
```diff
@@ -582,6 +582,37 @@ public function is_allowed() {
...
 	}
 
+	/**
+	 * Check if document is allowed to be displayed on the My Account page.
+	 *
+	 * @param string $default
+	 * @param string $output_format
+	 * @return bool
+	 */
+	public function is_my_account_allowed( string $default = '', string $output_format = 'pdf' ): bool {
```
MohamadNateqi: Can we have a better name for this function? Maybe like "can_show_in_my_account"

includes/Documents/OrderDocument.php
```diff
@@ -582,6 +582,37 @@ public function is_allowed() {
...
+		if ( ! $this->is_enabled() ) {
+			return false;
```
MohamadNateqi: Should we apply the filter here as well?

includes/Documents/OrderDocument.php
```diff
@@ -582,6 +582,37 @@ public function is_allowed() {
...
+		switch ( $button_setting ) {
+			case 'available':
+				$allowed = $this->exists();
+				break;
+			case 'always':
+				$allowed = true;
+				break;
+			case 'custom':
+				$allowed_statuses = $this->get_setting( 'my_account_restrict', array(), $output_format );
+				$allowed          = ! empty( $allowed_statuses ) && $this->order && in_array( $this->order->get_status(), array_keys( $allowed_statuses ), true );
```
MohamadNateqi: Can we have it in multiple line?

includes/Documents/OrderDocument.php
alexmigf: This is not specifically about the action button itself, so I’m not sure that name is the best fit. Maybe `is_available_in_my_account()` would be clearer?

MohamadNateqi: That, or `is_allowed_in_my_account()`, would still feel better.

MohamadNateqi (approved):
LGTM!

PR #1505, opened 2026-05-05
Fix: EDI delivery date serialization

### LINKED ISSUE #1504: EDI XML generation fails when using the order completed date as delivery date
<p class="editor-paragraph-block">EDI XML generation can fail when the delivery date is included and the order has a completed date. This prevents the XML document from being generated successfully.</p>

alexmigf:
closes #1504


alexmigf:
Customer approved!

PR #1514, opened 2026-05-12
Previewer search popup panels aren't expired

### LINKED ISSUE #1513: Previewer search popup panels aren't expired
<p class="editor-paragraph-block">Following <a target="_blank" rel="noopener noreferrer nofollow" class="text-custom-primary-300 underline underline-offset-[3px] hover:text-custom-primary-500 transition-colors cursor-pointer" href="https://wpovernight.slack.com/archives/C0121KYKGN6/p1769095425888589">the thread here</a>, when searching for an order in the previewer, the popup that notifies when an order is not found doesn't expire on it's own and can't be closed manually. It also blocks the input if a user wants to edit the search</p>

wpo-nnawaz:
Fixes #1513


alexmigf (approved):
LGTM!

PR #1517, opened 2026-05-18
Fix: EDI price rounding for discounted multi-quantity lines

### LINKED ISSUE #1516: Item price rounding can generate invalid XML for discounted multi-quantity lines
<p class="editor-paragraph-block">When a coupon discount is distributed across an order line with quantity greater than 1, the generated XML can export a 2-decimal unit price that no longer matches the line total.</p>

alexmigf:
closes #1516


alexmigf:
Customer document sent successfully with this fix.

PR #1521, opened 2026-05-20
Add email attachment success/failure logging to order notes

### LINKED ISSUE #1520: Email attachment log
<p class="editor-paragraph-block">Following on some customers that have attachment issues with the documents, we have documentation that ask them use WP Mail Logging to track invoice attachments. However, this only shows the log for future emails. Since we have an option to log the creation of documents, we can also add an option to enable email attachment success or failure.</p>

wpo-nnawaz:
When the "Log to order notes" debug setting is enabled, attachment success and failure are now logged to order notes for each email.

Closes #1520


BrunoPavlinic98 (approved):
LGTM!

includes/Main.php
```diff
@@ -181,7 +181,9 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
 
 						if ( $attachment ) {
 							$attachments[] = $attachment;
+							$this->log_email_attachment_to_order_notes( $document, $email_id, true );
 						} else {
+							$this->log_email_attachment_to_order_notes( $document, $email_id, false );
```
alexmigf: Maybe this can be removed here and replaced with `wcpdf_log_error()` instead.

includes/Main.php
```diff
@@ -1487,6 +1489,39 @@ public function log_unmark_document_printed_to_order_notes( $document ) {
 		}
 	}
 
+	/**
```
alexmigf: Maybe this function is not needed just to pass a single string. Calling `log_to_order_notes()` directly above should be enough.

alexmigf (changes requested):

alexmigf (commented):
This is generating two notes, even though I only received one email.

Image

wpo-nnawaz:
I couldn't reproduce this locally. Could you share what setup triggered the duplicate? Is the EDI extension active on your end?

alexmigf:
@wpo-nnawaz Yes, I forgot to check the attachments. My apologies.

Screenshot_20260526_153950

In this case the note should be:

- PDF Invoice ...
- XML Invoice ...

includes/Main.php
```diff
@@ -181,7 +181,17 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
...
+								$note = sprintf(
+									/* translators: 1. document title, 2. email ID */
+									__( 'PDF %1$s successfully attached to email (%2$s).', 'woocommerce-pdf-invoices-packing-slips' ),
+									$document->get_title(),
+									$email_id
+								);
+								$this->log_to_order_notes( $note, $document );
+							}
 						} else {
+							wcpdf_log_error( sprintf( 'PDF %1$s could not be attached to email (%2$s).', $document->get_title(), $email_id ), 'critical' );
```
MohamadNateqi, line 202: I think it would be good to report the failure in the order note as well, not only in the logs.

@alexmigf What do you think?

includes/Main.php
```diff
@@ -181,7 +181,17 @@ public function attach_document_to_email( $attachments, $email_id, $order, $emai
 
 						if ( $attachment ) {
 							$attachments[] = $attachment;
+							if ( ! empty( WPO_WCPDF()->settings->debug_settings['log_to_order_notes'] ) ) {
+								$note = sprintf(
+									/* translators: 1. document title, 2. email ID */
+									__( 'PDF %1$s successfully attached to email (%2$s).', 'woocommerce-pdf-invoices-packing-slips' ),
+									$document->get_title(),
+									$email_id
```
MohamadNateqi: It might be better to print the email title instead of the ID?

MohamadNateqi (changes requested):
As Alex also mentioned, the name should show the difference based on the format:

Image

includes/Main.php
alexmigf: To be honest, I don’t think this is the right place for that.

MohamadNateqi: But this way, as a user, I might refresh the order, and if I don't see any new note related to that action, I may get confused about why it's missing and assume that the absence of an order note means the attachment failed.

alexmigf: I was only replying to your question. Personally, I think that if an issue occurs, it should be investigated using the proper logs instead of cluttering the order notes. That said, I’m not against it, it’s just my opinion.

MohamadNateqi (approved):
LGTM!

PR #1525, opened 2026-05-22
Fix: Fatal error when locale value is missing

### LINKED ISSUE #1524: Fix fatal error caused by missing locale value
<p class="editor-paragraph-block">A fatal error can occur when the settings getter receives a missing locale value.</p><p class="editor-paragraph-block">The method currently requires the locale to be a string, but in some cases <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">null</code> is passed, causing a <code class="rounded bg-custom-background-80 px-[6px] py-[1.5px] font-mono font-medium text-orange-500 border-[0.5px] border-custom-border-200" spellcheck="false">TypeError</code>.</p>

alexmigf:
closes #1524

### DIFF
--- a/includes/Settings/SettingsGeneral.php
+++ b/includes/Settings/SettingsGeneral.php
@@ -881,11 +881,11 @@ class SettingsGeneral {
 	/**
 	 * Get a general setting key value, optionally using a locale-specific sub-key.
 	 *
-	 * @param string $key     The key of the setting to retrieve.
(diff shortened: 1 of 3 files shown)


alexmigf:
Customer approved!

PR #1533, opened 2026-05-28
Bulk Document: “Mark as printed” not applied to all orders in bulk export

### LINKED ISSUE #1528: Bulk Document: “Mark as printed” not applied to all orders in bulk export
[Plane work item: WPOVE-463](https://work.teamupdraft.com/team-updraft/browse/WPOVE-463)

The current implementation of the Bulk Document is not working as expected with the "Marked as printed" option because it does not properly loop through all orders/documents included in the bulk export.

 suggested to follow a similar approach used in [this snippet](https://wpovernight.slack.com/archives/C0121KYKGN6/p1779463659448259?thread_ts=1779367208.954679&cid=C0121KYKGN6), where the bulk document process iterates through all related order IDs and marks each generated document as printed individually.

BrunoPavlinic98:
closes #1528


includes/Main.php
```diff
@@ -534,6 +534,15 @@ public function generate_document_ajax(): void {
 			if ( $document ) {
 				do_action( 'wpo_wcpdf_document_created_manually', $document, $order_ids ); // note that $order_ids is filtered and may not be the same as the order IDs used for the document (which can be fetched from the document object itself with $document->order_ids)
 
+				if ( count( $order_ids ) > 1 && isset( $request['bulk'] ) && property_exists( $document, 'order_ids' ) && ! empty( $document->order_ids ) ) {
```
alexmigf: I think this should replace the code on line `515`.

alexmigf (changes requested):

alexmigf (approved):
LGTM!

PR #1536, opened 2026-06-03
Fix document number display setting not respected in preview

### LINKED ISSUE #1535: Fix preview document number handling to respect number display setting
[Plane work item: WPOVE-476](https://work.teamupdraft.com/team-updraft/browse/WPOVE-476)

The document preview currently does not fully respect the document number display setting.

wpo-nnawaz:
## What
The document number display setting (display_number) was not being respected in the document preview for any document type that has this setting, including Invoice and Packing Slip.

## Root cause
Two separate issues:

1. In `ajax_preview()`, when the document does not exist yet, a sequential number was unconditionally assigned via `get_next()` before checking the `display_number` setting. The second block that calls `get_document_number()` (which correctly handles `display_number = order_number`) only ran if the number was still empty, so it was always skipped.

2. For documents with historical settings enabled (Invoice), `$document->settings` was loaded with order meta settings from a previously generated document, overriding the current form data in preview.

## Changes
- `Settings.php`: Replaced the unconditional `get_next()` call in `ajax_preview()` with `get_document_number()`, which already handles all `display_number` options internally. Also added `$document->settings = $document->get_settings( true )` to ensure preview always reflects current settings, bypassing historical order meta.
- `OrderDocument.php`: Updated `get_number_title()` to return "Order Number:" when `display_number = order_number`.
- `Invoice.php`: Same fix in the overriding `get_number_title()`.
- `PackingSlip.php`: Same fix in the overriding `get_number_title()`.

## Notes
The packing slip number/date display is handled by the Pro extension via the `wpo_wcpdf_before_order_data` hook in `packing_slip_number_date()`. That method has its own preview block with the same unconditional `get_next()` pattern, so the packing slip number value fix will need a corresponding change in Pro.

Closes #1535

### DIFF
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -303,24 +303,19 @@ class Settings {
 				$document = wcpdf_get_document( $document_type, $order );
 
 				if ( $document ) {
+					// Always use latest settings in preview, historical settings from the order should not affect preview.
+    				$document->settings = $document->get_settings( true );
+
 					if ( ! $document->exists() ) {
 						$document->set_date( current_time( 'timestamp', true ) );
-						$number_store_method = $this->get_sequential_number_store_method();
-						$number_store_name   = apply_filters( 'wpo_wcpdf_document_sequential_number_store', "{$document->slug}_number", $document );
-						$number_store        = new SequentialNumberStore( $number_store_name, $number_store_method );
-						$document->set_number( $number_store->get_next() );
-					}
-
-					// Update document date.
-					$document->initiate_date();
-
-					// Update document number only if it is not already set.
-					if ( empty( $document->get_number() ) ) {
 						$document_number = $document->get_document_number();
 						if ( ! empty( $document_number ) ) {
 							$document->set_number( $document_number );
 						}
 					}
+
+					// Update document date.
+					$document->initiate_date();
 
 					$document_number = $document->get_number( $document->get_type() );
 
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -334,7 +334,7 @@ class Settings {
 						) {
 							$order_number = $order->get_order_number();
 							$number_data  = array(
-								'number'           => absint( $order_number ),
+								'number'           => (int) preg_replace( '/\D/', '', $order_number ),
 								'formatted_number' => "{$order_number}",
 							);
 						}
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -303,26 +303,54 @@ class Settings {
 				$document = wcpdf_get_document( $document_type, $order );
 
 				if ( $document ) {
-					// Always use latest settings in preview, historical settings from the order should not affect preview.
-    				$document->settings = $document->get_settings( true );
-
 					if ( ! $document->exists() ) {
 						$document->set_date( current_time( 'timestamp', true ) );
+						$number_store_method = $this->get_sequential_number_store_method();
+						$number_store_name   = apply_filters( 'wpo_wcpdf_document_sequential_number_store', "{$document->slug}_number", $document );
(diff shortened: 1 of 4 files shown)


wpo-nnawaz:
One thing worth noting for Pro: the packing slip number and date are rendered via `packing_slip_number_date()` in the Pro Functions class, which has its own preview block that follows the same old pattern (unconditional `get_next()` before checking `display_number`). The label fix works because `get_number_title()` is fixed in PackingSlip.php, but the value will still show the packing slip number instead of the order number in preview. That needs the same fix applied to the preview block inside `packing_slip_number_date()` in Pro.

alexmigf (changes requested):
I don’t think this is the fix we need. The issue seems to be in the `ajax_preview()` method, where we are not taking the number setting type into account.

wpo-nnawaz:
We did make changes to ajax_preview() to use get_document_number() instead of unconditionally calling get_next(). Is that not the right approach, or did you have a more minimal fix in mind?

alexmigf (commented):
You’re changing code that isn’t related to the actual problem. The document classes shouldn’t be touched here, since the issue is only in `ajax_preview()`.

Please try this:

```php
// Apply document number formatting.
if ( $document_number ) {
	$number_data = array();

	if (
		isset( $document->settings['display_number'] ) &&
		'order_number' === $document->settings['display_number']
	) {
		$order_number = $order->get_order_number();
		$number_data  = array(
			'number'           => absint( $order_number ),
			'number_formatted' => "{$order_number}",
		);
	}

	if (
		! empty( $document->settings['number_format'] ) &&
		is_array( $document->settings['number_format'] )
	) {
		$number_data = array_merge(
			$number_data,
			$document->settings['number_format']
		);
	}

	if ( ! empty( $number_data ) ) {
		$document_number->load_number_data( $number_data );
	}

	$document_number->apply_formatting( $document, $order );
}
```

wpo-nnawaz:
With the updated approach, the same fix needs to be applied to the formatting block inside `packing_slip_number_date()` in Pro.

alexmigf:
@wpo-nnawaz Correct!

By the way, I made a mistake in my suggestion, should be: `'formatted_number' => "{$order_number}",`

wpo-nnawaz:
Thanks for the correction, fixed!

alexmigf:
I think we don’t need to change any other files besides the `ajax_preview()` method.

wpo-nnawaz:
Got it, makes sense. I was trying to match the label with the selected value but now I understand what the setting is actually doing. Reverted those changes.

includes/Settings.php
```diff
@@ -326,8 +326,31 @@ public function ajax_preview() {
...
-							$document_number->load_number_data( $document->settings['number_format'] );
+						$number_data = array();
+
+						if (
+							isset( $document->settings['display_number'] ) &&
+							'order_number' === $document->settings['display_number']
+						) {
+							$order_number = $order->get_order_number();
+							$number_data  = array(
+								'number'           => absint( $order_number ),
```
alexmigf: Maybe we need to be careful here, as this value may contain non-numeric characters.

wpo-nnawaz: Good point. If the order number contains non-numeric characters, `absint()` would return `0` and the formatted number would be wrong. Should we use `(int) preg_replace( '/\D/', '', $order_number )` for the `number` key instead, keeping the full original in `formatted_number`? That's consistent with what `get_document_number()` already does for non-numeric values.

alexmigf: Yes, let's go with that.

alexmigf (approved):
LGTM!

PR #1538, opened 2026-06-04
Add partial invoice number search setting

### LINKED ISSUE #1537: Invoice number search comparison enhancement
[Plane work item: WPOVE-479](https://work.teamupdraft.com/team-updraft/browse/WPOVE-479)

Related to the [intake here](https://work.teamupdraft.com/team-updraft/browse/WPOVE-455/),  found a minor issue where when the "Invoice number" filter is selected on the Orders page, the search tries to match the exact input and not use the IN comparison.

wpo-nnawaz:
## What
Adds a new "Enable partial invoice number matching" setting in the invoice document settings, directly below the existing invoice number search setting.

## Why
When using the "Invoice numbers" HPOS search filter, the search only matched exact values. This adds opt-in partial/substring matching (e.g. searching "123" now matches "INV-123").

## How
- New `invoice_number_partial_search` checkbox setting in `Invoice.php`, registered under the `admin_display` settings category
- New `invoice_number_partial_search_enabled()` helper in `Admin.php` following the same pattern as `invoice_number_search_enabled()`
- `invoice_number_query_args()` in `ThirdPartyPlugins.php` now switches between `LIKE` (one clause per term, `OR` relation) and `IN` (exact) based on the setting
- JS in `admin-script.js` disables the partial search checkbox when invoice number search is off, following the existing `due_date`/`due_date_days` pattern

Disabled by default to preserve existing behavior.

Closes #1537


includes/Admin.php
```diff
@@ -449,6 +449,17 @@ public function invoice_number_search_enabled() {
 		return $is_enabled;
 	}
 
+	/**
+	 * Check if the invoice number partial search is enabled.
+	 *
+	 * @return boolean
+	 */
+	public function invoice_number_partial_search_enabled(): bool {
+		$invoice_settings = get_option( 'wpo_wcpdf_documents_settings_invoice', array() );
```
alexmigf, line 446: Maybe you could use `invoice_number_search_enabled()` passing a new parameter instead?

includes/Compatibility/ThirdPartyPlugins.php
```diff
@@ -435,12 +435,27 @@ public function invoice_number_query_args( array $order_query_args ): array {
...
-			$order_query_args['meta_query'][]  = [
-				'key'     => '_wcpdf_invoice_number',
-				'value'   => $invoice_numbers,
-				'compare' => 'IN',
-			];
+			$order_query_args['meta_query'] = $order_query_args['meta_query'] ?? array();
+			$partial_search_enabled         = WPO_WCPDF()->admin->invoice_number_partial_search_enabled();
+
+			if ( $partial_search_enabled ) {
+				$order_query_args['meta_query']['relation'] = 'OR';
```
alexmigf: Maybe we should avoid setting `relation => OR` directly on the root `meta_query`, as this could change the logic of any existing meta conditions.

alexmigf (changes requested):

assets/js/admin-script.js
```diff
@@ -116,6 +116,17 @@ jQuery( function( $ ) {
 		}
 	} ).trigger( 'change' );
 
+	$( '#wpo-wcpdf-preview-wrapper #invoice_number_search' ).on( 'change', function() {
```
alexmigf: You could avoid this by using `custom_attributes`, for example:

```php
array(
	'type'     => 'setting',
	'id'       => 'invoice_number_partial_search',
	'title'    => __( 'Enable partial invoice number matching', 'woocommerce-pdf-invoices-packing-slips' ),
	'callback' => 'checkbox',
	'section'  => $this->type,
	'args'     => array(
		'option_name'       => $option_name,
		'id'                => 'invoice_number_partial_search',
		'description'       => __( 'When enabled, searching for an invoice number will also match partial values (e.g. searching "123" will match "INV-123"). Requires invoice number search to be enabled.', 'woocommerce-pdf-invoices-packing-slips' ),
		'custom_attributes' => array(
			'data-show_for_option_name'   => $option_name . '[invoice_number_search]',
			'data-show_for_option_values' => wp_json_encode( array( '1' ) ),
			'data-keep_current_value'     => true,
		),
	)
),
```

alexmigf (changes requested):

alexmigf (approved):
LGTM!

PR #1540, opened 2026-06-04
Fix: Empty order item prices when WooCommerce subtotal is filtered

### LINKED ISSUE #1539: Order item prices can be empty when WooCommerce formatted subtotal is filtered
[Plane work item: WPOVE-481](https://work.teamupdraft.com/team-updraft/browse/WPOVE-481)

PDF item prices can be empty when the value returned by WooCommerce’s `get_formatted_line_subtotal()` is filtered to an empty string.

alexmigf:
closes #1539


MohamadNateqi (approved):
LGTM!

PR #1542, opened 2026-06-05
Fix: Prevent fatal when REST invoice number lookup returns bulk document

### LINKED ISSUE #1541: Fatal error when adding invoice number to WooCommerce REST API order response
[Plane work item: WPOVE-482](https://work.teamupdraft.com/team-updraft/browse/WPOVE-482)

A customer reported that the WooCommerce orders REST API can fail when our plugin adds the invoice number to the order response.

The bug happens when the document lookup returns a bulk document instead of a regular invoice document. The REST API logic expects a single invoice document and tries to read its invoice number, but a bulk document does not have an invoice number of its own.

As a result, the `/wp-json/wc/v3/orders` endpoint can trigger a fatal error instead of returning the order data.

alexmigf:
closes #1541

### DIFF
--- a/includes/Frontend.php
+++ b/includes/Frontend.php
@@ -183,22 +183,31 @@ class Frontend {
 	 * Retrieve formatted invoice number for a given order
 	 *
 	 * @param \WC_Abstract_Order|\WC_Order $order
-	 *
 	 * @return string
 	 */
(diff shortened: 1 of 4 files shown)


MohamadNateqi (approved):
LGTM!

PR #1543, opened 2026-06-09
1401 add global template color setting

### LINKED ISSUE #1401: Modern template color setting displayed in wrong location

wpo-nnawaz:
Closes #1401

- Registers `template_color` setting in `SettingsGeneral.php` under Display Settings, after `template_ink_saving`
- Adds `apply_template_color_styles()` in `Main.php` hooked on `wpo_wcpdf_template_styles`, fires `wpo_ips_template_color_css` filter so templates can apply their own color logic
- Fixes color input detection in `admin-script.js` for save button and preview refresh
- Introduces `wpo_ips_template_color_supported_templates` and `wpo_ips_template_color_default` filters for template opt-in


includes/Main.php
```diff
@@ -76,6 +76,9 @@ public function __construct() {
 		// set ink saving mode
 		add_filter( 'wpo_wcpdf_template_custom_styles', array( $this, 'apply_ink_saving_styles' ), 10, 2 );
 
+		// apply template color
+		add_filter( 'wpo_wcpdf_template_styles', array( $this, 'apply_template_color_styles' ), 10, 2 );
```
alexmigf, line 80: Maybe this can use the same filter as the save ink setting?

wpo-nnawaz: `wpo_wcpdf_template_custom_styles` doesn't reliably contain the `style.css` content (it's whatever other custom-styles hooks happen to echo, which for ink saving doesn't matter since ink saving only appends). Our `str_replace( '#2F62AF', $color, $css )` needs the actual `style.css` content, which only `wpo_wcpdf_template_styles` provides.

alexmigf (approved):
LGTM!

PR #1546, opened 2026-06-17
Fix automatic “Mark as printed” options

### LINKED ISSUE #1534: Fix automatic “Mark as printed” options
[Plane work item: WPOVE-474](https://work.teamupdraft.com/team-updraft/browse/WPOVE-474)

Some of the automatic options in the **Mark as printed** setting are not working as expected.

The setting currently allows documents to be marked as printed manually or automatically based on the document creation context, but some automatic triggers are not marking the document as printed.

BrunoPavlinic98:
closes #1534

### DIFF
--- a/includes/Admin.php
+++ b/includes/Admin.php
@@ -1508,9 +1508,7 @@ class Admin {
 					WPO_WCPDF()->main->log_document_creation_to_order_notes( $invoice, 'document_data' );
 				}
 
-				if ( isset( $document_data['number'] ) || isset( $document_data['date'] ) ) {
-					WPO_WCPDF()->main->mark_document_printed( $invoice, 'document_data' );
-				}
+				WPO_WCPDF()->main->mark_document_printed( $invoice, 'document_data' );
 			}
 
 			// allow other documents to hook here and save their form data
--- a/includes/Admin.php
+++ b/includes/Admin.php
@@ -1506,6 +1506,9 @@ class Admin {
 
 				if ( $is_new ) {
 					WPO_WCPDF()->main->log_document_creation_to_order_notes( $invoice, 'document_data' );
+				}
+
+				if ( isset( $document_data['number'] ) || isset( $document_data['date'] ) ) {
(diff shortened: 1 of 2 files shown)


includes/Admin.php
```diff
@@ -1506,6 +1506,9 @@ public function save_invoice_number_date( $order_id, $order ) {
 
 				if ( $is_new ) {
 					WPO_WCPDF()->main->log_document_creation_to_order_notes( $invoice, 'document_data' );
+				}
+
+				if ( isset( $document_data['number'] ) || isset( $document_data['date'] ) ) {
```
alexmigf: Maybe the previous version was already fine? I’m not seeing what issue this change is fixing.

includes/Main.php
```diff
@@ -507,7 +508,7 @@ public function generate_document_ajax(): void {
 
 		// if we got here, we're safe to go!
 		try {
-			if ( count( $order_ids ) > 1 && isset( $request['bulk'] ) ) {
+			if ( isset( $request['bulk'] ) ) {
```
alexmigf: Could you please explain the reason behind this change? I’m not sure it’s required, so I’d like to better understand the thinking here.

alexmigf (changes requested):

includes/Admin.php
BrunoPavlinic98: This is true, I forgot to remove this after testing something, good catch.

includes/Main.php
BrunoPavlinic98: Also a testing leftover, good catch.

alexmigf: Now that I’m testing it, I think this was a good fix. When generating a single document from the bulk action, the document is not being marked as printed.

alexmigf (changes requested):

includes/Main.php
BrunoPavlinic98: Yes, that is why I was confused. The issue occurs only when printing one bulk document, and if 'On bulk order action' is the only value for mark as printed option.

I reverted the code back.

alexmigf (approved):
Good work!

PR #1560, opened 2026-07-02
Make cancelled/failed order email labels distinct between admin and customer

### LINKED ISSUE #1557: Doubled "Cancelled order" and "Failed order" email notifications
[Plane work item: WPOVE-524](https://work.teamupdraft.com/team-updraft/browse/WPOVE-524)

The "Cancelled order" and "Failed order" email notifications are doubled in the "Attach to" settings of the documents.

wpo-nnawaz:
Closes #1557

Cancelled order and Failed order emails have both admin and customer versions with the same title, so they showed up as duplicate looking checkboxes in the Attach to settings. Added a dynamic check instead of hardcoding these two cases, so any future email title collisions get labeled automatically too.

Only labels that actually collide get the "(Admin email)" / "(Customer email)" suffix now, using `is_customer_email()` to tell them apart.

### DIFF
--- a/includes/Documents/OrderDocument.php
+++ b/includes/Documents/OrderDocument.php
@@ -1849,11 +1849,11 @@ abstract class OrderDocument {
 	*/
 
 	/**
-	 * get all emails registered in WooCommerce
-	 * @param  boolean $remove_defaults switch to remove default woocommerce emails
-	 * @return array   $emails       list of all email ids/slugs and names
+	 * Get all emails registered in WooCommerce
+	 * 
+	 * @return array
 	 */
-	public function get_wc_emails() {
+	public function get_wc_emails(): array {
 		// only run this in the context of the settings page or setup wizard
 		// prevents WPML language mixups
 		$request = stripslashes_deep( $_GET ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
@@ -1863,7 +1863,7 @@ abstract class OrderDocument {
 		}
 
 		// get emails from WooCommerce
-		if (function_exists('WC')) {
+		if ( function_exists( 'WC' ) ) {
 			$mailer = WC()->mailer();
 		} else {
 			global $woocommerce;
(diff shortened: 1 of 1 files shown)


includes/Documents/OrderDocument.php
```diff
@@ -1881,6 +1881,18 @@ public function get_wc_emails() {
 			'customer_new_account'
 		);
 
+		$title_counts = array();
```
alexmigf: Maybe we don't need this? I think the emails have a different ID.

alexmigf (changes requested):

includes/Documents/OrderDocument.php
wpo-nnawaz: They do have different IDs, that's not the problem. The array's already keyed by $email->id so no ID collision there. The issue is two different IDs (cancelled_order and customer_cancelled_order) end up with the same title text ("Cancelled order"), so they show up as duplicate looking labels in the UI. $title_counts is checking for that title collision, not an ID one. Let me know if you want a different approach.

alexmigf: This should be enough I think:

```php
case 'cancelled_order':
case 'failed_order':
		$emails[$email->id] = sprintf('%s (%s)', $email->title, __( 'Admin email', 'woocommerce-pdf-invoices-packing-slips' ) );
		break;
```

wpo-nnawaz: Actually, I went with the dynamic approach instead, since it also covers any other titles that might collide down the line, not just these two. Should still handle the cancelled/failed order case fine. Let me know if you'd rather keep it simple with the hardcoded cases though.

alexmigf: I understand the reasoning, but I’m not sure this change is necessary. It adds an extra loop, and since this can include custom emails from third-party plugins, we may be assuming too much about those email objects and their available properties/methods.

wpo-nnawaz: Fair point, have switched to the hardcoded cases for cancelled_order and failed_order instead.

alexmigf (approved):
Since you’re at it, could you also fix the function code style?

wpo-nnawaz:
> Since you’re at it, could you also fix the function code style?

Done

alexmigf (approved):
Thank you!

PR #1562, opened 2026-07-02
Fix: EDI total reconciliation with emitted line amounts

### LINKED ISSUE #1561: EDI validation fails when document totals do not match emitted line totals
[Plane work item: WPOVE-538](https://work.teamupdraft.com/team-updraft/browse/WPOVE-538)

EDI validation can fail when generated document totals are calculated from different source values than the emitted line amounts.

alexmigf:
closes #1561


alexmigf:
Customer approved!

PR #1568, opened 2026-07-08
IPS (Free): Function print_emoji_styles is deprecated since version 6.4.0

### LINKED ISSUE #1565: IPS (Free): Function print_emoji_styles is deprecated since version 6.4.0
[Plane work item: WPOVE-551](https://work.teamupdraft.com/team-updraft/browse/WPOVE-551)

I'm seeing the following deprecation while using the Wizard in the free IPS plugin:

```
Deprecated:  Function print_emoji_styles is deprecated since version 6.4.0! Use wp_enqueue_emoji_styles instead. in /wp-includes/functions.php on line 6170
```

BrunoPavlinic98:
closes #1565

### DIFF
--- a/includes/SetupWizard.php
+++ b/includes/SetupWizard.php
@@ -217,6 +217,7 @@ class SetupWizard {
 			<?php wp_print_scripts( 'wpo-wcpdf-setup' ); ?>
 			<?php wp_print_scripts( 'wpo-wcpdf-setup-confetti' ); ?>
 			<?php wp_print_scripts( wp_script_is( 'wc-select2', 'registered' ) ? 'wc-select2' : 'select2' ); ?>
+			<?php wp_enqueue_emoji_styles(); ?>
 			<?php do_action( 'admin_print_styles' ); ?>
 			<?php do_action( 'admin_head' ); ?>
 		</head>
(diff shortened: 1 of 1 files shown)


alexmigf (commented):
Maybe this is not an issue from our plugin? I don't think we use that function.

BrunoPavlinic98:
> Maybe this is not an issue from our plugin? I don't think we use that function.

@alexmigf Our plugin triggers it with `do_action( 'admin_print_styles' );`, and here is the code from the core that explains it a bit better:

`add_action( 'wp_print_styles', 'print_emoji_styles' ); // Retained for backwards-compatibility. Unhooked by wp_enqueue_emoji_styles().`

alexmigf:
Good catch!

I think in that situation, you just need to remove the callbacks that are causing the issue? (I haven't tested)

```php
remove_action( 'wp_head', 'print_emoji_detection_script', 7 );
remove_action( 'wp_print_styles', 'print_emoji_styles' );
```

BrunoPavlinic98:
> ```
> remove_action( 'wp_print_styles', 'print_emoji_styles' );
> ```

@alexmigf Yes, works as well! `remove_action( 'admin_print_styles', 'print_emoji_styles' ); `is enough, so I think we are good to go with this one.

alexmigf (approved):
LGTM!

YordanSoares (approved):
LGTM!

PR #1577, opened 2026-07-24
Add sections to settings result

### LINKED ISSUE #1551: Improve settings search to include section names
[Plane work item: WPOVE-520](https://work.teamupdraft.com/team-updraft/browse/WPOVE-520)

The settings search should also consider the settings section name when matching results.

At the moment, some settings can be missed when the field label is too generic or does not include enough context. For example, a setting labelled **Checkout Field** may not be found unless the user searches for the exact label, because the related section context is not considered.

MohamadNateqi:
closes #1551

This PR searches by settings sections and also displays the section title to the right of each result:

image


alexmigf (approved):
Brilliant!

PR #1579, opened 2026-08-05
Fix: EDI line totals for high-quantity items

### LINKED ISSUE #1578: Invoice line totals fail Peppol validation for high-quantity items
[Plane work item: WPOVE-695](https://work.teamupdraft.com/team-updraft/browse/WPOVE-695)

Invoices with high quantities and low unit prices can fail Peppol rule R120 because the exported unit price no longer reproduces the declared line net amount after rounding. The validation error may affect multiple invoice lines.

alexmigf:
closes #1578


alexmigf:
Customer approved!

PR #1597, opened 2026-08-25
New: Add support for plural document titles

### LINKED ISSUE #1572: Add support for a plural document title
[Plane work item: WPOVE-570](https://work.teamupdraft.com/team-updraft/browse/WPOVE-570)

Currently, `get_title()` always returns the singular form, which works well for individual documents but is not suitable in contexts where multiple documents are displayed or processed together (e.g. summary of documents). Providing a dedicated plural title (e.g. via a new parameter) would make these use cases easier to support while avoiding hardcoded translations or pluralization logic outside the plugin.

alexmigf:
closes #1572

### DIFF
--- a/includes/Documents/BulkDocument.php
+++ b/includes/Documents/BulkDocument.php
@@ -85,14 +85,14 @@ class BulkDocument {
 		$pdf_filters = apply_filters( 'wpo_wcpdf_pdf_filters', array(), $this );
 		\wpo_ips_add_filters( $pdf_filters );
 
-		$html = $this->get_html();
+		$html         = $this->get_html();
 		$pdf_settings = array(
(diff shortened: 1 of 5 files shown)


YordanSoares (approved):
@alexmigf tested and LGTM!

wpo-nnawaz (commented):
get_plural_title() and plural_title() look fine but nothing actually calls them right now. Found one spot where it matters though: the wrapper template used for bulk PDFs sets the <title> tag from the singular title, so a merged PDF with multiple invoices still shows "Invoice" in the file metadata. Is that fix part of this PR? And is hooking this into Pro's summary feature and other documents planned as well?

alexmigf:
> so a merged PDF with multiple invoices still shows "Invoice" in the file metadata. Is that fix part of this PR?

Good catch and done.

> And is hooking this into Pro's summary feature and other documents planned as well?

Yes, I think that's the goal.

wpo-nnawaz (approved):
LGTM!

PR #1601, opened 2026-08-26
New: Improve registration number identifiers and defaults

### LINKED ISSUE #1569: Improve company registration number field
[Plane work item: WPOVE-557](https://work.teamupdraft.com/team-updraft/browse/WPOVE-557)

Rename the default **“COC Number”** label to **“Registration Number”**. Where possible, use a country\-specific official label based on the shop location. Add contextual help explaining what number is required, with a link to documentation.

We also need to add a way to select the scheme. This option was previously available in the E\-Documents settings, so we may need to uncomment the existing code to restore it. Please [see the customer discussion](https://secure.helpscout.net/conversation/3350910296/82960?viewId=221731) for additional context.

alexmigf:
closes #1569

This PR adds company registration number support to e-documents and centralizes optional checkout field handling.

### Changes:

* Include customer company registration numbers in UBL, Peppol, and CII output, and add supplier registration number schemes to UBL and CII.
* Introduce `IdentifierMappings` for country-specific VAT and company registration number mappings, including EAS, ICD, and registration number labels.
* Introduce `CheckoutField` to centralize field types, labels, availability, country rules, and typed order and customer storage.
* Add support for restricting the checkout field to selected billing countries and using a different field type for customers outside the Shop Country.
* Migrate legacy checkout field settings while preserving existing legacy values under their original type.
* Integrate typed checkout field values with Classic Checkout, Checkout Block, My Account, admin order display, and subscription renewals.
* Preserve third-party VAT plugin support when VAT is configured as either the primary or alternative checkout field type.
* Display customer registration numbers in the order E-Documents identifiers metabox.
* Add an Automatic supplier ICD option based on Shop Country while preserving manual overrides.
* Update shop and checkout registration number labels immediately when Shop Country changes.
* Make EAS and ICD selectors searchable while keeping all options browsable, and give enhanced E-Document selectors a consistent full-width layout.
* Deprecate the old Peppol VAT mappings function and filter while retaining backwards compatibility.
* Fix related Checkout Block and Store API field persistence and cleanup, Peppol Classic Checkout handling, and Peppol My Account endpoint registration.


wpo-ips-functions.php
```diff
@@ -2350,6 +2398,81 @@ function wpo_ips_register_additional_checkout_field( array $options ): void {
...
+ * Get the configured checkout field type.
+ *
+ * @return string One of: custom, vat_number, registration_number.
+ */
+function wpo_ips_get_checkout_field_type(): string {
+	$general_settings = get_option( 'wpo_wcpdf_settings_general', array() );
+	$general_settings = is_array( $general_settings ) ? $general_settings : array();
+	$type             = sanitize_key( (string) ( $general_settings['checkout_field_type'] ?? '' ) );
+
+	$allowed = array(
```
MohamadNateqi: Do you want to allow this to be modified by users using a filter?

wpo-ips-functions-edi.php
```diff
@@ -965,349 +990,518 @@ function wpo_ips_edi_build_peppol_endpoint_candidates_from_vat( string $billing_
...
  *
- * @return array
+ * @param string $country Optional country code in ISO 3166-1 alpha-2 format.
+ * @param string $type    Optional identifier type, e.g. 'vat_number' or 'registration_number'.
+ * @param string $key     Optional mapping key, e.g. 'eas', 'icd' or 'label'.
+ * @return array|string
  */
-function wpo_ips_edi_get_peppol_vat_mappings(): array {
+function wpo_ips_edi_get_identifier_mappings( string $country = '', string $type = '', string $key = '' ): array|string {
 	$mappings = array(
```
MohamadNateqi: This data structure is almost 400 lines in the function body. Would it be better to create a new file and keep it all there? Like in `data/edi-identifier-,appings.php`

wpo-ips-functions.php
```diff
@@ -2350,6 +2398,81 @@ function wpo_ips_register_additional_checkout_field( array $options ): void {
...
 }
 
+/**
+ * Get the configured checkout field type.
+ *
+ * @return string One of: custom, vat_number, registration_number.
+ */
+function wpo_ips_get_checkout_field_type(): string {
+	$general_settings = get_option( 'wpo_wcpdf_settings_general', array() );
+	$general_settings = is_array( $general_settings ) ? $general_settings : array();
```
MohamadNateqi: Why not using `WPO_WCPDF()->get_instance( 'settings' )->get_instance( 'general' );`?

includes/Install.php
```diff
@@ -751,6 +751,42 @@ protected function upgrade( string $installed_version ): void {
...
 
+		// 6.0.0 update: migrate the optional checkout field settings.
+		if ( version_compare( $installed_version, '6.0.0', '<' ) ) {
+			$general_settings = get_option( 'wpo_wcpdf_settings_general', array() );
+
+			if ( is_array( $general_settings ) ) {
+				$is_legacy_checkout_field = ! array_key_exists( 'checkout_field_type', $general_settings );
+
+				if ( $is_legacy_checkout_field ) {
+					$settings_changed = false;
```
MohamadNateqi: This can be removed.

includes/Install.php
```diff
@@ -751,6 +751,42 @@ protected function upgrade( string $installed_version ): void {
...
+
+					// Migrate the old VAT number checkbox to the new field type.
+					$general_settings['checkout_field_type'] = ! empty( $general_settings['checkout_field_as_vat_number'] )
+						? 'vat_number'
+						: 'custom';
+
+					$settings_changed = true;
```
MohamadNateqi: Same

includes/Settings/SettingsGeneral.php
```diff
@@ -284,7 +299,11 @@ public function init_settings(): void {
 			array(
 				'type'     => 'setting',
 				'id'       => 'coc_number',
-				'title'    => __( 'Shop Chamber of Commerce Number', 'woocommerce-pdf-invoices-packing-slips' ),
+				'title'    => sprintf(
+					/* translators: %s: company registration number label */
+					__( 'Shop %s', 'woocommerce-pdf-invoices-packing-slips' ),
+					$company_registration_number_label
```
MohamadNateqi, line 308: Do you think this value needs to be updated by JS as well, when the country changes, or it's not necessary?

includes/Settings/SettingsGeneral.php
```diff
@@ -568,34 +587,40 @@ public function init_settings(): void {
 			),
 			array(
 				'type'     => 'setting',
-				'id'       => 'checkout_field_label',
-				'title'    => __( 'Label', 'woocommerce-pdf-invoices-packing-slips' ),
-				'callback' => 'text_input',
+				'id'       => 'checkout_field_type',
```
MohamadNateqi, line 593: Maybe you need to store the value of check field type in the order meta as well, in case they change this setting, since then the old order will be interpreted by the new type.

includes/Settings/SettingsGeneral.php
```diff
@@ -568,34 +587,40 @@ public function init_settings(): void {
...
+				'callback' => 'select',
 				'section'  => 'general_settings',
 				'args'     => array(
 					'option_name' => $option_name,
-					'id'          => 'checkout_field_label',
-					'default'     => __( 'Customer identification', 'woocommerce-pdf-invoices-packing-slips' ),
-					'description' => __( 'The label for the optional custom field displayed at checkout.', 'woocommerce-pdf-invoices-packing-slips' ),
-				)
+					'id'          => 'checkout_field_type',
+					'default'     => ! empty( $this->get_setting( 'checkout_field_as_vat_number' ) )
```
MohamadNateqi: Do we need to check the legacy `checkout_field_as_vat_number` setting value, considering you wrote a migration? Same about other cases.

includes/Settings/SettingsEDI.php
```diff
@@ -437,6 +440,44 @@ public function init_settings(): void {
...
 
+		// Supplier Registration Number scheme
+		$settings_fields[] = array(
+			'type'     => 'setting',
+			'id'       => 'registration_number_scheme',
+			'title'    => '',
+			'callback' => 'select',
+			'section'  => $section,
+			'args'     => array(
+				'title'            => __( 'Supplier Registration Number Scheme (ICD)', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi, line 456: Would it be good to add search functionality for this field, to improve the UX?

edi/Syntaxes/Cii/Handlers/SupplyChainTradeTransaction/ApplicableHeaderTradeAgreementHandler.php
MohamadNateqi: What if the `$billing_company` is empty?

wpo-ips-functions-edi.php
```diff
@@ -965,349 +990,518 @@ function wpo_ips_edi_build_peppol_endpoint_candidates_from_vat( string $billing_
...
 				),
+				'registration_number' => array(),
 			),
 		),
 	);
 
-	return (array) apply_filters(
-		'wpo_ips_edi_peppol_vat_mappings',
-		$mappings
+	$mappings = (array) apply_filters(
```
MohamadNateqi: Is it safe to suddenly remove `wpo_ips_edi_peppol_vat_mappings` filter?

includes/Settings/SettingsEDI.php
```diff
@@ -437,6 +440,44 @@ public function init_settings(): void {
...
+				'option_name'      => $option_name,
+				'id'               => 'registration_number_scheme',
+				'default'          => $default_registration_scheme,
+				'default_if_empty' => true,
```
MohamadNateqi: Shouldn't we allow users to turn it off?

includes/Settings/SettingsEDI.php
```diff
@@ -437,6 +440,44 @@ public function init_settings(): void {
...
+				'disabled'         => empty( $registration_number ),
+				'options'          => ( function () {
+					$options = array(
```
MohamadNateqi, line 468: Should we have a `none` option as well?

wpo-ips-functions-edi.php
```diff
@@ -1479,6 +1674,14 @@ function wpo_ips_edi_get_order_customer_identifiers_data( \WC_Order $order ): ar
 		),
 	);
 
+	if ( 'registration_number' === wpo_ips_get_checkout_field_type() ) {
```
MohamadNateqi: Would it be better to show this row, if the value is not empty, instead of relying on the type?

includes/Admin.php
```diff
@@ -2232,16 +2252,12 @@ private function get_order_meta_box_document_xml_action( string $document_type,
...
 	 *
 	 * @param \WC_Order $order
 	 * @return void
 	 */
-	private function get_order_meta_box_peppol_identifiers( \WC_Order $order ): void {
-		if ( ! wpo_ips_edi_peppol_is_available() ) {
-			return;
-		}
-
+	private function get_order_meta_box_customer_identifiers( \WC_Order $order ): void {
```
MohamadNateqi, line 2264: Removing the `wpo_ips_edi_peppol_is_available()` check means the table now shows for all EDI formats, and since `vat_number` is still required, orders without a VAT number get a "Missing" label.
Do all formats needs a VAT number?

MohamadNateqi (changes requested):

includes/Settings/SettingsEDI.php
alexmigf: I think I may have answered this question before. My concern is that if we hide the options, it can be difficult to know what to search for in the first place, so search may not be very helpful in that case.

wpo-ips-functions.php
alexmigf: I don’t see much benefit in doing that right now. We can always revisit it later if there’s a good reason to change it.

includes/Settings/SettingsGeneral.php
alexmigf: I think the migration already covers that, doesn’t it? What other cases are you thinking of?

wpo-ips-functions-edi.php
alexmigf: It's replaced by this: `wpo_ips_edi_identifier_mappings`

includes/Settings/SettingsEDI.php
alexmigf: I intentionally used `default_if_empty` because an empty value falls back to the registration scheme for the shop country during document generation anyway.

includes/Settings/SettingsEDI.php
alexmigf: Since the selector is disabled when the Registration Number is empty, I don’t think we need a `none` option. If a Registration Number is provided, we should normally have a scheme for it. The default is derived from the shop country, and users can override it if needed.

includes/Settings/SettingsGeneral.php
alexmigf: I wanted to keep it simple, but I had thought about that too. Since you mentioned it, maybe I should go ahead and implement it.

includes/Settings/SettingsEDI.php
MohamadNateqi: Yeah, you're right. What I had in mind was a search field above the dropdown list, not hiding the list.

Like this one: https://www.w3schools.com/howto/howto_js_filter_dropdown.asp

wpo-ips-functions.php
MohamadNateqi: Yeah, sure. I thought maybe they might want to add other options to the types.

includes/Settings/SettingsGeneral.php
MohamadNateqi: Yeah, what I meant is that since the migration covers it, maybe you don't need to check for the legacy setting key (`checkout_field_as_vat_number`)?

wpo-ips-functions-edi.php
MohamadNateqi: Yes, but would it be better to apply that deprecated filter for a while? In case users are already using it.

includes/Settings/SettingsEDI.php
MohamadNateqi: The runtime fallback makes sense, but `default_if_empty` stops it from running. It pre-selects the country's ICD, so saving the EDI settings stores it. 
For example, an NL shop saves the EDI tab and `0106` is stored. Later they change the Shop Country to BE, but documents keep using `0106` because the value isn't empty anymore.

But the empty option could show what will be used, for example: Automatic ([0106] KVK number).

Alternatively, we could show a note on the scheme field, or the shop country, or both, when the saved scheme differs from the Shop Country.

edi/IdentifierMappings.php
```diff
@@ -0,0 +1,565 @@
...
+	/**
+	 * Get identifier mappings by country.
+	 *
+	 * @param string $country Optional country code in ISO 3166-1 alpha-2 format.
+	 * @param string $type    Optional identifier type, e.g. 'vat_number' or 'registration_number'.
+	 * @param string $key     Optional mapping key, e.g. 'eas', 'icd' or 'label'.
+	 * @return array|string
+	 */
+	public static function get( string $country = '', string $type = '', string $key = '' ): array|string {
+		$mappings = array(
```
MohamadNateqi: Can you cache this data array so it doesn't get rebuilt on each invocation?

includes/Compatibility/VatPlugins.php
```diff
@@ -176,12 +178,24 @@ public function has_active(): bool {
...
+
+		// When no third-party VAT plugin is active, only use our generic
+		// checkout field as a VAT source if it is configured as such.
+		$checkout_field = \WPO_WCPDF()->get_instance( 'checkout_field' );
+
+		if (
+			empty( $info['active'] ) &&
+			( ! $checkout_field->is_enabled() || ! $checkout_field->is_type( CheckoutField::TYPE_VAT_NUMBER ) )
+		) {
+			return '';
```
MohamadNateqi: Since it returns an empty string, it can raise this error in the console now:

Image

It happens because in `peppol-classic-checkout.js`, on line `134`, the `VAT_SELECTOR` is empty:

`$( document ).on( 'input change', COUNTRY_SELECTOR + ', ' + VAT_SELECTOR, ...`

includes/CheckoutField.php
```diff
@@ -0,0 +1,436 @@
...
+					self::LEGACY_ORDER_META_KEY,
+					true
+				)
+			);
+
+			if ( '' === $value ) {
+				return null;
+			}
+
+			$order->update_meta_data(
```
MohamadNateqi: This method is a getter, but it also rewrites a value and deletes metadata. Could it be confusing?
Would it be better to remove this part and keep the method only as a getter, considering that we have the `save_order_value()` method that cleans up the legacy metadata?

includes/Frontend.php
```diff
@@ -45,29 +45,32 @@ public function __construct() {
...
 			// Blocks/store-api hooks
 			$this->checkout_field_display_checkout_block_field();
 			$this->checkout_field_set_checkout_block_field_value();
 
 			add_action( 'woocommerce_set_additional_field_value', array( $this, 'checkout_field_save_checkout_block_field' ), 10, 4 );
 			add_action( 'woocommerce_store_api_checkout_order_processed', array( $this, 'checkout_field_remove_order_checkout_block_field_meta' ), 10, 1 );
 
+			add_action( 'woocommerce_customer_loaded', array( $this, 'checkout_field_remove_customer_checkout_block_field_meta' ) );
+			add_action( 'woocommerce_cart_loaded_from_session', array( $this, 'checkout_field_remove_session_checkout_block_field_meta' ) );
+			$this->checkout_field_remove_session_checkout_block_field_meta();
```
MohamadNateqi, line 60: These lines trigger `delete_user_meta()`, which runs on every frontend page for logged in users, up to three times per request, even when there's nothing to delete.

Can we make some improvements to this? For example, run it only on special pages, like checkout.

MohamadNateqi (changes requested):
The dropdown comes on top of the top nav:
image

This PR has a major conflict with the https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/pull/1495 PR, so I guess we should close that one and apply the changes in another PR.

includes/Admin.php
```diff
@@ -692,8 +693,8 @@ public function edi_actions_meta_box( object $post_or_order_object ): void {
 			'</p></div>';
 		}
 
-		// Peppol specific
-		echo $this->get_order_meta_box_peppol_identifiers( $order ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		// Customer identifiers
+		echo $this->get_order_meta_box_customer_identifiers( $order ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
MohamadNateqi: The method signature is `void`, but it's echoing the result.

wpo-ips-functions.php
```diff
@@ -1266,18 +1266,14 @@ function wpo_wcpdf_get_order_customer_vat_number( \WC_Abstract_Order $order ): ?
 		'_billing_btw_nummer'     // Some Belgium customers use this key as a custom field
 	), $order );
 	
-	$frontend_instance = WPO_WCPDF()->get_instance( 'frontend' );
+	$checkout_field = \WPO_WCPDF()->get_instance( 'checkout_field' );
+	$vat_number     = $checkout_field->get_order_value( $order, \WPO\IPS\CheckoutField::TYPE_VAT_NUMBER );
+	$meta_key       = null;
```
MohamadNateqi, line 1271: It seems this can be removed.

MohamadNateqi (changes requested):

includes/Settings/SettingsGeneral.php
```diff
@@ -566,6 +588,60 @@ public function init_settings(): void {
...
+						/* translators: %s: WooCommerce EU VAT Compliance plugin link */
+						__( 'For advanced VAT validation, reporting, and full compliance with EU VAT rules, we recommend using %s.', 'woocommerce-pdf-invoices-packing-slips' ),
+						'<a href="https://wpovernight.com/downloads/woocommerce-eu-vat-compliance/?utm_medium=plugin&utm_source=ips&utm_campaign=general-tab&utm_content=woocommerce-eu-vat-compliance-cross" target="_blank" rel="noopener noreferrer">WooCommerce EU VAT Compliance</a>'
+					) . $vat_plugin_notice,
+				),
+			),
+			array(
+				'type'     => 'setting',
+				'id'       => 'checkout_field_alternative_enable',
+				'title'    => __( 'Use a different field type for foreign customers', 'woocommerce-pdf-invoices-packing-slips' ),
```
MohamadNateqi, line 622: Can we have a better title for this? It's like a description for a title. Maybe just "Foreign customers"

includes/Settings/SettingsGeneral.php
```diff
@@ -566,6 +588,60 @@ public function init_settings(): void {
...
+				'id'       => 'checkout_field_alternative_type',
+				'title'    => __( 'Alternative field type', 'woocommerce-pdf-invoices-packing-slips' ),
+				'callback' => 'select',
+				'section'  => 'general_settings',
+				'args'     => array(
+					'option_name' => $option_name,
+					'id'          => 'checkout_field_alternative_type',
+					'options'     => \WPO_WCPDF()->get_instance( 'checkout_field' )->get_type_options(),
+					'custom_attributes' => array(
+						'data-show_for_option_name'   => $option_name . '[checkout_field_alternative_enable]',
```
MohamadNateqi, line 642: Would it be better not to hide it, just disable/gray it out? For better UX.

includes/CheckoutField.php
```diff
@@ -0,0 +1,539 @@
...
+
+			$type  = $type ?: $this->get_type();
+			$label = $type === $this->get_type()
+				? trim( (string) $general_settings->get_setting( 'checkout_field_label' ) )
+				: '';
+
+			if ( '' === $label ) {
+				$label = $this->get_default_label(
+					$type,
+					$this->get_shop_country()
```
MohamadNateqi, line 226: If the "shop country" setting is set to a country like NL, for which we have a label in the `IdentifierMappings`, and the foreign customer option is enabled and set to the "Company registration number", and the billing country is something different than the shop country, the label would fall back to the shop country's label, like KVK, while the billing country is something else:

Image

includes/CheckoutField.php
```diff
@@ -0,0 +1,539 @@
...
+			return $this->get_type() === $this->normalize_type( $type );
+		}
+
+		/**
+		 * Get the order meta key for a field type.
+		 *
+		 * @param string $type Field type. Defaults to the configured type.
+		 * @return string
+		 */
+		public function get_order_meta_key( string $type = '' ): string {
```
MohamadNateqi, line 161: Would it be good to add the word " type" to the function name? It can be confusing to know what kind of meta key this is returning without looking into the method body.

wpo-ips-functions.php
```diff
@@ -1266,44 +1266,95 @@ function wpo_wcpdf_get_order_customer_vat_number( \WC_Abstract_Order $order ): ?
...
 
-	if ( ! empty( $frontend_instance ) && is_callable( array( $frontend_instance, 'checkout_field_is_vat_number' ) ) ) {
-		$checkout_field_is_vat_number = $frontend_instance->checkout_field_is_vat_number();
```
MohamadNateqi, line 1215: Shouldn't you check if a VAT plugin is active, same as before?

includes/Settings/SettingsGeneral.php
```diff
@@ -68,21 +68,39 @@ public function output( string $section, string $nonce ): void {
...
+		$settings_instance                 = \WPO_WCPDF()->get_instance( 'settings' );
+		$checkout_field_type               = \WPO_WCPDF()->get_instance( 'checkout_field' )->get_type();
+
+		$checkout_field_default_label = \WPO_WCPDF()->get_instance( 'checkout_field' )->get_default_label(
+			$checkout_field_type,
+			$shop_country
+		);
+
+		$registration_labels = array();
+		foreach ( array_merge( array( '' ), array_keys( \WC()->countries->get_countries() ) ) as $country_code ) {
```
MohamadNateqi: We only have registration numbers for almost 10 countries, but this loop iterates over many countries just to process those 10. Can we improve this?

MohamadNateqi (changes requested):

includes/CheckoutField.php
alexmigf: In the Checkout must follow the Shop Country.

includes/Settings/SettingsGeneral.php
alexmigf: I don't think your suggestion is descriptive enough. It should be clear at first glance what the setting is for, without having to read the description below.

includes/Settings/SettingsGeneral.php
alexmigf: We already use this approach for several fields in our settings, so I just reused the same pattern here. It helps save space, and I don’t see it causing any UX issues. We also use accordions in the settings, which you previously suggested, and they follow a similar principle.

wpo-ips-functions.php
alexmigf: The function now reads the VAT value saved on the order, independently of the current checkout configuration. An active VAT plugin disables our checkout field, but we still retain previously saved VAT values.

includes/CheckoutField.php
MohamadNateqi: So for a foreign customer, having a different country than the shop country, we should show the label based on the shop country?

includes/Settings/SettingsGeneral.php
MohamadNateqi: Yeah, I didn't mean it was an issue, but when I was toggling this option, it felt confusing to me which settings had been changed or added after enabling it, so I thought keeping the option grayed out might improve this.
And of course, if we decide to do it, it should change everywhere, not just here. But it's not for this PR.

includes/Settings/SettingsGeneral.php
MohamadNateqi: I thought we might add this text to the description and use a short title, but if a short title isn't descriptive enough, let's keep it.

includes/CheckoutField.php
alexmigf: SIREN is required in France, but that doesn’t necessarily mean you need to provide it if your company is based in the Netherlands and the shop is in France. I initially thought the same as you, but after looking into it further, I believe that assumption is incorrect.

Likewise, if the customer is in the Netherlands, they shouldn’t need to fill in this field just because the store is in France, since KVK does not apply in that context.

MohamadNateqi: So if I understand correctly, for foreign customers the company registration number label should follow the customer's country rather than the shop country, right?

At the moment, it's following the Shop Country, as I mentioned in my example.

alexmigf: No, the inverse, it should follow the Shop Country.

MohamadNateqi: Got it, thanks.

includes/CheckoutField.php
alexmigf: I’d keep `get_order_meta_key()` since it describes what the method returns, and the `$type` argument already makes the selection clear. Adding "type" could suggest it returns the meta key where the type itself is stored, rather than the field’s value.

includes/CheckoutField.php
```diff
@@ -0,0 +1,539 @@
...
+				: '';
+
+			if ( '' === $label ) {
+				$label = $this->get_default_label(
+					$type,
+					$this->get_shop_country()
+				);
+			}
+
+			return (string) apply_filters( 'wpo_ips_checkout_field_label', $label );
```
MohamadNateqi: Maybe passing `$type` as well?

includes/Install.php
```diff
@@ -751,6 +751,44 @@ protected function upgrade( string $installed_version ): void {
...
+
+				// Keep this outside the editable settings so type changes cannot reinterpret old values.
+				add_option(
+					'wpo_ips_checkout_field_legacy_type',
+					! empty( $general_settings['checkout_field_as_vat_number'] ) ? 'vat_number' : 'custom'
+				);
+
+				if ( $is_legacy_checkout_field ) {
+					// Migrate the old VAT number checkbox to the new field type.
+					$general_settings['checkout_field_type'] = ! empty( $general_settings['checkout_field_as_vat_number'] )
```
MohamadNateqi, line 769: Maybe you should check `\WPO_WCPDF()->get_instance( 'vat_plugins' )->has_active()` here, and set the type to `custom`, in case `has_active()` returns `true`?

includes/Compatibility/VatPlugins.php
```diff
@@ -176,12 +178,33 @@ public function has_active(): bool {
...
+				return '';
+			}
+
+			if ( in_array( $field_id, array( CheckoutField::ALTERNATIVE_BLOCK_FIELD_ID, CheckoutField::ALTERNATIVE_CLASSIC_FIELD_KEY ), true ) ) {
+				return 'block' === $context
+					? '.wc-block-components-address-form__' . str_replace( '/', '-', $field_id ) . ' input'
+					: '#' . $field_id . '_field input:not(:disabled)';
+			}
+
+			if ( 'classic' === $context && $checkout_field->get_alternative_type() ) {
```
MohamadNateqi: Should it apply to any kind of classic checkout fields, not just alternative types?

includes/Frontend.php
```diff
@@ -429,56 +497,144 @@ static function ( $value, string $group, \WC_Data $wc_object ) {
...
+	 * Enable customer metadata cleanup for Store API checkout callbacks.
+	 *
+	 * @param mixed            $response REST response.
+	 * @param array            $handler  Route handler.
+	 * @param \WP_REST_Request $request  REST request.
+	 * @return mixed
+	 */
+	public function checkout_field_enable_rest_cleanup( mixed $response, array $handler, \WP_REST_Request $request ): mixed {
+		// Inspect the dispatched route so this also works for Store API batch requests.
+		$this->checkout_field_rest_cleanup = (bool) preg_match( '#^/wc/store/v[0-9]+/checkout(?:/|$)#', $request->get_route() );
```
MohamadNateqi: I noticed that Woo has an [unversioned namespace](https://developer.woocommerce.com/docs/apis/store-api/?utm_source=chatgpt.com#:~:text=If%20the%20version%20is%20omitted%2C%20v1%20will%20be%20served.) as well, which falls back to `v1` in case it's not provided. Maybe you need to consider that here as well?

wpo-ips-deprecated-hooks.php
```diff
@@ -49,6 +49,7 @@
 	'wpo_wc_ubl_orderTaxData'                           => array( 'wpo_ips_edi_ubl_order_tax_data',                    '5.0.0' ),
 	'wpo_wc_ubl_handle_TaxTotal'                        => array( 'wpo_ips_edi_ubl_tax_total',                         '5.0.0' ),
 	'wpo_wc_ubl_handle_UBLVersionID'                    => array( 'wpo_ips_edi_ubl_version_id',                        '5.0.0' ),
+	'wpo_ips_edi_peppol_vat_mappings'                   => array( 'wpo_ips_edi_identifier_mappings',                   '6.0.0' ),
```
MohamadNateqi: Should it be removed from here? Since I think it'll duplicate the log.

MohamadNateqi (changes requested):
I have selected an option for the `Peppol Endpoint Scheme (EAS)` setting, then tried to change it to "Select...", but it didn't change after save, it returned to the previous selected option. Is it fine for it to be like this?

alexmigf:
> I have selected an option for the Peppol Endpoint Scheme (EAS) setting, then tried to change it to "Select...", but it didn't change after save, it returned to the previous selected option. Is it fine for it to be like this?

This doesn’t seem related to this PR, but I don’t see a problem with it. You have to choose one if you want to use Peppol anyway, and if you don’t, it doesn’t hurt anything.

includes/Install.php
alexmigf: No, I don’t think so. That’s a decision for the user to make, and we shouldn’t make it on their behalf. We treat whatever is saved in the settings as the source of truth.

MohamadNateqi: I was wondering: what if a VAT plugin is active? That would mean our checkout field acts as a custom field, so it can include non-VAT values for orders that have already been submitted.

alexmigf: Good catch! I thought activating a VAT plugin cleared the saved checkbox value, but it only disables the checkbox. I’ve updated the migration to account for an active VAT plugin.

MohamadNateqi (approved):
Great job!

YordanSoares (commented):
I've tested both UBL and CII syntax and confirmed that the registration number identifiers are being saved correctly. I also checked the generated documents with ecosio.com, and both types passed validation.

However, I found the introduction of the alternative field type for foreign customers a bit confusing from a UI perspective.

For example, when you enable the **"Use a different field type for foreign customers"** option, a new conditional field appears immediately below it. Since these options for the alternative field are placed in the middle, the fields that follow, which are used to customize the labels and the field displayed based on the country selection, become somewhat out of context.

At that point, it's not immediately clear whether those settings apply to the main custom field, the alternative field, or both.

I think we could make this clearer by grouping the settings for each field together, or otherwise visually indicating which settings apply to which field.

PR #1602, opened 2026-08-27
Add log note for failed document creations

### LINKED ISSUE #1600: Add log note for failed document creations
[Plane work item: WPOVE-732](https://work.teamupdraft.com/team-updraft/browse/WPOVE-732)

[This user](https://secure.helpscout.net/conversation/3413109814/84366?viewId=221726) a while ago couldn't create credit notes with the Montanio payment plugin active. The Montanio devs had investigated on their end and found that the the issue was from a code snippet on the user's site that worked with Mollie payments. The snippet returned a `null` in the case of all documents except invoices. Removing the snippet solved the issue. The devs have mentioned that it would be good to add a warning in the UI or logs for cases where documents weren't created because a `null` was returned, instead of the document creation failing silently in these cases.

BrunoPavlinic98:
closes #1600

### DIFF
--- a/wpo-ips-functions.php
+++ b/wpo-ips-functions.php
@@ -137,8 +137,7 @@ function wcpdf_get_document( string $document_type, mixed $order, bool $init = f
 }
 
 /**
- * Log a document object that was discarded by a callback hooked to one of our document filters,
- * and store it so it can be surfaced in the admin.
+ * Log a document object that was discarded by a callback hooked to one of our document filters.
  *
  * @param string $hook_name      Filter that discarded the document.
  * @param string $document_type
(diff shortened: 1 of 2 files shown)


alexmigf:
@BrunoPavlinic98 I think this would be better handled at the log level rather than adding a notice and potentially cluttering the UI.

BrunoPavlinic98:
> @BrunoPavlinic98 I think this would be better handled at the log level rather than adding a notice and potentially cluttering the UI.

@alexmigf I agree, and I have removed the UI notices, so only logs remain.

wpo-ips-functions.php
```diff
@@ -44,17 +44,25 @@ function wcpdf_get_document( string $document_type, mixed $order, bool $init = f
...
-			? $document
-			: false;
+		if ( is_object( $filtered ) ) {
+			return $filtered;
+		}
+
+		// An explicit false means a callback deliberately skipped the document, anything else
+		// (usually a missing return statement) discards it, so warn instead of failing silently.
+		if ( false !== $filtered && is_object( $document ) ) {
+			wpo_ips_log_discarded_document( 'wcpdf_get_document', $document_type, $order, $filtered );
```
alexmigf: Maybe just calling `wcpdf_log_error()` here would be enough. This seems like a lot of code to handle a relatively simple situation that doesn't occur very often.

alexmigf (changes requested):

wpo-ips-functions.php
BrunoPavlinic98: Good catch, and I agree. It is ready for review again.

alexmigf (approved):
LGTM!

PR #1604, opened 2026-08-31
Fix: Refund AJAX order permission checks

### LINKED ISSUE #1603: Align refund AJAX access checks with order permissions
Update the refund-related AJAX handler to validate access for each requested order before processing it, and add coverage for both permitted and denied cases.

[WPOVE-761](https://work.teamupdraft.com/team-updraft/browse/WPOVE-761/)

alexmigf:
closes #1603

### DIFF
--- a/includes/Main.php
+++ b/includes/Main.php
@@ -641,7 +641,32 @@ class Main {
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


MohamadNateqi (approved):
LGTM!

PR #1606, opened 2026-09-01
New: Add abstract `SummaryDocument` base class

### LINKED ISSUE #1605: Add abstract Summary document base class
Add an abstract `SummaryDocument` class to the core plugin with the common PDF, HTML, template, filename, date, and output handling for summary documents.

[WPOVE-767](https://work.teamupdraft.com/team-updraft/browse/WPOVE-767/)

alexmigf:
closes #1605

### DIFF
--- a/includes/Documents/SummaryDocument.php
+++ b/includes/Documents/SummaryDocument.php
@@ -16,7 +16,6 @@ if ( ! class_exists( '\\WPO\\IPS\\Documents\\SummaryDocument' ) ) :
 		public array $settings           = array();
 		public array $common_settings    = array();
 		public array $output_formats     = array();
-		public ?object $wrapper_document = null;
 
 		/**
 		 * Constructor.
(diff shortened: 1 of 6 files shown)


MohamadNateqi (commented):
I have some thoughts about these changes, I would like to ask:
1. A good chunk of this new class is a duplication of OrderDocument, and it has some small changes here and there. Would it be better to extract the common methods into a shared parent so that both SummaryDocument and OrderDocument can extend it? That would also remove the need to use a union type hint.
2. I'm wondering if the summary concept itself belongs in core, since the actual implementation lives in Pro. Following the first point, would it make more sense for the shared parent to be a generic document base, one that covers documents not tied to a single order, and leave the summary parts to Pro, so it can extend this generic document base class and implement the Summary?


Update:
Just to explain a bit more what I was thinking:

For example, the shared parent would hold methods common to all documents, such as `get_type()`, `get_title()`, `get_setting()`, `get_pdf()`, and so on.
Everything tied to a specific kind of document stays in its own class, like order numbers, dates, etc in `OrderDocument`, and the order IDs and other logic in the summary class.                                
That way we can keep the rendering part in one place, and each document type only defines what actually makes it different.

What do you think?

alexmigf:
@MohamadNateqi 

I agree with your point of view; it makes sense to me. This PR came to mind because I noticed some type hint errors while testing the Bulk Export changes made by Noman, but the issue was probably caused by the Pro Summary abstract not extending `OrderDocument` itself.

I think I’ll close this for now and suggest making that change in Pro instead. That said, I still think your idea is valid as a future improvement, so I’d suggest opening an intake for future consideration.

